PR #3, opened 2020-02-04
static fov feature

eatmyvenom:
A simple bit of code to allow for static fov in sodium.
No performance improvement is made by this and is only here as a feature


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
```diff
@@ -31,6 +31,7 @@
         public boolean enableVignette = true;
         public boolean enableFog = true;
         public boolean enableClouds = true;
+        public boolean staticFov = true;
```
TheGlitch76: This should be false by default since it's a cosmetic change to vanilla behavior, IMO.

eatmyvenom: Yeah that's true, I'll change this over in a few.

PR #6, opened 2020-04-09
An option to disable the ingame hurtcam

eatmyvenom:
This is to make it so if the player wants to not be disturbed by the ingame hurtcam while they might take damage, they can disable it here.


liach:
Hmm, I wonder how this relates to optimizing the game

eatmyvenom:
> Hmm, I wonder how this relates to optimizing the game

This is here as a feature that could be added in this mod. Not all of it pure optimizations, such as the smooth lighting (which is a fix) or the ability to disable the vignette. Overall no its not quite an optimization but rather a nice to have and ultimately is jelly's decision whether to include things like this in her mod. I'm simply doing this as a easy way for her to add this kinda stuff.

TheGlitch76:
This looks like a cheat and would give you an edge in PVE and PVP.

jellysquid3:
I agree with the idea that this could be used to give players an unfair advantage, however we do already include an option to disable the in-game fog (even though it's so cheap that I seriously doubt it makes any difference on hardware made after 2009...) That option is also up for similar debate.

Anyways, as per the other closure, these kind of features are better suited (in my opinion) to a standalone Fabric micro-mod to help prevent feature creep.

PR #7, opened 2020-04-13
Explicitly depend on Java 9+

dexman545:
Crash and state that Java 9+ is needed on startup.


jellysquid3:
Sodium does not actually require Java 9 or newer. The issue arises when a newer version of Java is used to compile the mod, as it will build against the newer standard library and produce code that is backwards incompatible with Java 8 because of a change on the return type of `ByteBuffer#flip()`.

It would be better to replace these usages with a helper method which casts `ByteBuffer` to `Buffer`, then calls `Buffer#flip()` upon the cast type.

PR #8, opened 2020-04-17
Static fov feature

eatmyvenom:
this is #3 but dont mind me Im just switching branches around


jellysquid3:
I'm closing this issue as in my opinion, such enhancements or tweaks would be better suited for a standalone Fabric micro-mod. This is largely to prevent feature creep into Sodium, as the primary focus is that of optimization.

PR #11, opened 2020-05-07
Change Biome Blend Default to match vanilla

CGDoctor8:
Vanilla's default biome blend setting is 5x5 but Sodium's is 3x3

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
@@ -38,7 +38,7 @@ public class SodiumGameOptions {
         public boolean enableClouds = true;
 
         public LightingQuality smoothLighting = LightingQuality.HIGH;
-        public int biomeBlendDistance = 3;
+        public int biomeBlendDistance = 5;
     }
 
     public enum DefaultGraphicsQuality implements TextProvider {


TheMadHau5:
Sodium's biome blend is also much smoother so 3x3 should be fine

franiscoder:
Yeah, Sodium can do the same effect vanilla did with 7 x 7 in just 1x1 so having 3x3 (AND BEING FASTER AT THE SAME TIME) is just a better idea.

jellysquid3:
It looks like this PR wasn't updated following the discussion on Discord a few weeks back.

Sodium uses an alternative biome blending algorithm which colorizes each vertex of a block face independently by linearly interpolating over the averaged "biome color" vanilla computes for each adjacent block. This allows it to take advantage of the graphics card to render smooth gradients using the existing color data attached to each vertex, which in turn produces much better looking results.

This is in comparison to vanilla, which colorizes all vertices of a block face the same way given the computed biome color for a block. Most users agree that Sodium's biome blender looks significantly better than vanilla with the current defaults, and the reduction in sample count (121 -> 49) is a big win for chunk rendering performance as well.

For a visual comparison, here is vanilla's blender at 5x5 (top) vs Sodium's blender at 3x3 (bottom).

image

As such, I think the current defaults are sufficient compromise between performance and render quality.

PR #27, opened 2020-06-20
Distribute lighting unloads over multiple ticks

Gegy:
This reduces the freezes that come from handling `UnloadChunkS2CPacket` by splitting the consequent lighting updates over multiple ticks. 
It does this in a batch size of 64, so in the worst case at 32 chunk render distance when teleporting to a completely new area, it should finish processing all the updates over ~3.2s (64 ticks).

This isn't an ideal solution, but there's no urgency for lighting updates when unloading chunks anyway. A better solution in the case of teleporting to a completely new area would be eliding light propagation when the chunk has no neighbours, but this might add a lot of extra complexity that's not really worth looking into.

This PR is also fairly invasive unfortunately with 4 mixins. I couldn't think of any way to reduce this nicely, though. Can also add documentation if needed. :)


jellysquid3:
I believe this pull request would be better suited for [Phosphor](https://github.com/jellysquid3/phosphor-fabric) as that mod targets the lighting engine specifically both on the client and server.

Gegy:
Ah okay, cool, will do- I initially thought it’d make sense here due to this only making changes to the clientside

PR #31, opened 2020-06-22
Some Tiny QoL changes & additions

juliand665:
Changes should be summed up in commit subjects/messages. I hope I followed your format correctly!

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -28,6 +28,15 @@ public class SodiumGameOptionPages {
 
         groups.add(OptionGroup.createBuilder()
                 .add(OptionImpl.createBuilder(int.class, vanillaOpts)
+                        .setName("GUI Scale")
+                        .setTooltip("Controls how many pixels wide/high a pixel of the GUI is, effectively scaling it. Set to 0 for automatic scaling.")
+                        .setControl(option -> new SliderControl(option, 0, 16, 1, v -> v == 0 ? "Automatic" : v + "x"))
+                        .setBinding((opts, value) -> {
+                            opts.guiScale = value;
+                            MinecraftClient.getInstance().onResolutionChanged();
+                        }, opts -> opts.guiScale)
+                        .build())
+                .add(OptionImpl.createBuilder(int.class, vanillaOpts)
                         .setName("View Distance")
                         .setTooltip("The view distance controls how far away terrain will be rendered. Lower distances mean that less terrain will be " +
                                 "rendered, improving frame rates.")
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -40,7 +40,7 @@ public class SodiumGameOptionPages {
                         .setName("View Distance")
                         .setTooltip("The view distance controls how far away terrain will be rendered. Lower distances mean that less terrain will be " +
                                 "rendered, improving frame rates.")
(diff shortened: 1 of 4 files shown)


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
```diff
@@ -27,11 +27,20 @@ public static OptionPage general() {
         List<OptionGroup> groups = new ArrayList<>();
 
         groups.add(OptionGroup.createBuilder()
+                .add(OptionImpl.createBuilder(int.class, vanillaOpts)
+                        .setName("GUI Scale")
+                        .setTooltip("Controls how many pixels wide/high a pixel of the GUI is, effectively scaling it. Set to 0 for automatic scaling.")
+                        .setControl(option -> new SliderControl(option, 0, 16, 1, v -> v == 0 ? "Automatic" : v + "x"))
```
jaskarth, line 33: Setting the max to 16 by default will probably confuse users as with most monitors it doesn't reach that high. Ideally this would check the dimensions of the screen and then reduce the number accordingly, but this causes issues when resizing the window while the sodium gui is open. MC provides an automatic way to do this, through `MinecraftClient.getInstance().getWindow().calculateScaleFactor(0, MinecraftClient.getInstance().forcesUnicodeFont())`, but listening to the resize is the issue here.

jaskarth (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
AMereBagatelle: Screen has a method called `resize`, I think vanilla uses it for adjusting as well.

juliand665: I'm not sure that's better. Perhaps a lower maximum like 8 would do, but I wouldn't want it to have a lower maximum just because my window is smaller at the time i'm looking at it.

Minenash: Maybe change the name to something like "Max GUI Scale" and in the tooltip it shows what the current possible maximum is. (A greyed out, not interactible slider bar to also show this would be awesome, but probably too complex to be worth it)

So the maximum doesn't go down just because the current screen size doesn't support it and it tells users what the actual scale would be.

jellysquid3:
The reason why the GUI Scale option wasn't added immediately is because it depends on the client's "capabilities" (similar to the full-screen resolution.) As other users have pointed out here, there's not an easy way currently to notify controls that the capabilities have changed and as such it's rather annoying to implement something similar to how vanilla does it.

I think other work is needed in the options screen first before 0.1's release anyways, and it would be good to get this sorted out with some kind of "capabilities changed" event that could notify option controls that their valid value range has changed. The proposed solutions here to add a "Max scale" or >10 values do not feel great.

juliand665:
Yeah, I get wanting to get things just right, and in the end it's your mod, but I think the utility of at least having a setting there far outweighs the potential annoyance with it not quite working as expected. Not to mention other mods that turn it into a slider have done it this way for ages, so most people are used to it anyway. We can always refine it later, but in my opinion this is a clear improvement over the current situation.

jellysquid3:
Sure, I agree that sorting this out now is more important. The current solution in this PR is fine for the time being. I'm just leaving my commentary on the underlying problem and that this should be looked into more closely in the future.

jellysquid3 (approved):
LGTM.

jellysquid3:
Commit 3ebf8ec7 now includes a GUI scale option in the spirit of this PR.

juliand665:
What do you think about the hotkey approach to quickly apply and/or exit? I feels pretty natural to me, while avoiding accidentally discarding unapplied changes.

PR #40, opened 2020-06-29
[Fix] Restore Performance Impact Text Style

CDAGaming:
This typo caused an LF character to appear, as \n was used instead of the proper migrated Style.
This commit restores that behavior from 1.15.x


CDAGaming:
Closed as Resolved

PR #41, opened 2020-06-29
Move bottom-right buttons next to options

sammko:
I found it annoying having to go all the way to the other side of the screen to apply.
image

Of course, this is purely subjective, feel free to close just based on opinion.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
@@ -77,11 +77,11 @@ public class SodiumOptionsGUI extends Screen {
         }
 
         this.rebuildGUIPages();
-        this.rebuildGUIOptions();
+        int y = this.rebuildGUIOptions() + 4;
 
-        this.undoButton = new FlatButtonWidget(new Dim2i(this.width - 211, this.height - 30, 65, 20), "Undo", this::undoChanges);
-        this.applyButton = new FlatButtonWidget(new Dim2i(this.width - 142, this.height - 30, 65, 20), "Apply", this::applyChanges);
-        this.closeButton = new FlatButtonWidget(new Dim2i(this.width - 73, this.height - 30, 65, 20), "Close", this::onClose);
+        this.closeButton = new FlatButtonWidget(new Dim2i(10, y, 64, 20), "Close", this::onClose);
+        this.applyButton = new FlatButtonWidget(new Dim2i(78, y, 64, 20), "Apply", this::applyChanges);
+        this.undoButton = new FlatButtonWidget(new Dim2i(146, y, 64, 20), "Undo", this::undoChanges);
 
         this.children.add(this.undoButton);
         this.children.add(this.applyButton);
@@ -110,7 +110,7 @@ public class SodiumOptionsGUI extends Screen {
         }
     }
 
-    private void rebuildGUIOptions() {
+    private int rebuildGUIOptions() {
         int x = 10;
         int y = 28;
 
@@ -130,6 +130,7 @@ public class SodiumOptionsGUI extends Screen {
             // Add padding beneath each option group
             y += 4;
         }
+        return y;
     }
 
     @Override


TheMadHau5:
What's wrong with using `this.height - 30` and leaving `rebuildGUIOptions` with the `void` return type? (I just want to know your reasoning behind this)

sammko:
@TheMadHau5: The reason is the same as why I did it in the first place: 1440p (or higher) on big screens. Here is a full screenshot of my game:
image
The bottom left corner is still pretty far away from the rest of the controls. I just like having them all in one place, not scattered all over.

kittokin:
I agree that having the button on the other side is quite inconvinient.

TwinShards:
Something i would suggest is moving the options and the apply/close buttons to the center like we are already use to in vanilla/optifine settings.
Of course that is all up to jelly to decide which is the best.

deralmas:
> Something i would suggest is moving the options and the apply/close buttons to the center like we are already use to in vanilla/optifine settings.
> Of course that is all up to jelly to decide which is the best.

Yeah, I think that having it centered might be a good idea, as IMO the current approach leaves a lot of whitespace.
Maybe there could be added some type of config option to revert to the old style, if someone likes it best that way.

DragonEggBedrockBreaking:
This may be a bit complicated, but maybe this could be configurable, whether to put it on the right or on the left, so that the user decides

TheMadHau5:
Now that I look at this PR again, wouldn't it be a better idea to just scale the UI, so that there won't be much of a problem anymore?

deralmas:
> Now that I look at this PR again, wouldn't it be a better idea to just scale the UI, so that there won't be much of a problem anymore?

Mhh, actually that's a pretty good idea. Would it look nice, though?

LucilleTea:
> 
> 
> Now that I look at this PR again, wouldn't it be a better idea to just scale the UI, so that there won't be much of a problem anymore?

The GUI is scaled according to the GUI Scale setting.

deralmas:
> > Now that I look at this PR again, wouldn't it be a better idea to just scale the UI, so that there won't be much of a problem anymore?
> 
> The GUI is scaled according to the GUI Scale setting.

I think they're proposing to make the GUI fit the whole screen, not just a quarter of it.

TheMadHau5:
> I think they're proposing to make the GUI fit the whole screen, not just a quarter of it.

Exactly. The GUI will be scaled, but _automatically_ (i.e. without user intervention) for convenience. 
I also suggest that this "stack" on the GUI Scale so that it can be larger or smaller than the auto-detected size.

LucilleTea:
Minecraft's GUI scale is already automatic, until the user overrides it. The user in this screenshot has done so. I don't think it makes any sense for our options screen to single itself out as not using the standard GUI scale, or scaling itself further on top? It's just a GUI like any other.

TheMadHau5:
I wasn't aware that they had manually set the GUI Scale. Apologies for any confusion.
In that case, their PR seems reasonable as it just moves the buttons up.

DragonEggBedrockBreaking:
Since 1.16.x/dev is no longer worked on, I cherry-picked the commit over to a new 1.16.x/next version (had to resolve conflicts), and it works fine. Since 1.16.x/dev is no longer worked on as far as i can see, would it be necessary to edit this pr somehow to move the feature to 1.16.x/next? My way of resolving commits may not be the best, but you can see [how i resolved conflicts](https://github.com/DragonEggBedrockBreaking/sodium-fabric/commit/edc5042038ff0fb8646e28da33fa9ed43ff9d473) in case you want a way that I have already tested that should work (note the word 'should').

sammko:
Thanks for the heads-up. I rebased onto `1.16.x/next`. @DragonEggBedrockBreaking, you resolved the conflicts correctly, as it seems we've ended up with the same result.

DragonEggBedrockBreaking:
> you resolved the conflicts correctly, as it seems we've ended up with the same result.

That's nice to hear considering I don't actually know java

TheMadHau5:
(Just a tangent)
If 1.16.x/next becomes the default, will Indium get integrated into Sodium?

DragonEggBedrockBreaking:
> (Just a tangent)
> If 1.16.x/next becomes the default, will Indium get integrated into Sodium?

I believe there are some plans to merge Indium into Sodium, but there is no ETA, and I believe comp500 wants to rewrite Indium before that happens.

douira:
There should maybe also be some new discussions around if this is a wanted change in the first place or not. The last time this was discussed was pretty long ago.

PR #59, opened 2020-07-06
Attempt fix for IndexOutOfBoundsException while rendering boats

iczero:
Note: The fix is bad.

The exception occurs when the `BufferBuilder` instance has a `VertexFormat` of size less than 32. `BufferBuilder.vertexQuad` writes at least 32 bytes to the internal `ByteBuffer`, but the buffer is grown by the size of the vertex format, so if the wrong vertex format is provided, a crash will occur. If the "Use memory intrinsics" operation is enabled, a JVM crash will occur instead.

The vanilla `BufferBuilder` implementation checks the field `field_21594` (labeled `canUseVertexPath`) before writing to the internal buffer. Adding the same check seems to have fixed the issue.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/mixin/buffers/MixinBufferBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/mixin/buffers/MixinBufferBuilder.java
@@ -67,6 +67,16 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer implem
     public void vertexQuad(float x, float y, float z, int color, float u, float v, int overlay, int light, int normal) {
         if (this.colorFixed) {
             throw new IllegalStateException();
+        } else if (!this.field_21594) {
+            int r = ColorARGB.unpackRed(color);
+            int g = ColorARGB.unpackGreen(color);
+            int b = ColorARGB.unpackBlue(color);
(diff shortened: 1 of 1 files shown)


Sturmlilie:
How does this check relate to [this one in `.vertex()`](https://github.com/jellysquid3/sodium-fabric/blob/86aad51df689e463959ae6262f8f67aa76f31343/src/main/java/me/jellysquid/mods/sodium/mixin/buffers/MixinBufferBuilder.java#L57) and [this one in `quad()`](https://github.com/jellysquid3/sodium-fabric/blob/86aad51df689e463959ae6262f8f67aa76f31343/src/main/java/me/jellysquid/mods/sodium/mixin/buffers/MixinBufferBuilder.java#L237)?

iczero:
I do not know. It seems that it was missing.

jellysquid3:
The solution here is actually perfectly fine, and looks to be the right thing to do. I've incorporated it into a cleanup of the interfaces [here](https://github.com/jellysquid3/sodium-fabric/blob/411b953b14ddedcf8e95a71e4171c47b04702fc2/src/main/java/me/jellysquid/mods/sodium/mixin/buffers/MixinBufferBuilder.java#L139). Thanks.

PR #62, opened 2020-07-07
Mod overrides for config values

Earthcomputer:
Same as https://github.com/jellysquid3/lithium-fabric/pull/74 but for Sodium. It works in exactly the same way, except `sodium:options` is used instead of `lithium:options` in the `fabric.mod.json`.


mrmangohands:
Seems this functionality was merged manually forever ago in cdfc2eb74ca46759013e0f07de34ae19b28866a3 :P

PR #66, opened 2020-07-08
new: Item rendering optimization

Leo40Git:
Apparently `ItemRenderer.renderBakedItemModel` creates a new `Random` instance every time it's called, I.E. *every time an item is drawn on-screen or in the world*. Which is a lot. Thanks, Mojang.

Also, there's a call to `Direction.values()` there.

Props to `vini2003#0875` on the Fabricord for putting this stupidity on blast.


jellysquid3:
I [took a shot](https://github.com/jellysquid3/sodium-fabric/blob/1.16.x/dev/src/main/java/me/jellysquid/mods/sodium/mixin/item_rendering/MixinItemRenderer.java#L34-L55) at optimizing item rendering in general, and have included the fixes to avoid the object allocations this PR avoids.

Leo40Git:
fair enough, my stuff barely counted for optimizations anyway

PR #80, opened 2020-07-12
fix: clear cloud framebuffer when changing cloud setting with fabulou…

### LINKED ISSUE #79: Clouds are badly broken with Fabulous graphics
### Expected Behavior
With Fabulous graphics settings, clouds should render behind other objects and stop rendering when clouds are disabled.

### Actual Behavior
Clouds render in front of all other objects and, while they stop moving, continue to be displayed when clouds are disabled while in-game.

### Reproduction Steps
1. Enable Fabulous graphics and clouds.
2. Load a world.

### Attachments
Clouds On
Clouds Off

### System Information
- Java Version: OpenJDK 14.0.1
- CPU: Ryzen 5 3600
- GPU: RX 5700 (reproduced on Quadro 8000)

LucilleTea:
…s graphics on

fixes #79


jellysquid3:
I really do not like this hack (hard to believe even vanilla does it), but it's good enough for now.

PR #81, opened 2020-07-13
Fix: Remove Math.round() from Entity Distance Slider

### LINKED ISSUE #65: Entity Distance does not save
### Expected Behavior
When altering entity distance you set it to what you want and click apply

### Actual Behavior
It just resets to 100% each time

### Reproduction Steps
Not exactly causing a crash just what I noticed

1. Go into settings
2. Alter entity distance, click apply and leave settings
3. go back into settings and see 100%

### System Information

- Java Version: [1.8.0_51 64-bit]
- CPU: [Intel i5 5300 U @2.30GHz]
- GPU: [Intel HD Graphics 5500]
- DirectX: [Version 12]
- RAM: [8 GB]

amnotbananaama:
fixes #65 .

Note: I have no Java programming experience or training, I just thought I would open the PR to fix the issue. Are there any code quality concerns with this PR? I tested this and it seemed to save the option fine. I'm not sure what the implications are of not rounding here.


jellysquid3:
The reasoning for this was to make sure that floating point errors wouldn't sneak their way into the config file, but it doesn't work in the reverse direction when saving. I think it's _probably_ fine anyways, as we're only dividing the slider's integer value by a constant. No way to avoid the precision loss.

PR #100, opened 2020-07-17
Translations for Sodium

AMereBagatelle:
This PR makes the following changes:

- Adds the `en_us.json`.  This file previously was named as `en_US.lang`, however the Fabric API Resource Loader accepts JSON instead.

- Adds and includes the Fabric API Resource Loader when Sodium is built.

- Adds translations for every string literal that I could find within the `gui` package.

I did not make translations for logging messages or crash reports.  Anything that is not shown to the user in game has not been translated.  Please let me know if translations would be wanted for those areas.  Also just let me know if there are any translations missing, and I'll add them.


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
```diff
@@ -197,7 +198,7 @@ private void renderOptionTooltip(MatrixStack matrixStack, ControlElement<?> elem
         OptionImpact impact = option.getImpact();
 
         if (impact != null) {
-            tooltip.add(new LiteralText(Formatting.GRAY + "Performance Impact: " + impact.toDisplayString()));
+            tooltip.add(new TranslatableText("options.tooltips.performance_impact_string", Formatting.GRAY, impact.toDisplayString()));
```
Gegy: Might make sense to not include the formatting in the translation string itself, and rather set the style on the `Text` directly

AMereBagatelle: That's true, I will change that.  Thanks!

AMereBagatelle: Just tried it, turns out that that breaks the special formatting on the `impact.toDisplayString()` call so I'm not going to include that here.  (Could append that call, but imo that doesn't make it look nearly as nice to read)

juancarloscp52:
I think we should add translation for the options defined in SodiumGameOptions.java. Enums ChunkRendererBackendOption, GraphicsQuality and LightingQuality.

jellysquid3 (changes requested):
We should avoid using the global namespace here. `options.tooltips...` should be changed to `sodium.option.NAME.tooltip`, and `options.names...` should be changed to `options.option.NAME.name`. 

If we can also try to use vanilla language keys where possible for things like V-Sync, Render Distance, etc, that would save a ton of time translating here.

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
```diff
@@ -97,9 +98,9 @@ private static ChunkRendererBackendOption pickBestBackend() {
     }
 
     public enum GraphicsQuality implements TextProvider {
-        DEFAULT("Default"),
-        FANCY("Fancy"),
-        FAST("Fast");
+        DEFAULT(I18n.translate("generator.default")),
```
jellysquid3: Using `I18n#translate` here means that the translated strings will not change until the game is fully restarted, since the string will be cached and re-used. It should instead be moved into the TextProvider#getLocalizedName method below.

jellysquid3 (changes requested):
Looks good aside from one small thing.

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
AMereBagatelle: I see.  Will fix then :+1:

src/main/java/me/jellysquid/mods/sodium/client/gui/options/OptionImpact.java
```diff
@@ -1,13 +1,14 @@
...
+import net.minecraft.client.resource.language.I18n;
 import net.minecraft.util.Formatting;
 
 public enum OptionImpact {
-    LOW(Formatting.GREEN, "Low"),
-    MEDIUM(Formatting.YELLOW, "Medium"),
-    HIGH(Formatting.GOLD, "High"),
-    EXTREME(Formatting.RED, "Extreme"),
-    VARIES(Formatting.WHITE, "Varies");
+    LOW(Formatting.GREEN, I18n.translate("sodium.option_impact.low")),
```
creeper123123321, line 7: ~~Will this affect in-game language changing? (Using translated strings in enum constructor)~~ already fixed

PR #104, opened 2020-07-18
Change: Remove Unused Translation String

amnotbananaama:
`sodium.options.v_sync.name` is unused since we're using `options.vsync` [here](https://github.com/jellysquid3/sodium-fabric/blob/f8c3a67822ca594ee934d5da0fd12097f4daf5d1/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java#L104). Alternatively, if "Vsync" is preferred to Vanilla's "Use Vsync" we should swap the string out with a custom one.


Madis0:
Also worth noting that other translation strings refer to "V-Sync", a slightly different stylization.

amnotbananaama:
Closing as this has been translated frequently enough that the proper fix is to swap `options.vsync` with `sodium.options.v_sync.name`.

Issue #113, opened 2020-07-18
Scroll to change sliders


luximus-hunter:
**Is your feature request related to a problem? Please describe.**
It's not a problem, more a convenience.

**Describe the solution you'd like**
The ability to use my scroll wheel to change the value of sliders.

**Describe alternatives you've considered**
x

**Additional context**
x

FlashyReese:
I'm looking on how to implement this now. I'm running into an issue where the widget itself isn't registering mouse scrolled.

PR #115, opened 2020-07-19
Add Mouse Scroll Support for SliderControl

FlashyReese:
Just a quality of life change to address #113 

- Added Mouse Scroll Support for SliderControl.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
@@ -272,6 +272,15 @@ public class SodiumOptionsGUI extends Screen {
         return super.keyPressed(keyCode, scanCode, modifiers);
     }
 
+
+    @Override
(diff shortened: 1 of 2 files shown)


jellysquid3:
My biggest concern with this is that in the near future, the list of options will grow enough to necessitate a scroll bar in order to cover everything. This would create overlapping scroll areas and make things rather non-intuitive when users want to scroll down the list, since random sliders would get adjusted instead. Need to think on it.

luximus-hunter:
It could prioritize the slider scroll over the vertical scrolling. But that's just my mind speaking 🤷‍♂️

FlashyReese:
It would depend on how the widget list box is implemented. It shouldn't be a problem(for now).

blai30:
I agree with Jelly here, UI/UX is important. I wouldn't want to be scrolling through the settings only to accidentally change something. If anything I would rather it be shift + scrollwheel.

FlashyReese:
I forgot to point out that this feature only scrolls while within sliderbounds. I will be looking into implement a scrollable list(as the list of settings grow) to address those concerns.

PR #124, opened 2020-07-20
Fix translation & add Chinese.

KrisCris:


KrisCris:
- Fix all missing translation.
- Add a translation for zh_cn based on that.

jellysquid3:
Translations are being added through [OneSky](jellysquid.oneskyapp.com/). You should contribute translated strings there. In the future, Sodium will automatically download these translations when building the mod.

KrisCris:
Ah, I didn't notice that.

KrisCris:
But I think some translations are missing in that json file, e.g. off, crosshair, Hotbar.

FlashyReese:
Those translations are included in Minecraft itself. Take a look at #116

Leo40Git:
Those translations are pulled from Minecraft's own translation JSON files, I'm pretty sure. Why reinvent the wheel, after all?

PR #150, opened 2020-07-22
Change layout of settings menu

FlashyReese:
This PR makes the following changes:
- Added Mouse Scroll Support for SliderControl.
- Fixes Missing Translations Strings in SodiumGameOptionPages and SodiumGameOptions
- Added TabControlScrollPaneWidget exclusively for OptionPages
     - Also addresses current issue with localizations
     - Also adds scrollbar support (to address concerns with a growing settings menu)

Addresses the following:
- #113 
- #115 

Preview: https://youtu.be/mKzq4oIqF0s


Madis0:
It seems like the performance impact is no longer displayed in the tooltip? 
Also, for sidebar tabs I'd suggest a blue rectangle around the button, not just a line below.

Titaniumtown:
I think it looks nice!

FlashyReese:
> It seems like the performance impact is no longer displayed in the tooltip?
> Also, for sidebar tabs I'd suggest a blue rectangle around the button, not just a line below.

Performance impact was removed by Jelly on the dev branch.
I will look into that when Jelly moves these mods into an org, she is currently deciding on a name.

FlashyReese:
Superseded by #293

PR #174, opened 2020-07-25
Compatibility with IP

qouteall:
Make every WorldRenderer have a SodiumWorldRenderer.
Put per-frame information into an object and provide interface to switch it.
Add maven publish things.


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ChunkRenderManager.java
```diff
@@ -82,8 +78,23 @@
...
+            chunkRenderLists = new ChunkRenderList[renderPassCount];
+            for (int i = 0; i < this.chunkRenderLists.length; i++) {
+                this.chunkRenderLists[i] = new ChunkRenderList<>();
+            }
+        }
+    }
+
     @SuppressWarnings("unchecked")
-    public ChunkRenderManager(SodiumWorldRenderer renderer, ChunkRenderBackend<T> backend, ClientWorld world, int renderDistance) {
+    public ChunkRenderManager(SodiumWorldRenderer renderer, ChunkRenderBackend<T> backend, BlockRenderPassManager renderPassManager, ClientWorld world, int renderDistance) {
```
jellysquid3: Why was `BlockRenderPassManager renderPassManager` added here? No code seems to use it, and it can be obtained through a call to `ChunkRenderBackend.getRenderPassManager()`

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ChunkRenderManager.java
```diff
@@ -183,7 +190,7 @@ private void addChunkToRenderLists(ChunkRenderContainer<T> render) {
             T state = states[i];
```
jellysquid3, line 190: We should hoist `this.renderContext` into a local variable to help the JVM optimize here (as the field can be null). For example:

```java
RenderContext<?> ctx = this.renderContext;
// ...

if (state != null) {
  ChunkRenderList<T> list = ctx.chunkRenderLists[i];
// ...
```

src/main/java/me/jellysquid/mods/sodium/client/IWorldRenderer.java
```diff
@@ -0,0 +1,7 @@
+package me.jellysquid.mods.sodium.client;
+
+import me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer;
+
+public interface IWorldRenderer {
```
jellysquid3: Avoid using C# style prefixes on interface names. For example, a better name would be `WorldRendererAccess`.

jellysquid3 (changes requested):
Just some minor nitpicks, looks good otherwise. Please see the [Style Guide](https://github.com/jellysquid3/sodium-fabric/blob/1.16.x/dev/CONTRIBUTING.md#code-style) for fixing up some small style issues.

machinesmith42:
Could this be merged?

Issue #219, opened 2020-07-31
Sodium crashes when sodium-options.json is corrupt.


LucilleTea:
### Reproduction Steps
1. Have a corrupt sodium-options.json in the config folder.
2. View sodium's settings or open a world.

### Crash Report File
https://pastebin.com/61ESLENC
https://cdn.discordapp.com/attachments/602805020861267968/737681352349974668/latest.log

### Additional Information
A sample corrupt file:
https://cdn.discordapp.com/attachments/602805020861267968/737956575867699275/sodium-options.json

jellysquid3:
Commit https://github.com/CaffeineMC/sodium-fabric/commit/6f1762fd8844cef131056b9f24927408dd369959 might resolve this.

jellysquid3:
Looks to be fixed in 1.16.x/next.

silversquirl:
This is apparently happening again in 0.3.0, see #862

RandomPersonAmogus:
Still crashing even though Ive deleted sodium-settings.json

MeeniMc:
As of 0.4.11 (presumably earlier) a correct error screen is shown and proposes to reset the problematic option files. Issue is resolved.

pajicadvance:
Not sure when exactly this was fixed either, but Sodium 0.4.10 now shows this screen and the game can proceed to run normally. Thanks for the bump.

![Screenshot_4](https://github.com/CaffeineMC/sodium-fabric/assets/32738446/8425b744-41e9-48da-9071-3cb1ebe8a8ba)

PR #234, opened 2020-08-04
Automatically catching and fixing invalid option files.  Fix for #219

AMereBagatelle:
This PR adds the following:

- One catch clause to check if the JsonSyntaxException is thrown and if so revert to the defaults for Sodium's options as the file is invalid.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
@@ -3,6 +3,7 @@ package me.jellysquid.mods.sodium.client.gui;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonSyntaxException;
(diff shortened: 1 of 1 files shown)


FlashyReese (approved):

2No2Name:
Does that delete the whole file? (Or does it just use the defaults?) I would at least rename it to prevent the users from losing their settings permanently after introducing a syntax error

AMereBagatelle:
It won't actually overwrite the file unless changes are applied in the menu, however I don't expect most users to be editing the file directly and introducing such errors.  (aka there isn't a case where you do need to edit it directly)
It does reset everything to the defaults.

ishland (approved):

2No2Name (approved):

bytzo (approved):

jellysquid3:
I think this is treating the symptoms of our config file's serialization problem rather than the actual underlying issue. Even if we can't figure out what's causing that (which is probably that it isn't being fsync'd), we could work around it with atomic filesystem operations. This could be done by writing the config to a temporary file, then replacing the old config with the new one using a rename operation.

PR #256, opened 2020-08-14
Remove a few unnecessary lines from MixinBufferBuilder

dyc3:
In the modified functions, since `i` goes out of scope and is not used after the function completes, it's unnecessary to increment `i`.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
@@ -104,7 +104,6 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer
         i += 4;
 
         buffer.putInt(i, light);
-        i += 4;
     }
 
     private void vertexParticleUnsafe(float x, float y, float z, float u, float v, int color, int light) {
@@ -130,7 +129,6 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer
         i += 4;
 
         unsafe.putInt(i, light);
-        i += 4;
     }
 
     @Override
@@ -357,7 +355,6 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer
         i += 4;
 
         buffer.putInt(i, light);
-        i += 4;
     }
 
     private void vertexGlyphUnsafe(float x, float y, float z, int color, float u, float v, int light) {
@@ -383,6 +380,5 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer
         i += 4;
 
         unsafe.putInt(i, light);
-        i += 4;
     }
 }


jellysquid3:
These lines exist so that mods injecting into the methods can capture the write pointer without making assumptions. This might be extremely unlikely, but that is the reasoning. The JVM should optimize them away if nothing captures them after the fact.

dyc3:
Ah, I see. In that case, should `vertexQuadSafe` and `vertexQuadUnsafe` increment the pointer at the end as well?

deeprobin:
> 
> 
> These lines exist so that mods injecting into the methods can capture the write pointer without making assumptions. This might be extremely unlikely, but that is the reasoning. The JVM should optimize them away if nothing captures them after the fact.

Aren't there other possibilities with the help of mixins or can you somehow have them injected manually? And you should ask yourself why of all things a mod should inject at such a place.

src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
```diff
@@ -130,7 +129,6 @@ private void vertexParticleUnsafe(float x, float y, float z, float u, float v, i
         i += 4;
 
         unsafe.putInt(i, light);
-        i += 4;
```
deeprobin, line 133: Here one `@At("RETURN")` injection should be sufficient.

src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
```diff
@@ -383,6 +380,5 @@ private void vertexGlyphUnsafe(float x, float y, float z, int color, float u, fl
...
```
deeprobin, line 386: Here one `@At("RETURN")` injection should be sufficient.

src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
```diff
@@ -357,7 +355,6 @@ private void vertexGlyphSafe(float x, float y, float z, int color, float u, floa
         i += 4;
 
         buffer.putInt(i, light);
-        i += 4;
```
deeprobin, line 360: Here one `@At("RETURN")` injection should be sufficient.

src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/intrinsics/MixinBufferBuilder.java
```diff
@@ -104,7 +104,6 @@ private void vertexParticleSafe(float x, float y, float z, float u, float v, int
...
```
deeprobin, line 107: Here one `@At("RETURN")` injection should be sufficient.

PR #258, opened 2020-08-15
refactor: reduce calls to nextInt

dyc3:
We can reduce the number of calls to `random.nextInt` here a little bit.


jaskarth:
This was actually [tested and then reverted](https://github.com/jellysquid3/sodium-fabric/commit/8b3082f4fff2e09410ed7056546202b3418f16a0), as the two functionalities have different behaviors. The vanilla behavior uses a triangle distribution while the change uses a uniform distribution. Plus, `nextInt` isn't very computationally heavy (1 multiplication, addition, modulo, xor, and bit shift) so I doubt this would change much either way :P

dyc3:
Ah, I see. There should probably be a comment about that there. Closing

PR #250, opened 2020-08-13
Updated to Minecraft 1.16.3

ishland:
It compiles and works just fine. Did not do further testing.

PR #261, opened 2020-08-16
Update to 1.16.2

Cuhnt:
Just a quick conversion to 1.16.2. Nothing added, just fixed the errors after changing the version.


ThexXTURBOXx:
This PR has some issues at edges of blocks:
https://prnt.sc/u0e4pe
#250 also has the same issue.
The issue only occurrs when using a resource pack (I am using Sphax atm)

Cuhnt:
I'm blind, I didn't notice that #250 existed, I'll close this.

PR #375, opened 2020-11-03
change: Change GitHub Actions to JDK 8

FlashyReese:
As said in title.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,10 +8,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


ayangd:
Done in jellysquid3#276.

amnotbananaama:
> Done in 276.  

That PR only affected the development branch (`1.16.x/dev`), not the stable branch (`1.16.x/stable`). At the time, development and new commits only happened on `1.16.x/dev`, with `1.16.x/stable` only being used for publishing releases. That's shifted a bit given time constraints and the fact that things are a bit in flux right now with Sodium

PR #378, opened 2020-11-04
fix: Use maximum GUI scale as GUI scale limit

### LINKED ISSUE #259: GUI size limited to 4

# Workaround

bytzo:
_Pull request closed due to a mistake while attempting to change the target branch._

---

Instead of using a fixed value of 4 for the maximum GUI scale in the Sodium options page, the largest calculated GUI scale (auto) will be used. This behavior is especially useful on high-dpi displays, as they require significantly larger GUI scales to match the scale of low-dpi displays. The small change allows the maximum GUI scale to reach higher values than 4 on larger displays, matching vanilla behavior and resolving #259.

**Edit:** Issue will not link using "resolving". Adding: Fixes #259


bytzo:
This got very messed up while changing base, closing PR.

PR #379, opened 2020-11-04
fix: Update project homepage to curseforge

1oglop1:
Hi, this is a tiny change but quite useful, unless project homepage points to your website intentionally.


jellysquid3:
My project homepage now points to the projects. In the future, this will probably point to a dedicated website for the mods.

PR #390, opened 2020-11-10
fix: Use auto GUI scale as GUI scale limit

### LINKED ISSUE #259: GUI size limited to 4

bytzo:
**EDIT:**
No longer valid, as new default branch is [`1.16.x/next`](https://github.com/jellysquid3/sodium-fabric/tree/1.16.x/next) instead of [`1.16.x/dev`](https://github.com/jellysquid3/sodium-fabric/tree/1.16.x/dev).

---

Instead of using a fixed value of 4 for the maximum GUI scale in the Sodium options page, the auto GUI scale (largest calculated size) will be used. Allowing the GUI scale to extend above 4 is helpful on high-dpi displays, where significantly larger GUI scales are necessary to match the physical size of equivalent low-dpi displays. This small change matches vanilla behavior and fixes #259.

_Note: Correction for #378, target branch is now [`1.16.x/dev`](https://github.com/jellysquid3/sodium-fabric/tree/1.16.x/dev)._

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -74,7 +74,9 @@ public class SodiumGameOptionPages {
                 .add(OptionImpl.createBuilder(int.class, vanillaOpts)
                         .setName(I18n.translate("options.guiScale"))
                         .setTooltip(I18n.translate("sodium.options.gui_scale.tooltip"))
-                        .setControl(option -> new SliderControl(option, 0, 4, 1, ControlValueFormatter.guiScale()))
+                        .setControl(option -> new SliderControl(option, 0, MinecraftClient.getInstance().getWindow().
+                                calculateScaleFactor(0, MinecraftClient.getInstance().forcesUnicodeFont()), 1,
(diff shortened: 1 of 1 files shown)


MeeniMc:
Why did this never get merged? Do you need help wrangling with github cherry-picking? I can help.

DragonEggBedrockBreaking:
> Why did this never get merged? Do you need help wrangling with github cherry-picking? I can help.

[This is the line in sodium next as far as I can see.](https://github.com/CaffeineMC/sodium-fabric/blob/b319e3d75d3d3a0db956fbe1bdbe49fb07e702c2/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java#L71)

bytzo:
> Why did this never get merged? Do you need help wrangling with github cherry-picking? I can help.

Sorry for this extremely late reply. This was not merged because I realized the solution proposed was not ideal, as it caused issues when changing window size while in the Sodium options menu. Since the control elements are only created once when opening the Sodium options menu, the user must exit and enter the Sodium options menu each time the window size is changed in order to see the updated slider values. Because there is currently no good solution for updating controls on window resize, I did not bother creating a new pull request for [`1.16.x/next`](https://github.com/jellysquid3/sodium-fabric/tree/1.16.x/next).

PR #401, opened 2020-11-21
Center title, logo, and shields in README.md

proudmuslim-dev:
This is a slight cosmetic change to the README.md file of this repository, centering the title, logo, and shields at the top. You can preview these changes [here](https://github.com/proudmuslim-dev/sodium-fabric/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Sodium (for Fabric)
 ![GitHub license](https://img.shields.io/github/license/jellysquid3/sodium-fabric.svg)
-![GitHub issues](https://img.shields.io/github/issues/jellysquid3/sodium-fabric.svg)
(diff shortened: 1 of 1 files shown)


ghost:
In my opinion,if the title,logo and shields are going to be centered,the text should be too,like the titles of the different paragraphs should be centered and the text inside the paragraph should too.

proudmuslim-dev:
> In my opinion,if the title,logo and shields are going to be centered,the text should be too,like the titles of the different paragraphs should be centered and the text inside the paragraph should too.

That would look pretty messy imo, however if the maintainers want that they can do so

jellysquid3:
This seems to be a matter of subjective taste as it doesn't improve legibility, etc.

Issue #415, opened 2020-12-09
guardian beam visual bug


lolcuslol:
### Expected Behavior
Guardian beam is rendered as expected

### Actual Behavior
guardian beam is replaced by a line of bubbles

### Reproduction Steps
It happens only in some of my worlds and isn't connected to any datapack or mod because the only one i have installed is Sodium.
### Attachments

![2020-12-09_20 12 51](https://user-images.githubusercontent.com/75753130/101676120-06799b80-3a5b-11eb-9817-f9c0a533743a.png)

here is a screenshot of this glitch

Teh404Gal:
are you using the 1.16.3 version of sodium or the discord #testers snapshot
since in the latter version 
i still see the lazer 
![2020-12-10_13 37 38](https://user-images.githubusercontent.com/62482190/101774103-efd05480-3aed-11eb-9679-e0ecc9a54e6f.png)
![2020-12-10_13 44 49](https://user-images.githubusercontent.com/62482190/101774152-feb70700-3aed-11eb-8c94-d2dd011381ab.png)

Teh404Gal:
ignore the low fps im running the game on battery power

lolcuslol:
Im using the 1.16.3 version from curseforge. But that can't be the problem because when i tried it in my other world it looked normal like in your screenshot. And that's why i think it's something world specific.

Teh404Gal:
seems weird
its probably a world dependent bug
if you want you to, can send me the world or the seed for it in discord so i can check if my client also reacts the same
also last question but did you load this world in multiple version?

lolcuslol:
Well the seed is 364819863 the world type is large biomes.
 and i can send you the world save but i dont know if discord can send files that big its about 1,1GB 
and yes i did load it in multiple versions it was originally generated in 1.16.1 but I'm not sure about that.

Teh404Gal:
well it seems the seed its self isnt the issue
if you want you can send me the world but before you do that
click the edit button while in the world selection menu
click on optimize after a while it will finish then click make a backup
and then send me the world so i can examine it
MegaloStudio#7838

Teh404Gal:
you can use google drive/ one drive/ nextcloud to send the world

lolcuslol:
Ok I'll send you the google drive link as soon as i can 
And thanks for being so helpful I really appreciate it.

mrgrim:
This is a vanilla bug related to world age: https://bugs.mojang.com/browse/MC-165595

[...]

PR #419, opened 2020-12-11
Fix guardian beams disappearing after long periods of time in a world (#415)

AMereBagatelle:
I saw how easy the fix was, so I couldn't help myself :P

Adds one mixin, to use the time of day instead of the world time when rendering the guardian beams.


2No2Name:
https://github.com/jellysquid3/sodium-fabric/issues/415 was marked as won't fix and is now marked as external issue. It is not clear whether the authors do not want to invest time in fixing the vanilla issue, or whether a fix is in the scope of sodium.

jellysquid3:
It's more incidental that Sodium fixes a lot of vanilla graphical issues in block rendering because we're already replacing huge swaths of code. This patch requires us to go out of our way to fix what is a logic error in Guardian rendering, and I don't think it's a great fit for the mod as such.

PR #426, opened 2020-12-18
fix: Use block height to check for invalid subchunks and remove unnecessary height check 

### LINKED ISSUE #425: sodium latest commit in 1.16.x/next crashes when joining deep caverns in hypixel skyblock
### Reproduction Steps
use latest commit (as of writting its ef6c9bd in 1.16.x/next) 
join hypixel
play skyblock
reach deep cavers 
crash

### Crash Report File
https://gist.github.com/teh404guy/091d058665cc758e4a354633df804cb6

### Additional Information
well not-enough-crashes
told me its sodium
and every time i join a server after crashing it just crashes immediatly without loading joining or loading the world

bytzo:
This PR addresses two issues brought up in commit [`eb664ec`](https://github.com/jellysquid3/sodium-fabric/commit/eb664ec9bf22428678691f761fa0a6a73c916410) relating to checks for valid subchunks.

In summary, the chunk height (ranging from 0-15) was used to check for valid chunks, when the block height (ranging from 0-255) is expected instead, causing invalid heights (ex. 16 chunk, 256-272 block) to pass the check (0-255 block) incorrectly. This fixes #425. Additionally, one of these checks was removed due to it causing subchunks directly above and below world limits to have no light arrays, making block faces adjacent to these appear black.

See commit messages for more details.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
@@ -454,7 +454,7 @@ public class WorldSlice extends ReusableObject implements BlockRenderView, Biome
     private static ChunkSection getChunkSection(Chunk chunk, ChunkSectionPos pos) {
         ChunkSection section = null;
 
-        if (!World.isHeightInvalid(pos.getY())) {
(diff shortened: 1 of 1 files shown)


jellysquid3:
Thanks for your contribution! This looks good to me. Must've been something that cropped up while splitting the large loop into different methods in that commit.

PR #428, opened 2020-12-19
fix: Set current world to null when unloading a world

### LINKED ISSUE #430: Game crashing when switching servers, or rejoining the same server [Sodium's ''Next'' branch, version 0.1.1-SNAPSHOT}
### Reproduction Steps
Provide information on how to reproduce this game crash. This step is critical for ensuring that your crash can be debugged by other developers. You can either fill in the example below or do something else -- just make sure your instructions are minimal and clear. If applicable, provide a Minecraft save containing a minimal example on how to reproduce this crash as a compressed ZIP file.

There are 2 ways to trigger this
A] Join a Minecraft server, leave, rejoin it and game crashes, 
B] Join a Minecraft server, leave, join a other server and game crashes

### Crash Report File
https://gist.github.com/CheesyTheArtist/34f90f4068612f718cedc5280059e0a1

### Additional Information
If needed, my specs are:
CPU: Intel i5-9300H
GPU: Nvidia GTX 1650 Mobile
RAM: 16GBs of DDR4

bytzo:
This PR adds a single line of code which fixes #430, a crash caused by commit [`ef6c9bd`](https://github.com/jellysquid3/sodium-fabric/commit/ef6c9bd29b412897997f1fe3a26dfe227399baba). This crash occurs when a user leaves a world and joins another in the same session due to the attempted unloading of a world twice.

_(Copied from commit message)_
When the client world is changed, a check is present to decide if the current world should be unloaded or not. Currently, the world is only unloaded if the current world is not null. However, the current world is never set to null when unloading it. This means the check (if world is not null) becomes meaningless after a world is loaded for the first time, allowing the unload method to run whenever the client world changes even if the world is already unloaded. Because of this behavior, an unhandled exception is thrown when the unload method attempts to unload a world that doesn't exist (specifically having no render context). This specific situation can occur when leaving a world and joining another. When the second world is being loaded, the unload method is called even though the first world has already been unloaded, leading to the crash. In order to prevent this from happening, a single line is added after the unload method which sets the current world to null.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -117,6 +117,7 @@ public class SodiumWorldRenderer implements ChunkStatusListener {
 
     private void unloadWorld() {
         GlobalRenderContext.destroyRenderContext(this.world);
+        this.world = null;
 
         if (this.chunkRenderManager != null) {
             this.chunkRenderManager.destroy();


NoComment1105:
I can confirm that this does fix the issue I found.

Teh404Gal:
well this works really well so please merge

jellysquid3:
Fixed with 7071f3a.

Issue #15, opened 2020-05-12
Incompatibility with Immersive Portals


Neubulae:
https://github.com/qouteall/ImmersivePortalsMod/issues/254

qouteall:
https://github.com/qouteall/ImmersivePortalsMod/issues/271

Jarfeh:
I noticed that you've since closed the pull request under sodium for compatibility. Has the compatibility issue been resolved?

IMS212:
This has been fixed for a long time now.

PR #429, opened 2020-12-19
Compatibility with Immersive Portals in the "next" branch

qouteall:
One thing that you may not like is the "duplicated" vertex shader. Because to use the gl_ClipDistance feature the glsl version must be at least 130. But upgrading the shader file version may cause incompatibility with older devices. So I made it switch to the clipping vertex shader only when IP is present and keep the old-versioned shader.


DragonEggBedrockBreaking:
You forgot to mention:
Fixes #15

srnyx:
@comp500 Could you implement this into your fork?

comp500:
@srnyx No. My change is trivial to merge into qouteall's fork, just [edit the single line](https://github.com/comp500/sodium-fabric/commit/8cbe9ebc04df818b82b8af218c853c50fef64d89) yourself.

machinesmith42:
@srnyx  my build contains that fix

Issue #433, opened 2020-12-21
Add option to change the load distance function to a circle


cavallium:
**Is your feature request related to a problem? Please describe.**
The default chunk load distance loads chunks in a square.
In this way the chunks in the corners are rendered farther from the player than other chunks that are along the x or z axis.
This can affect performance since chunks far from the player are rendered.

**Describe the solution you'd like**
Add an option to change the rectangular chunk load radius function with the circular radius function:
in ```me.jellysquid.mods.sodium.client.world.SodiumChunkManager```, line 157:
```java
if (((x - this.centerX) * (x - this.centerX)) + ((z - this.centerZ) * (z - this.centerZ)) <= this.radius * this.radius) {
```

A circular region means that fewer chunks will be loaded and rendered. This can affect performance, especially with Amplified maps or maps with a lot of complex blocks to render.

Another improvement is that chunks that are loaded and unloaded when a player moves will be better spreaded across time, since a square render distance loads a line of chunks (1 * render distance) at time, but a circular one loads less chunks at the same time.

**Describe alternatives you've considered**
An alternative is to always use the circular radius function, without adding an option, or leaving the chunk radius function as it is now.

**Additional informations**
This will also remove the two Math.abs calls from the function, removing two comparations and therefore speeding up the function.
I didn't use the Math.pow function since it's known to be much slower than simply multiplying the same value two times.

altrisi:
There is no need to do that, since sodium already doesn't render chunks that are inside the (circular) fog AFAIK (although they are still loaded).

cavallium:
I closed the issue because I noticed in when I tested the change.

PR #435, opened 2020-12-22
Add render distance formula option

cavallium:
Implements #433.
This option reduces the amount of loaded chunks by 20%.
A circular area is always 20% smaller than a square with the width equal to the diameter of that circle.


Cavallium:
I'm not sure about adding REQUIRES_RENDERER_RELOAD flag to the option, since it only affects loading of new chunks and not the already loaded ones.

cavallium:
Also the description could be improved.

PR #453, opened 2020-12-29
Update Gradle + Update fabric.mod.json (Cancelled)

Kichura:
A typo has been spotted via the main json where the "an free" string can be seen instead of "a free", The compiling system is also capable of running 6.6.1 however i am unaware if a newer version than that can break something.

One more thing: The mod should warn the user of using thallium because it is a mod that basically fakes the FPS, And thous we don't like this at all and want to prevent this in future.


Kichura:
*Some additional changes may be included to correct some spelling texts.

Kichura:
Cancelling pull request as too many commit mistakes were made.

PR #472, opened 2021-01-09
Update SodiumClientMod.java

NightHunter56:


amnotbananaama:
Contributions to improve Sodium are always welcome. However, this is a minor PR, and only changes the whitespace inside one file. The change results in the whitespace being inconsistent throughout the file, so I am going to close this as not needed

NightHunter56:
thanks for looking into it

PR #514, opened 2021-02-04
Fix javadoc task

silversquirl:
`./gradlew javadoc` would previously fail, complaining about invalid syntax and tags.
This PR adds javadoc configuration to `build.gradle` in order to fix this issue and make the task succeed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,6 +75,16 @@ task sourcesJar(type: Jar, dependsOn: classes) {
     from sourceSets.main.allSource
 }
 
+javadoc {
+    options {
+        encoding = "UTF-8"
+        charSet = "UTF-8"
+        addStringOption("Xdoclint:none", "-quiet")
+    }
+    classpath = sourceSets.main.compileClasspath
+    include("me/jellysquid/mods/sodium/**")
+}
+
 jar {
     from "LICENSE.txt"
 }


build.gradle
```diff
@@ -75,6 +75,16 @@ task sourcesJar(type: Jar, dependsOn: classes) {
     from sourceSets.main.allSource
 }
 
+javadoc {
+    options {
+        encoding = "UTF-8"
+        charSet = "UTF-8"
+        addStringOption("Xdoclint:none", "-quiet")
```
liach, line 82: This is actually some ancient remnant from I think spongepowered? This is actually wrong, as on command line it becomes like `-Xdoclint:none=-quiet`. You should do it like https://github.com/FabricMC/yarn/blob/288f20b7d5a2521b475ab88e773acaa056bbdbc1/build.gradle#L749
```suggestion
        addBooleanOption "Xdoclint:none", true
```
which is the right way.

Also muting doclint isn't the best. I recommend enabling these doclints https://github.com/FabricMC/yarn/blob/288f20b7d5a2521b475ab88e773acaa056bbdbc1/build.gradle#L746-L748 and adding `failOnError = false` so you can build javadoc while still viewing the warnings from doclint.

PR #519, opened 2021-02-05
Add order-independent transparency for translucent blocks

### LINKED ISSUE #38: Translucent blocks do not render correctly (fixed in Sodium 0.6)
### Expected Behavior
Water will be visible inside glass stained pipes, commonly used in water elevators

### Actual Behavior
Water is not visible through stained glasses, although items can be seen flowing

### Reproduction Steps
Make a water tube, doesn't matter if its flowing or a water elevator type.

### Attachments
https://cdn.discordapp.com/attachments/389371678859460611/726798724206166066/received_707934516451025.png

### System Information
- Java Version: [Java 8 Update 251]
- CPU: [Intel Core i5-7200U]
- GPU: [Intel HD Graphics 620]

silversquirl:
The algorithm used is ["Weighted, Blended Order-Independent Transparency"][wboit], which allows for a fast and simple implementation at a small cost to color accuracy. This should solve all issues of translucent faces not rendering behind others, such as with water behind stained glass or with blocks containing translucent faces, such as slime and honey blocks.

More testing is required to confirm compatibility with limited OpenGL 2.0 implementations, such as those of macOS or older iGPUs, as well as to determine the impact on performance. **This PR should not be merged until this testing has been done and any issues addressed.**

Fixes #38

[wboit]: https://casual-effects.blogspot.com/2015/03/implemented-weighted-blended-order.html

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gl/TranslucencyFramebuffer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gl/TranslucencyFramebuffer.java
@@ -1,0 +1,162 @@
+package me.jellysquid.mods.sodium.client.gl;
+
+import com.mojang.blaze3d.platform.GlStateManager;
+import com.mojang.blaze3d.platform.FramebufferInfo;
+import com.mojang.blaze3d.systems.RenderSystem;
+import me.jellysquid.mods.sodium.client.gl.shader.GlProgram;
+import me.jellysquid.mods.sodium.client.gl.shader.ShaderConstants;
+import me.jellysquid.mods.sodium.client.gl.shader.ShaderLoader;
+import me.jellysquid.mods.sodium.client.gl.shader.ShaderType;
+import net.minecraft.client.gl.Framebuffer;
+import net.minecraft.client.texture.TextureUtil;
+import net.minecraft.util.Identifier;
+import org.lwjgl.opengl.GL20;
+import org.lwjgl.opengl.ARBTextureFloat;
+import org.lwjgl.system.MemoryUtil;
+
+import java.nio.IntBuffer;
+
+public class TranslucencyFramebuffer extends Framebuffer {
+    private static GlProgram clearProgram = null;
+
+    private int accumAttachment;
+    private int revealAttachment;
+    private int depthAttachment;
+
+    public TranslucencyFramebuffer(int width, int height, boolean useDepth, boolean getError) {
+        super(width, height, useDepth, getError);
+        this.accumAttachment = -1;
+        this.revealAttachment = -1;
+        this.depthAttachment = -1;
+    }
+
+    @Override
+    public void delete() {
+        super.delete();
+        if (this.accumAttachment > -1) {
+            TextureUtil.deleteId(this.accumAttachment);
+            this.accumAttachment = -1;
+        }
+        if (this.revealAttachment > -1) {
+            TextureUtil.deleteId(this.revealAttachment);
+            this.revealAttachment = -1;
+        }
+        if (this.depthAttachment > -1) {
+            TextureUtil.deleteId(this.depthAttachment);
+            this.depthAttachment = -1;
+        }
+    }
+
+    @Override
+    public void initFbo(int width, int height, boolean getError) {
+        RenderSystem.assertThread(RenderSystem::isOnRenderThreadOrInit);
+        this.viewportWidth = width;
+        this.viewportHeight = height;
+        this.textureWidth = width;
+        this.textureHeight = height;
+
+        if (this.useDepthAttachment) {
+            this.depthAttachment = TextureUtil.generateId();
+            GlStateManager.bindTexture(this.depthAttachment);
+            GlStateManager.texImage2D(GL20.GL_TEXTURE_2D, 0, GL20.GL_DEPTH_COMPONENT, width, height, 0, GL20.GL_DEPTH_COMPONENT, GL20.GL_FLOAT, null);
+        }
+
+        this.accumAttachment = TextureUtil.generateId();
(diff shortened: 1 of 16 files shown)


Genau6502:
Any luck with turning up the opacity of honey/slime blocks?

pwouik:
transparency blending isn't correct,every layer seem to have the same factor, and we cant clearly see depth

MeeniMc:
There is the following defect with culling the back-face of the 'targeted block' wire mesh: what we are looking at is a tower of white stained glass (6 tall), over white concrete. The first block is targeted, hence has the selected block wire-mesh around it rendered: 



Otherwise the improvement in correctness is impressive. I have found that water-bubble columns in stained glass (as found in bubblevators) work properly, and, unlike reported earlier, layered multiplicative blending does work properly for me. 
(AMD Rx580, MultiDraw 4.3)

MeeniMc:
Test setup for transparency blending:


(rendered with vk patch

Vanilla | vktec:oit | Sodium
--- | --- | ---
Smooth 'fog' effect | Fog effect works, but edges are more defined, are you using additive blending instead of multiplicative? | Completely broken
 |  |  
 |  | 
--- | --- | ---
 |   | 
Bubblevator, water renders through glass | water AND BUBBLES render through glass | water renders incorrectly through glass, sometimes not at all




Slime blocks in front of honey blocks behind stained glass work properly (although again blending appears to be different than vanilla, could again be explained by a difference in the blending operator).

MeeniMc:
The bug where multiple layers of stained glass are not rendered correctly across chunk boundaries is also fixed. In prior versions of sodium, looking through multiple horizontal layers of glass (separated by 1 air layer inbetween), when looking from the bottom (looking towards the sky) would render only one layer for blocks outside the current chunk. This is now rendering correctly. 



Overall, the improvement in correctness is MASSIVE. Performance drop is negligible with a 256px resource pack. It is more substantial with 16px textures.

pwouik:
transparent blocks behind have too much weight compared to blocks in front
with a lot of transparent blocks it make front blocks blend with further blocks and sometimes blocks behind completely override blocks in front

@Meeni-mc oit doesn't really use a blending operator,but an order independant function

MrJake222:
- Clouds not visible through stained glass and water.
- Water looks weird when seen through another water.

silversquirl:
> Clouds not visible through stained glass and water.

This is a vanilla issue that will not be fixed by this PR. Whether Sodium should fix it is a separate issue.

As for the blending issues, I'm aware of them and will work on a fix after I've fixed OpenGL 2.0 support (which is basically only used on macOS). However, this may take a while as I have a lot going on irl right now.

CalXee:
> * Clouds not visible through stained glass and water.

Does turning on Fabulous graphics fix this?

MrJake222:
> Does turning on Fabulous graphics fix this?

No.

Boobies:
Subchunk boundary issue in current Sodium (vk suggested that I upload it here):

pwouik:
reported here:https://github.com/CaffeineMC/sodium-fabric/issues/38#issuecomment-699734580

MeeniMc:
> Subchunk boundary issue in current Sodium (vk suggested that I upload it here):
> 
>
Are you using this PR, or a sodium release (or sodium-0.1.1 as provided by JellySquid in Nov?). Asking because in my tests this bug was present in prior versions, but is fixed with this PR.

Boobies:
> Are you using this PR, or a sodium release (or sodium-0.1.1 as provided by JellySquid in Nov?). Asking because in my tests this bug was present in prior versions, but is fixed with this PR.

I said "current Sodium," didn't I? That means the latest release. As your tests confirmed already. Anyway, I don't know whether the PR solves the issue or not as I have not tested it but I'm assuming there might be issues regarding it since vktec said that it was new information that I should upload here so he does not forget about it. He probably wants to at least review the issue.

silversquirl:
Ah, I had thought you were observing this behaviour with this PR

silversquirl:
> Resource packs without transparency on normally transparent blocks have some faces visible threw them.

Thanks, this is useful information. I expect it's because those blocks use translucent rendering despite not actually being translucent, meaning they get blended.
Not sure if there's much I can do about it, but hopefully when I fix the other blending issues it'll become less noticeable at least.

MrJake222:
Also water bubbles seem to have issues with rendering.

Heitzenberg:
Do you plan on rebasing this on the current sodium-next, to resolve the merge conflicts?

machinesmith42:
any luck with this?

MeeniMc:
> Do you plan on rebasing this on the current sodium-next, to resolve the merge conflicts?

I have merged this PR on top of PR #581 (Indium/LBG patch), on top of build 352 
https://github.com/MeeniMc/sodium-fabric/commits/pr/519

MeeniMc:
Updated on top of build 359
https://github.com/MeeniMc/sodium-fabric/commits/pr/519+581+598

jellysquid3:
Closing as abandoned. The author (vktec) has talked about not having the time available to continue updating this PR, and many changes have occurred in Sodium since which break this. Furthermore, while WBOIT is a (very fast) approximate function for OIT, it has significant trouble with mostly opaque/translucent surfaces, something which many users reported resulted in major graphical downgrades. Given that the original paper doesn't really acknowledge this outside of saying "don't do it", it's safe to say that this is just an issue with the algorithm and not something we can really correct for.

There are other OIT functions such as [Moment-based OIT (MBOIT)](https://cg.ivd.kit.edu/english/mboit.php), which appear very promising, offering both decent performance (at least relative to sorting all translucent geometry on the CPU) and good rendering quality even in the cases Minecraft often sees. We may at some point want to investigate these options instead.

PR #525, opened 2021-02-12
fix: Animate Only Visible Textures

### LINKED ISSUE #458: Only animate visible textures not animating any textures when enabled
### Expected Behavior
All visible textures should be animated

### Actual Behavior
No textures are animated.

### Reproduction Steps
Turn on Only Animate Visible Textures. Look at a block that has an animated texture (water or lava). Note, the only way to fix this (without coding) is to disable it in the settings. To be honest, there wasn't any performance impact when disabling it.

### Attachments
I am using the latest Next release from Github. d352852 to be exact.

### System Information
- Java Version: [15.0.1 ]
- CPU: [AMD FX 8350]
- GPU: [AMD Radeon RX 480]

Grayray75:
This fixes the feature **Animate Only Visible Textures**.

**This PR should not be merged yet!**

Closes: #458

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuffers.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuffers.java
@@ -1,8 +1,11 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile.buffers;
 
 import me.jellysquid.mods.sodium.client.model.quad.properties.ModelQuadFacing;
+import me.jellysquid.mods.sodium.client.render.chunk.data.ChunkRenderData;
 import me.jellysquid.mods.sodium.client.render.chunk.format.ModelVertexSink;
 
 public interface ChunkModelBuffers {
     ModelVertexSink getSink(ModelQuadFacing facing);
+
+    ChunkRenderData.Builder getRenderData();
 }
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuffers.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuffers.java
@@ -6,6 +6,4 @@ import me.jellysquid.mods.sodium.client.render.chunk.format.ModelVertexSink;
 
 public interface ChunkModelBuffers {
     ModelVertexSink getSink(ModelQuadFacing facing);
-
-    ChunkRenderData.Builder getRenderData();
 }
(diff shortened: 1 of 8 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -138,6 +138,8 @@ private void renderQuad(BlockRenderView world, BlockState state, BlockPos pos, M
             int lm = light.lm[srcIndex];
 
             sink.writeQuad(x, y, z, color, u, v, lm);
+
+            buffers.getRenderData().addSprite(src.getSprite());
```
comp500: This doesn't need to be done for every vertex - you should move this outside the loop. You should also check if the sprite is null first - see this code from the stable branch: https://github.com/CaffeineMC/sodium-fabric/blob/1.16.x/stable/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkMeshBuilder.java#L108

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
```diff
@@ -393,6 +393,8 @@ private void flushQuad(ChunkModelBuffers buffers, ModelQuadView quad, Direction
 
             sink.writeQuad(x, y, z, color, u, v, light);
 
+            buffers.getRenderData().addSprite(quad.getSprite());
```
comp500: See previous comment on BlockRenderer - the same applies here.

comp500 (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
Grayray75: I will try to find a better solution

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuffers.java
```diff
@@ -1,8 +1,11 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile.buffers;
 
 import me.jellysquid.mods.sodium.client.model.quad.properties.ModelQuadFacing;
+import me.jellysquid.mods.sodium.client.render.chunk.data.ChunkRenderData;
 import me.jellysquid.mods.sodium.client.render.chunk.format.ModelVertexSink;
 
 public interface ChunkModelBuffers {
     ModelVertexSink getSink(ModelQuadFacing facing);
+
+    ChunkRenderData.Builder getRenderData();
```
FlashyReese: This could be pass into `ChunkModelVertexTransformer` instead. You can expand `ModelVertexSink#writeQuad` (I'm sure there is a better approach than this) to get `Sprite` and perform the sprite check there.

Grayray75:
Closing this pr because i messed up

PR #526, opened 2021-02-13
fix: Broken block and fluid animations

### LINKED ISSUE #458: Only animate visible textures not animating any textures when enabled

Grayray75:
This fixes broken block and fluid animations when **Animate Only Visible Textures** is enabled.

Closes: #458

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
@@ -9,6 +9,7 @@ import me.jellysquid.mods.sodium.client.model.quad.blender.BiomeColorBlender;
 import me.jellysquid.mods.sodium.client.model.quad.properties.ModelQuadFacing;
 import me.jellysquid.mods.sodium.client.model.quad.properties.ModelQuadOrientation;
 import me.jellysquid.mods.sodium.client.render.chunk.compile.buffers.ChunkModelBuffers;
+import me.jellysquid.mods.sodium.client.render.chunk.data.ChunkRenderData;
 import me.jellysquid.mods.sodium.client.render.chunk.format.ModelVertexSink;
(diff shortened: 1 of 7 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/BlockRenderer.java
```diff
@@ -105,14 +106,14 @@ private void renderQuadList(BlockRenderView world, BlockState state, BlockPos po
...
-            this.renderQuad(world, state, pos, sink, offset, colorizer, quad, light);
+            this.renderQuad(world, state, pos, sink, offset, buffers, colorizer, quad, light);
         }
 
         sink.flush();
     }
 
     private void renderQuad(BlockRenderView world, BlockState state, BlockPos pos, ModelVertexSink sink, Vec3d offset,
-                            BlockColorProvider colorProvider, BakedQuad bakedQuad, QuadLightData light) {
+                            ChunkModelBuffers buffers, BlockColorProvider colorProvider, BakedQuad bakedQuad, QuadLightData light) {
```
jellysquid3: This should simply pass the `ChunkRenderData.Builder` type directly to avoid `getRenderData()` calls below. We shouldn't need to access anything else on `ChunkModelBuffers`, anyways.

jellysquid3:
LGTM, this is basically what we did anyways, just never copied it into the new types during the vertex system changes.

PR #528, opened 2021-02-13
remove redundant chmod +x gradlew

deltanedas:
CI adds the execute bit to `gradlew`.
Screenshot_14:15:19

I simply removed these two lines as it's already set.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -12,8 +12,6 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 14
-    - name: Grant execute permission for gradlew
-      run: chmod +x gradlew
     - name: Build with Gradle
       run: ./gradlew build
     - name: Upload build artifacts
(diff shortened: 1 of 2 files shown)


liach:
Don't think it's done by the ci. You can just do `git update-index --chmod +x gradlew` to change file mode for git

deltanedas:
it's already executable, ci doesn't need to make it so.

LucilleTea:
Does this change actually have any benefit? It seems like you're removing something that's harmless at worst, and a helpful failsafe if it needs to be. I don't see any reason to do that.

magneticflux-:
It's akin to fixing a spelling mistake. It doesn't cause any harm either way.

ImperatorStorm:
Fixing a spelling mistake is correct something wrong.

This, i'd say, is more akin to removing a redundancy.

jellysquid3:
Yes, the bits are already set in Git source tree. There is no reason for the action to be setting the executable bit, and removing the build step to do so is fine. It's just part of the original action template and exists for Windows users without unix-style permissions.

Anyways. Parkinson's law of triviality strikes again. Thank you for your PR -- this is attention to detail, even if it's a small change.

jellysquid3 (approved):

PR #553, opened 2021-03-01
new: Add smooth lighting for paintings

### LINKED ISSUE #90: Smooth lighting for paintings
It's been done for water, so why not these? Super noticeable and annoying for the bigger paintings.

Not sure if it'd be easy to do though since they're entities

FlashyReese:
Painting light is calculated using WorldRenderer#getLightmapCoord unlike traditional Entity Rendering hence it is not covered by the current implementation. This commit calculates the light using Sodium's EntityLighter and redirects it. Fixes #90.

Without

With


muzikbike:
Does this also work for map item frames?

PR #556, opened 2021-03-01
JDK Upgrade to Java 15, compiling for Java 8

Justsnoopy30:
Should fix builds compiled with Java 9+ not running on Java 8. Uses an if condition so it is only activated when compiled with a Java 9 or higher.


Justsnoopy30:
https://bugs.openjdk.java.net/browse/JDK-8206937 is stopping building on Java 9+ due to usage of sun.misc package and its Unsafe API in Sodium. https://bugs.openjdk.java.net/browse/JDK-8214165 may also be useful in details although it's closed.

Justsnoopy30:
Closing, making new PR with new branch.

PR #557, opened 2021-03-01
Change Block method calls to use BlockState method calls

FoundationGames:
Certain methods in `AbstractBlock` are not meant to be called directly, as shown by their being deprecated. They are meant to be called from a `BlockState`. This wasn't respected in Sodium, and it created an issue with Enhanced Block Entities.


FoundationGames:
Resolves FoundationGames/EnhancedBlockEntities#3

jellysquid3 (approved):

PR #559, opened 2021-03-01
new: Fabric API compatability hooks

### LINKED ISSUE #228: Fabric API's ClientChunkEvents aren't invoked
### Expected Behavior
The events in `ClientChunkEvents` should be invoked when appropriate.

### Actual Behavior
Due to how Sodium works, the code that does so is overriden, so they're never invoked.

### Attachments
The events: <https://github.com/FabricMC/fabric/blob/1.16/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/api/client/event/lifecycle/v1/ClientChunkEvents.java>
The mixin responsible for invoking said events: <https://github.com/FabricMC/fabric/blob/1.16/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/client/ClientChunkManagerMixin.java>

Leo40Git:
Adds hooks for the `fabric-rendering-v1` module of Fabric API to invoke events usually invoked by a mixin into `WorldRenderer`.  (these hooks are currently unused, since Sodium doesn't break any of these events yet - future proofing FTW!)


Leo40Git:
Merge conflicts fixed!

DragonEggBedrockBreaking:
> Merge conflicts fixed!

More conflicts now

jellysquid3:
We're no longer using our own chunk manager implementation, so the hooks we were (previously) overwriting no longer need to be added by this PR.

Leo40Git:
Lemme just delete the Lifecycle Events compat stuff real quick-
Should I leave this PR open with just the (currently unused) Rendering hook stuff or just close it?

Leo40Git:
> More conflicts now

Guess what, those are fixed now too :P

src/main/java/me/jellysquid/mods/sodium/client/compat/CompatibilityHooks.java
```diff
@@ -0,0 +1,192 @@
...
+    public static final FabricRenderingHooks FABRIC_RENDERING;
+
+    static {
+        final FabricLoader loader = FabricLoader.getInstance();
+        LOGGER.info("Sodium is now looking for mods to enable compatibility hooks for...");
+        FABRIC_RENDERING = createFabricRenderingHooks(loader);
+        LOGGER.info("Compatibility hooks done! Enjoy your improved performance!");
+    }
+
+    private static FabricRenderingHooks createFabricRenderingHooks(FabricLoader loader) {
```
valoeghese, line 41: Since this is always called on client init, won’t it crash the game if fabric rendering api isn’t installed? Since classloading will search for the fabric api class in the return value, and not find it. Ergo, JVM crashes

valoeghese: Therefore I recommend that anything that directly touches frapi be moved to a new class, which is only classloaded in the case of frapi being installed

valoeghese (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/compat/CompatibilityHooks.java
Leo40Git: The code that actually references FRAPI classes is only called if FRAPI is installed (checked via `loader.isModLoaded("fabric-rendering-api-v1")`. So no, this should not result in a "class not found" error.

Leo40Git:
Closing this for now, will reopen when needed.

PR #561, opened 2021-03-02
Fix sodium crashing on Java 8 when compiled with Java 9+

Justsnoopy30:
Fixes Java 9+ compiled sodium crashing on Java 8, improves build consistency with all builds being compiled with Java 15, which is downloaded if it is not already installed, updates github actions to build with Java 15, and updates gradle wrapper to version 6.8.3.


build.gradle
```diff
@@ -20,6 +17,14 @@ if (build_id != null) {
     version += "+build.${build_id}"
 }
 
+java {
+    toolchain {
+        // Gradle will download the JDK for compiling/running if one isn't found
+        // This should prevent anyone from using an outdated JDK
+        languageVersion.set(JavaLanguageVersion.of(15))
```
ImperatorStorm: Gradle projects should not download JDKs, and I don't think its a good idea to restrict what java version a person wishes to compile with.

ImperatorStorm (changes requested):

build.gradle
Justsnoopy30: Removed, now the user can compile with whatever JDK they want again.

jellysquid3:
In my head, there are really only two solutions to the problem:

- Specify the [`release` argument](https://openjdk.java.net/jeps/247) to the Java compiler using Gradle when compiling with JDK 9+ so an older (see: compatible) classpath is used. This should fix method signature mismatches between versions, and would be the solution I prefer.
- Update the documentation to make it clear that compiling with newer JDKs requires the game's JRE to also be updated, and trust that people messing with the code know how to maintain Java toolchains on their computer... This has merits as well.

The approach this PR is taking does neither, and unfortunately I don't think littering the code with strange casts to fix _one_ signature, in a way that is both messy and impossible to maintain, is a viable solution.

PR #568, opened 2021-03-05
Fixes for a few minor bugs in the Sodium GUI

bytzo:
### Overview

This pull request contains fixes for 6 minor bugs in Sodium's GUI. See each commit and corresponding commit message for more detail, or read on below with some visual examples.

#### Prevent slider thumb from rendering outside of slider area (c8d5645)

The slider thumb in the slider control can currently render outside of the slider area when the value of the slider is outside of its maximum range. While this shouldn't happen during normal use, it can occur if the game options file is manually edited, which is common in speedrunning when the player wants maximum brightness (maximum brightness is 12.0 when the slider goes up to 1.0). Instead of having the slider thumb render outside of the slider area when situations like this occur, we can limit its position to be within the minimum and maximum values of the slider.

#### Remove offset from mouse when setting value of slider control (afb3158)

Currently, when adjusting a value using the slider control, the range of the mouse is shortened by 4 pixels on each side. This means that if the mouse is four pixels inside from the left of the slider, the value is incorrectly set as the minimum. 

An example is shown in the clip below. The top slider shows the current behavior, with the range of the mouse shortened by 4 pixels on each side. The bottom slider shows the proposed fix, which is to remove the 4 pixel offset.

https://user-images.githubusercontent.com/69266322/110059787-c81e1080-7d19-11eb-8056-cb26ddcb888f.mov

Fixing this issue makes the positioning of the slider control and the set value the same, preventing "drift" of the slider thumb when using it.

#### Allow controls to reset themselves when modified options are undone (6bd054f and 4fda9e5)

Currently, control elements keep their fields when modified options are undone. This causes issues when changing a control element and pressing "Undo," because fields such as the slider position of the slider control or the current index of the cycle control are not reset.

An example is shown in the clip below. The clip shows the issue when the control element (in this case the render distance slider) does not reset after it is modified and the "Undo" button is pressed. In the clip, the render distance is modified from 12 to 24 before resetting, but the slider still displays the value of 24 instead of 12.

https://user-images.githubusercontent.com/69266322/110069147-10dec500-7d2c-11eb-999e-ece12098846d.mov

To solve these issues, the control elements will now fetch the updated values when necessary, either on hover (in the case of the slider) or on click (in the case of the cycle).

#### Compare current and modified option values using `equals()` (1970cc2)

The current and modified option values in the Sodium options need to be compared to see if the value is changed or not. Previously, this was done using the `==` operator. However, since the current and modified options are wrapper types, it will compare them as references, making them unequal when their values are the same. This issue does not happen in a majority of the Sodium options, since Java may cache the objects as values. To solve this issue, the current and modified options can be compared using `equals()`, which will compare by value instead. See [this article](https://programming.guide/java/boxed-values-equality.html) about this.

#### Prevent hidden button widgets from being pressed (24158b1)

Currently, button widgets that are not visible can be clicked on, since only the enabled state of the button is considered if an action should be taken or not. To fix this issue, button widgets should also consider their own visibility state when taking action after a click.

#### Allow for more precise widget rendering (69e031b)

Currently, the rendering of rectangles and quads in the Sodium GUI is defined using integers. This restricts the position of widgets to each in-game pixel, meaning using a scaled GUI scale causes the rendered position to be incorrect in some scenarios. For example, a user on a HiDPI display using a GUI scale of 8 would not see the brightness slider thumb move when setting it from 0 to 1, because it must be aligned to every in-game pixel (in this scenario, each in-game pixel is 8 real pixels on the display, because of a GUI scale of 8). To solve this, the rendering of rectangles and quads should use doubles to define position, allowing for greater positioning precision when needed.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/SliderControl.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/SliderControl.java
@@ -2,6 +2,7 @@ package me.jellysquid.mods.sodium.client.gui.options.control;
 
 import me.jellysquid.mods.sodium.client.gui.options.Option;
 import me.jellysquid.mods.sodium.client.util.Dim2i;
+import me.jellysquid.mods.sodium.common.util.Resettable;
 import net.minecraft.client.util.Rect2i;
 import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.util.math.MathHelper;
@@ -42,7 +43,7 @@ public class SliderControl implements Control<Integer> {
         return 130;
     }
 
-    private static class Button extends ControlElement<Integer> {
+    private static class Button extends ControlElement<Integer> implements Resettable {
         private static final int THUMB_WIDTH = 2, TRACK_HEIGHT = 1;
 
         private final Rect2i sliderBounds;
@@ -156,6 +157,11 @@ public class SliderControl implements Control<Integer> {
             }
 
             return false;
+        }
+
+        @Override
+        public void reset() {
+            this.thumbPosition = this.getThumbPositionForValue(option.getValue());
         }
     }
 
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/SliderControl.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/SliderControl.java
@@ -2,7 +2,6 @@ package me.jellysquid.mods.sodium.client.gui.options.control;
 
 import me.jellysquid.mods.sodium.client.gui.options.Option;
 import me.jellysquid.mods.sodium.client.util.Dim2i;
-import me.jellysquid.mods.sodium.common.util.Resettable;
 import net.minecraft.client.util.Rect2i;
 import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.util.math.MathHelper;
@@ -43,7 +42,7 @@ public class SliderControl implements Control<Integer> {
         return 130;
     }
 
-    private static class Button extends ControlElement<Integer> implements Resettable {
+    private static class Button extends ControlElement<Integer> {
         private static final int THUMB_WIDTH = 2, TRACK_HEIGHT = 1;
 
         private final Rect2i sliderBounds;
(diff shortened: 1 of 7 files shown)


FlashyReese (approved):

src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/SliderControl.java
```diff
@@ -157,6 +158,11 @@ public boolean mouseDragged(double mouseX, double mouseY, int button, double del
 
             return false;
         }
+
+        @Override
+        public void reset() {
+            this.thumbPosition = this.getThumbPositionForValue(option.getValue());
```
FlashyReese: I have just come to realize how needless that interface was, you can simply call whatever you need to reset on the render method. In `SliderControl`'s case I would place it in `renderSlider` method. Same thing applies for `CyclingControl`.

bytzo: I have removed the interface in 4fda9e5ac2e541358b45c48416e3c3f273e530ae.

CalXee:
Entity distance, if you set it to anything that's not 100% or 125% causes some issues. Namely, if you for example set it to 500%, then slide it to another value then back to 500%, it will still prompt you to apply the entity distance changes despite you didn't actually change anything. I hope I explained that well. This isn't an issue specific to this commit, I'm just saying this in case you guys are unaware since I feel it is best to address that issue with this PR.

bytzo:
> Entity distance, if you set it to anything that's not 100% or 125% causes some issues. Namely, if you for example set it to 500%, then slide it to another value then back to 500%, it will still prompt you to apply the entity distance changes despite you didn't actually change anything. I hope I explained that well. This isn't an issue specific to this commit, I'm just saying this in case you guys are unaware.

I added commit 1970cc2074f12c1236d4adab8172ded78279eb34 to fix this. It turns out to be an issue where Sodium compares two wrapper types as references, not values. And since these can be cached as values, it is not seen on sliders that do not exceed 127. I found [this article](https://programming.guide/java/boxed-values-equality.html) when trying to understand the issue, it explains what is happening well.

jellysquid3 (approved):

jellysquid3:
Everything here looks good to me. Apologies for the delay on merging this, and thanks for your extensive documentation on the issues + your solutions.

PR #570, opened 2021-03-07
fix java 8 bytebuffer error

imreallybadatnames:
fix java.lang.NoSuchMethodError: java.nio.ByteBuffer.clear()Ljava/nio/ByteBuffer; on java 8


imreallybadatnames:
i gotta leave this pull request behind for some reasons

PR #571, opened 2021-03-07
change: Better configuration handling

Leo40Git:
- Corrupt config files will be replaced with a copy of the default values.
  - The corrupt file will be backed up for inspection.
  - The game will still crash, just to give people a heads-up that their config's hosed and has been replaced.
- Changed the way in-game options are saved to the file. This *should* fix the "corrupt config" bug.
  - More specifically, the new file is written to a temporary file, and then that temporary file is moved into the original file's place.
- The `core` mixin package can no longer be disabled.
- I/O now uses NIO `Path`s, because they're cool

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
@@ -188,7 +188,7 @@ public class SodiumGameOptions {
     }
 
     public void writeChanges() {
-        Path dir = path.getParent();
+        Path dir = this.path.getParent();
 
         if (!Files.exists(dir)) {
             try {
@@ -201,12 +201,12 @@ public class SodiumGameOptions {
             LOGGER.error("Parent directory \"{}\" is, in fact, not a directory!", dir);
         }
 
-        try (OutputStream os = Files.newOutputStream(path);
+        try (OutputStream os = Files.newOutputStream(this.path);
              OutputStreamWriter osw = new OutputStreamWriter(os);
              BufferedWriter writer = new BufferedWriter(osw)) {
             gson.toJson(this, writer);
         } catch (IOException e) {
-            LOGGER.error("Could not save config file to \"" + path + "\"!", e);
+            LOGGER.error("Could not save config file to \"" + this.path + "\"!", e);
         }
     }
 }
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
@@ -221,7 +221,7 @@ public class SodiumGameOptions {
         try {
             Files.move(tempPath, this.path, StandardCopyOption.REPLACE_EXISTING);
         } catch (IOException e) {
-            LOGGER.error("Failed to move config file \"" + tempPath + "\" into place at \"" + this.path + "!", e);
+            LOGGER.error("Failed to move config file \"" + tempPath + "\" into place at \"" + this.path + "\"!", e);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
```diff
@@ -139,22 +144,38 @@ public String getLocalizedName() {
...
 
             config.sanitize();
         } else {
+            LOGGER.info("Could not find options file, loading default values");
             config = new SodiumGameOptions();
         }
 
-        config.file = file;
+        config.path = path;
         config.writeChanges();
```
FlashyReese: I have mentioned this in `#sodium` before. Perhaps we should only write changes once when creating the config file for this first time.
```suggestion
        if (!File.exists(path)) {
                config.writeChanges();
        }
```

src/main/java/me/jellysquid/mods/sodium/common/util/PathUtil.java
```diff
@@ -0,0 +1,21 @@
...
+     * Resolves a timestamped sibling of the base path - {@code "{base_dir}/{base_name}_{tag}_{timestamp}.{base_ext}"}.
+     * @param base Base path
+     * @param tag Tag to add to sibling's name
+     * @return Sibling path
+     */
+    public static Path resolveTimestampedSibling(Path base, String tag) {
+        LocalDateTime now = LocalDateTime.now();
+        String backupName = base.getFileName().toString();
+        backupName = backupName.substring(0, backupName.lastIndexOf('.') - 1);
+        backupName += "_" + tag + "_" + now.format(DateTimeFormatter.ISO_LOCAL_DATE_TIME) + ".json";
```
magneticflux-: This causes issues with path resolution because `:` is not a valid file name. Reported on the Discord here: https://discord.com/channels/602796788608401408/778685768041627719/819478369418215424

src/main/java/me/jellysquid/mods/sodium/common/util/PathUtil.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Resolves a timestamped sibling of the base path - {@code "{base_dir}/{base_name}_{tag}_{timestamp}.{base_ext}"}.
+     * @param base Base path
+     * @param tag Tag to add to sibling's name
+     * @return Sibling path
+     */
+    public static Path resolveTimestampedSibling(Path base, String tag) {
+        LocalDateTime now = LocalDateTime.now();
+        String backupName = base.getFileName().toString();
+        backupName = backupName.substring(0, backupName.lastIndexOf('.') - 1);
```
magneticflux-: ~This is not a reliable way to determine the file name without extension. In fact, this website's first example of how not to ask a technical question uses file extension handling to convey the point: https://xyproblem.info/~

Nevermind, it appears the Microsoft .NET standard library uses the "last `.` truncation" technique: [https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilenamewithoutextension?redirectedfrom=MSDN&view=net-5.0#System_IO_Path_GetFileNameWithoutExtension_System_String_](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilenamewithoutextension?redirectedfrom=MSDN&view=net-5.0#System_IO_Path_GetFileNameWithoutExtension_System_String_)

magneticflux- (changes requested):

src/main/java/me/jellysquid/mods/sodium/common/util/PathUtil.java
Leo40Git: great job, past me. what do I even pay you for

Leo40Git: Should be fixed in https://github.com/CaffeineMC/sodium-fabric/pull/571/commits/220c417fca9ae23a142661874879a31ee300cab5.

src/main/java/me/jellysquid/mods/sodium/common/util/PathUtil.java
magneticflux-: @Leo40Git Hey, turns out a relevant PR in OpenJDK was opened recently! https://github.com/openjdk/jdk/pull/2319

jellysquid3:
This PR is overly complex in my opinion and doesn't solve the actual issue resulting in corruption. It's not normal for applications to see such rampant issues with file corruption (we are the only mod that seems to run into this), so I'm convinced the issue has something to do with our usage of FileWriter or some other I/O construct.

I've made changes with https://github.com/CaffeineMC/sodium-fabric/commit/6f1762fd8844cef131056b9f24927408dd369959 so that our config system uses Java's NIO library, which has a better API and much more clearly defined behavior.

I'm hopeful it will resolve these issues we're seeing with config file corruption... if not, we can come back to this PR and/or other workarounds.

Issue #445, opened 2020-12-26
Another mod applying an AccessWidener to a Sodium @Overwrite'd method causes startup crash


exaskye:
### Reproduction Steps

1. Install any mod that uses an AccessWidener on a method that Sodium `@Overwrite`s. (Such as my own [Fabrication](https://www.curseforge.com/minecraft/mc-mods/fabrication) prior to v1.2.6)
2. Try to launch the game.
3. Fail.

### Crash Report File
https://gist.github.com/unascribed/6f2274e193de6675fe42d4909ec61326

### Additional Information
This can be fixed by telling Mixin to be less strict about access in `@Overwrite`s by adding the following to `sodium.mixins.json` (This is how I currently fix it and is how I've recommended others fix it):
```json
"overwrites": {
  "conformVisibility": true
}
```

This however will still log a warning, and I'm unsure if I'd consider this a "true" fix. I also am unsure if I'd consider this Sodium's "fault"; the alternative on my end is to use MethodHandles. Even if that is the "true" fix, I figured I should bring it up here ~~because I don't believe this will be an isolated incident~~. (Discovered #385, this is *not* an isolated incident.)

Related: unascribed/Fabrication#30 (though I've known about this for a while longer than this issue's reporting)

----

I see #385 exists, but there is no mention of the conformVisibility fix/workaround, and the response suggests it is thought there is *nothing* that can be done about it on Sodium's side. I discovered it via digging in Mixin; it's not documented as far as I know. (Though, about 60% of Mixin isn't documented, so who's surprised.)

I will accept a closed issue and punt back to my side and use MethodHandles. Wanted to send over what I knew.

jellysquid3:
I did not know this workaround existed, thanks for bringing it up. As you mention, this won't really _fix_ the problem, but continuing on with a loud warning should be fine. These method visibility conflicts are not very frequent, and I don't really expect us to encounter runtime errors after enabling this for those few cases. It should be fine to enable on Sodium's side.

Issue #385, opened 2020-11-08
Game Crashing with Sodium


ghost:
### Reproduction Steps
1. Sodium is placed into mod folder.
2. The game crashes if Sodium is present.

### Crash Report File
[GitHub Gist](https://gist.github.com/windows-virus/873998a661e3144da6754d5115a955ba) 

### Additional Information
The game does not crash if Sodium is not present so most likely a mod/s is conflicting is with Sodium (I have like 150+ mods, so finding the culprit mod/s would not be an easy task). Not Enough Crashes show that Sodium is conflicting with the Fabric Mod Loader.

LambdAurora:
This crash is caused by the library mod Geckolib which is present in the crash report you sent, this is a known issue which has been already reported on geckolib bernie-g/geckolib#26.

This issue is caused by something used in Geckolib (access widener) which badly conflicts with Sodium as it changes something in the code which Sodium doesn't expect to. This can't really be fixed by Sodium and has to be fixed by Geckolib, in the linked issue the author said the compatibility problem will be fixed in a newer version of Geckolib which is not yet released.

ghost:
Thank you for the quick response. Crashing has stopped.

PR #575, opened 2021-03-10
change: Allow overwrites to apply to access widened methods

LucilleTea:
Applies the accepted proposal in #445, which should occasionally help mod compatibility.
I've confirmed it fixed the crash with Fabrication 1.2.5, before the author worked around the issue. This type of crash has also been reported in #385 (with geckolib, which seems to have also worked around the issue by now).

I'm not aware of any mod currently affected by this, but it still seems sensible to apply the change.

### DIFF
--- a/src/main/resources/sodium.mixins.json
+++ b/src/main/resources/sodium.mixins.json
@@ -6,6 +6,9 @@
   "injectors": {
     "defaultRequire": 1
   },
+  "overwrites": {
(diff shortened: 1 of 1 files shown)


comp500:
This was part of my fork for a while (to fix the GeckoLib 2.x compat issue when it was in AOF3), it seems to work quite well.

jellysquid3 (approved):

Issue #189, opened 2020-07-26
Crash on Joining a Minigame 


frqnny:
### Reproduction Steps
Provide information on how to reproduce this game crash. This step is critical for ensuring that your crash can be debugged by other developers. You can either fill in the example below or do something else -- just make sure your instructions are minimal and clear. If applicable, provide a Minecraft save containing a minimal example on how to reproduce this crash as a compressed ZIP file.

I went into a world and for some reason it crashed, IDK why to have enough reasons. 

### Crash Report File
Please upload your crash report as a file to [GitHub Gist](https://gist.github.com/) and replace this text with a link to the uploaded report.
https://pastebin.com/JuMYXa2i

the bottom crash is the one that killed my game and froze it

### Additional Information
Provide any additional information or context which may be relevant to the issue. If you have none to add, you can remove this section.

embeddedt:
I've run into this issue as well. I'm using HolographicDisplays 2.4.6 on a Paper 1.16.5 server proxied through the latest version of Waterfall.

The issue can be worked around by disabling HolographicDisplays or removing Sodium (neither of which I want to do :wink:).

I should note that the issue doesn't occur if you create the hologram (armor stand) after the world has been loaded. It only happens if the armor stand is pre-existing when you join the world.

Crash log: https://gist.github.com/embeddedt/81cd9b52fff933fb399827d339caf5e1

comp500:
#581 should fix this crash, though it's surprising that it occurs so rarely. I've managed to replicate the crash by setting the water BlockState's shape cache to null in a debugger - it doesn't seem to happen when the shape cache is set as it usually would be. Not sure exactly why the shape cache would be null, as it's set during tag loading... it could be a race condition when syncing that data with the server?

Edit: Could be related to the use of protocol translation mods/plugins - see https://github.com/CaffeineMC/sodium-fabric/issues/208#issuecomment-665831827

comp500:
@embeddedt I can't replicate this behaviour (and I think it's actually https://github.com/CaffeineMC/sodium-fabric/issues/218 which I don't think is a duplicate of this issue) - does this happen with a brand new world? Is there specific text which makes it occur, or does it happen with any holographic display?

embeddedt:
[...]

PR #581, opened 2021-03-14
fix: Increase WorldSlice neighbor radius to match vanilla

comp500:
Several mods (primarily LambdaBetterGrass) require extra context in chunk rendering, that isn't provided by the 18x18x18 (5832 blocks) chunk section copy that Sodium makes. This change increases it to 20x20x20 (8000 blocks), which matches vanilla - so all mods should expect to have this radius of context for rendering. Unfortunately, as more blocks are being copied, this will impact chunk building performance very slightly, though much better than my original 48x48x48 change.

Other methods for increasing this radius without affecting performance (as this is very hot code) may be worth exploring, though it may not be worth the additional complexity.

The change has been tested in my Sodium fork, though the performance implications have not been directly measured.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
@@ -48,7 +48,7 @@ public class WorldSlice extends ReusableObject implements BlockRenderView, Biome
     private static final int SECTION_BLOCK_COUNT = SECTION_BLOCK_LENGTH * SECTION_BLOCK_LENGTH * SECTION_BLOCK_LENGTH;
 
     // The radius of blocks around the origin chunk that should be copied.
-    private static final int NEIGHBOR_BLOCK_RADIUS = 1;
+    private static final int NEIGHBOR_BLOCK_RADIUS = 2;
 
     // The radius of chunks around the origin chunk that should be copied.
     private static final int NEIGHBOR_CHUNK_RADIUS = MathHelper.roundUpToMultiple(NEIGHBOR_BLOCK_RADIUS, 16) >> 4;


notsatvrn:
Been testing this fork with Indium, and I don't notice any performance decrease, on Linux atleast.

coderbot16:
It is possible that this could fix #189, since that crash is caused by an out-of-bounds block access in WorldSlice, and this change makes it so that WorldSlice copies the correct amount of blocks needed to match what vanilla does.

coderbot16 (approved):

comp500:
I think you're right - that crash looks like it is caused when a block has a fluid block directly above it in the ChunkSection above it, and Sodium tries to compute the lighting data for it. The fluid block tries to look for a block above when calculating the collision shape, which is outside the WorldSlice bounds.

~~I can't seem to replicate the crash though, presumably as it requires some caches to be empty (if the ChunkSection containing the fluid block was rendered first, it might cache the correct light data?).~~ See my comment on the issue, I have managed to replicate it.

MeeniMc:
Some data to feed the performance discussion: comparing build artifacts 324 (next) and 325 (this PR) (spec of system are in the screenshots). There is no visible performance difference with the default texture pack (16px). 

## Stationary testing, Render Distance 32

From top to bottom: build.324 (next), build.325 (this pr), build.325+Indium+Lambd(grass+light)





## 256 px textures

I did see a small performance difference with 256px (110fps down to 107fps) that appears to be more  than just random deviation. Not crashing is worth such an insignificant loss in any case. 

## In-motion testing, backpedaling

This is harder to do a fair and square comparison here, as I am doing the 'motion' by hand. To try to discriminate between chunk-loading lag, and render lag, I am moving 'backward', which loads less chunks at any given time. 


<details><summary>OBSOLETE RESULTS</summary>
<p>
Again, order is 324, 325, 325+Indium





The screens are not super conclusive, the pr 324 and pr 325 felt the same (there is some spiking/lag when traveling, especially when crossing chunk boundaries, could still be chunk-loading, despite using 'backpedaling'). It does not feel smooth. Prior Sodium versions felt smoother, but that's not related to this PR. 

Indium+lambd adds a significant amount of spiking/lagging, to the point of being honestly annoying when flying forward. 
</p>
</details>

I redid the experiments after rebasing over 6caf15f and all performance anomalies are resolved. 

## Player stationary, rotating motion

Player stationary, and turning around at high angular speed (doing a 360 panorama), there is absolutely no lag in any of the versions.

DragonEggBedrockBreaking:
> Some data to feed the performance discussion: comparing build artifacts 324 (next) and 325 (this PR) (spec of system are in the screenshots). There is no visible performance difference with the default texture pack (16px).

You have been measuring FPS difference, but I believe this pull request should not directly affect that, and should only affect chunk rebuilding times. Also, your 256x screenshots had many lagspikes, and it seems you measured the last one during a lagspike, which may've altered the results.

MeeniMc:
Maybe someone should try  the same @1080p

> > Some data to feed the performance discussion: comparing build artifacts 324 (next) and 325 (this PR) (spec of system are in the screenshots). There is no visible performance difference with the default texture pack (16px).
> 
> You have been measuring FPS difference, but I believe this pull request should not directly affect that, and should only affect chunk rebuilding times. Also, your 256x screenshots had many lagspikes, and it seems you measured the last one during a lagspike, which may've altered the results.

I think you are correct, the PR would affect more chunk loading than rendering speed. 

Indeed, the lagspikes are in the 'in-motion' tests. They are not random noise, they clearly corelate with the player crossing chunk boundaries/loading new chunks for rendering. (I did not present the screens for 256px, there was nothing to say beside the fps being marginally lower).

In more detail, the 'in-motion' test is done with the following method:
1. Load the game, wait for the FPS to stabilize at max-fps.
2. fly around the area to pregenerate (or load from disk) all the chunks in the neighboring regions. As you can see I have plenty of memory, and half of it is still un-used, so presumably all disk-loaded chunks for the regions are now resident in memory. 
3. Return to starting point, wait for FPS to stabilize
4. use the 's' key to fly backward with speed of 80 m/s; this will bring new chunks into view, but only some at a time (the closest ones), this will cause massive lag spikes, as you note on the screen. The feel is not smooth, but 'ok'. This is not noise that mask the signal, this is the signal. 
5. you can do the same test using the 'w' key to fly forward, with render-distance 32, this is lagging pretty bad and is feeling sluggish and annoying. 

In-motion smoothness is much worse than prior Sodium versions (and with Indium active, another level of worse), probably due to chunk-loading/rebuilding being slower. The thing is that **whatever performance defect is not introduced by this PR, as it is also present in 324.** Maybe this discussion should go in its own issue. 


I may try to run with Sparkle, but I have never used it before, so no promises of getting anything useful from that.

MeeniMc:
I just rebased this PR over   6caf15fcb84dcaa26175fe88d29cd9cd0e3ab6f3  and all performance issues are gone. It's all smooth and can't tell the difference, even when flying around (after preloading  the area), or hitting f3-a.

hxr404 (approved):
In my personal tests (on Windows), I didn't experience any (notable) performance loss or any other issues.
I think this could be merged

valoeghese (approved):

papuSpartan:
fixed a recurring crash for me ( was already on the latest release of the next branch when I was getting the crash)

jellysquid3:
Important changes have happened here with https://github.com/CaffeineMC/sodium-fabric/commit/f4d08208a91da6166f82a6deb40dc4af1007b2b9, but for the better. It makes slicing the world significantly faster and moves the cost associated with this PR off-thread.

I will investigate increasing the radius later this weekend.

PR #598, opened 2021-03-21
new: Show mod version on F3 screen

Leo40Git:
Shows "(!!) Dev Environment" in red if we're in the mod's development environment (IE if the version is the ${version} placeholder).
**Note:** This preview is outdated! Scroll down for an up-to-date preview.
Preview

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
@@ -3,18 +3,34 @@ package me.jellysquid.mods.sodium.client;
 import me.jellysquid.mods.sodium.client.gui.SodiumGameOptions;
 import me.jellysquid.mods.sodium.client.util.UnsafeUtil;
 import net.fabricmc.api.ClientModInitializer;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import net.minecraft.util.Formatting;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.File;
 
 public class SodiumClientMod implements ClientModInitializer {
+    private static String VERSION;
     private static SodiumGameOptions CONFIG;
     private static Logger LOGGER;
 
     @Override
     public void onInitializeClient() {
 
+    }
+
+    public static String version() {
+        if (VERSION == null) {
+            ModContainer ourContainer = FabricLoader.getInstance().getModContainer("sodium")
+                    .orElseThrow(() -> new IllegalStateException("Mod with ID \"sodium\" couldn't be found... but that's us!"));
+            VERSION = ourContainer.getMetadata().getVersion().getFriendlyString();
+            if ("${version}".equals(VERSION))
+                VERSION = Formatting.RED + "(!!) Devel";
+        }
+
+        return VERSION;
     }
 
     public static SodiumGameOptions options() {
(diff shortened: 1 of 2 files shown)


ImperatorStorm (approved):

CalXee:
May I ask what "!! Devel" means?

ImperatorStorm:
> May I ask what "!! Devel" means?

commit [f55422b](https://github.com/CaffeineMC/sodium-fabric/pull/598/commits/f55422bcef6520674889f3af424cd36436e51d55)
> Show "(!!) Devel" in red if we're in the mod's development environment (IE if the version is the ${version} placeholder)

DragonEggBedrockBreaking:
> May I ask what "!! Devel" means?

"(!!)" is likely an exclamation/a small warning, and "Devel" is short for development environment

machinesmith42:
would deven be right then?

src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
```diff
@@ -17,6 +21,18 @@ public void onInitializeClient() {
 
     }
 
+    public static String version() {
+        if (VERSION == null) {
+            ModContainer ourContainer = FabricLoader.getInstance().getModContainer("sodium")
+                    .orElseThrow(() -> new IllegalStateException("Mod with ID \"sodium\" couldn't be found... but that's us!"));
+            VERSION = ourContainer.getMetadata().getVersion().getFriendlyString();
+            if ("${version}".equals(VERSION))
+                VERSION = Formatting.RED + "(!!) Devel";
```
silversquirl: For clarity, and in order to match the style elsewhere in Sodium, this should be either on the same line as the if:

```java
            if ("${version}".equals(VERSION)) VERSION = Formatting.RED + "(!!) Devel";
```

Or wrapped in braces:

```java
            if ("${version}".equals(VERSION)) {
                VERSION = Formatting.RED + "(!!) Devel";
            }
```

I would recommend the latter.

src/main/java/me/jellysquid/mods/sodium/mixin/features/debug/MixinDebugHud.java
```diff
@@ -35,6 +36,9 @@ private void appendRightText(CallbackInfoReturnable<List<String>> cir) {
             }
         }
 
+        strings.add("");
+        strings.add("Sodium Version: " + SodiumClientMod.version());
+
         strings.add("");
```
silversquirl: I think it would make sense to remove the spacing here, so the two things added by Sodium are grouped

silversquirl (changes requested):

Leo40Git:
Updated preview:
image

silversquirl (approved):

Boobies:
Is this really necessary? If every mod did this, the debug screen would be a disaster. Presumably, the user who installed the mod, knows what version they are running and, if not, they can easily check Mod Menu or their mods directory. I see this PR as mildly detrimental at best.

silversquirl:
Knowing the version is incredibly helpful for dealing with support queries when many people are running builds of the next branch or forks of the project.

DragonEggBedrockBreaking:
> Knowing the version is incredibly helpful for dealing with support queries when many people are running builds of the next branch or forks of the project.

Just to add on to what vktec said, people do not usually run development builds of most other mods (in my experience), so there is very little reason for "every mod" to do this.

MeeniMc:
Maybe it would make sense to have the git commit hash rather than a generic 'Dev'.

Also having more frequent releases (publish anything that is an improvement in A.B, do not hold until feature X is finally implemented, I't ll go in A.B+1 just fine) may help with people using Dev versions left and right.

Leo40Git:
> Maybe it would make sense to have the git commit hash rather than a generic 'Dev'.

The generic "Dev" is only used if someone's playing around in Sodium's development environment (_not_ in any development environment).

Boobies:
> Knowing the version is incredibly helpful for dealing with support queries when many people are running builds of the next branch or forks of the project.

Knowing that stuff is useful but I don't see what that has to do with putting it on the debug screen. Could also print it on the hotbar and under each button in the main menu since knowing the version is useful. Or, when people are asking for support, they could say what they've installed instead of posting F3 screenshots.

silversquirl:
We ask for Alt+F3 screenshots anyway since they provide other useful information. It's called the debug screen for a reason, after all :)

PR #601, opened 2021-03-22
Remove `vendor` property from build.gradle, allowing sodium to be bui…

ImperatorStorm:
…lt with jdk's other than those from Adoptopenjdk

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,6 @@ java {
 
     toolchain {
         languageVersion = JavaLanguageVersion.of(8)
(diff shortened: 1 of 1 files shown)


DragonEggBedrockBreaking:
> …lt with jdk's other than those from Adoptopenjdk

But then which JDK does it choose? I don't want to download Oracle Java.

jellysquid3:
@DragonEggBedrockBreaking https://docs.gradle.org/current/userguide/toolchains.html#sec:consuming

PR #608, opened 2021-03-26
fix: Use default value when block state palette entry is missing

### LINKED ISSUE #565: Freezing when in certain chunks
### Expected Behavior
I can play.

### Actual Behavior
The game freezes.

### Reproduction Steps
Go to a certain chunk.

### Attachments
https://pastebin.com/k7qw6wfU

### System Information
- Java Version: Oracle JDK 12
- CPU: i5 6400K
- GPU: GTX 1050

Johni0702:
This mirrors vanilla behavior (see `PalettedContainer.get(index)`).

Fixes #565

Not entirely sure why the `paletteId` was invalid in the first place (in my case, playing on wynncraft.com, probably an artifact of multi-version support or other server-side shenanigans) but Vanilla (and Sodium 0.1.0) can deal with those just fine.
The freezing mentioned in the issue is because the main thread gets stuck waiting for important chunk updates after all chunk worker threads have died from NPEs in random places.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
@@ -261,6 +261,9 @@ public class WorldSlice extends ReusableObject implements BlockRenderView, Biome
                         state = prevPaletteState;
                     } else {
                         state = palette.getByIndex(paletteId);
+                        if (state == null) {
+                            state = container.defaultValue;
+                        }
 
                         prevPaletteState = state;
                         prevPaletteId = paletteId;
--- a/src/main/resources/sodium.accesswidener
+++ b/src/main/resources/sodium.accesswidener
@@ -9,3 +9,4 @@ accessible method net/minecraft/client/render/Frustum isVisible (DDDDDD)Z
 
 accessible field net/minecraft/world/chunk/PalettedContainer data Lnet/minecraft/util/collection/PackedIntegerArray;
 accessible field net/minecraft/world/chunk/PalettedContainer palette Lnet/minecraft/world/chunk/Palette;
+accessible field net/minecraft/world/chunk/PalettedContainer defaultValue Ljava/lang/Object;


comp500:
I've investigated a similar bug with this before (see [this Caffeine Discord conversation](https://discord.com/channels/602796788608401408/651120262129123330/804142782989336617)) - this can happen when an unregistered block is generated into the world. As it is unregistered, it has a BlockState that isn't in the registry, so when the server tries to serialise it to send to the client, it gets a raw ID of `-1`. This `-1` is sent across the network, and when deserialised becomes a `null` BlockState in the palette, which then causes the NPE you've observed.

I'm not sure if the use of unregistered blocks is related to the server issue causing this, but it's certainly a good idea to improve Sodium's handling of invalid data from the server - it may be better to patch this in deserialisation code instead (though such a fix would not be in the scope of Sodium).

Johni0702:
> it's certainly a good idea to improve Sodium's handling of invalid data from the server - it may be better to patch this in deserialisation code instead

This is not, technically, invalid data. Vanilla already handles it properly in `PalettedContainer.get()` which is normally the only way to access the block states in the container.
Sodium however since $someCommit (it was fine in 0.1.0) directly accesses the private fields of the `PalettedContainer`, presumably because snapshotting the entire container is a lot faster this way, but does not properly account for the missing entry special case (whereas Vanilla does).

jellysquid3:
> it's certainly a good idea to improve Sodium's handling of invalid data from the server - it may be better to patch this in deserialisation code instead

I don't believe there is anything to patch in de-serialization here... We are the ones hacking into PalettedContainer's internal structures here, and these are simply unmapped values in the palette that need to be dealt with. The overhead of checking for this case is very minimal and should just compile down to a single CMOV, so I'm really not concerned anyways.

> presumably because snapshotting the entire container is a lot faster this way

Yes, this is correct, but I'm not 100% happy with the approach. I believe it might be a better idea to just copy the PalettedContainer's storage and palette using memory copies than go through the relatively much more complex logic here. If that means slightly slower block reads for significantly faster initialization of render tasks, it will still probably end up being a net win since this code is in our critical path for chunk updates.

jellysquid3 (approved):

PR #610, opened 2021-03-27
fix: `capacity` not updated after `ChunkDrawParamsVector.growBuffer`

Johni0702:
In `pushChunkDrawParams` the need to grow the buffer is determined by comparing the write offset to the `capacity`
field. But after the buffer had been grown, the `capacity` field had been left unchanged, thereby triggering another
grow on each call to `pushChunkDrawParams`, quickly running out of memory.

Looks like this was just forgotten about cause `onBufferChanged` already exists and `growBuffer` is already overwritten in the Unsafe version (which I'd guess most people are using).

I ran across this when disabling `Use Memory Intrinsics` (so it'll use NIO rather than Unsafe) at high render distance (so the initial buffer is too small), while trying to track down another issue.


jellysquid3 (approved):

jellysquid3:
Good catch, this code path didn't see a lot of testing.

PR #615, opened 2021-03-29
fix: Propagate exceptions from chunk build tasks to the main thread

comp500:
Chunk build tasks in Sodium can throw due to issues with mods or bugs in chunk building, so exceptions from these cause the chunk task threads to exit. This causes the main thread to hang in `ChunkRenderManager.updateChunks` waiting for completed build tasks, without creating a crash report for the user to submit.

This change adds a catch-all exception handler around chunk build task execution, which propagates the exception so that it is re-thrown in `FutureDequeDrain.next()` and generates a crash report on the main thread.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuilder.java
@@ -343,20 +343,18 @@ public class ChunkBuilder {
                 }
 
                 ChunkBuildResult result;
+
                 try {
                     // Perform the build task with this worker's local resources and obtain the result
                     result = job.task.performBuild(this.pipeline, this.bufferCache, job);
                 } catch (Exception e) {
-                    // Release resources attached to the failed task
-                    job.task.releaseResources();
-
                     // Propagate any exception from chunk building
                     job.future.completeExceptionally(e);
                     continue;
+                } finally {
+                    // After the task has executed, it's safe to release any resources attached to the task
+                    job.task.releaseResources();
                 }
-
-                // After the result has been obtained, it's safe to release any resources attached to the task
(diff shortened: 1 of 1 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuilder.java
```diff
@@ -342,8 +342,18 @@ public void run() {
...
                 }
 
-                // Perform the build task with this worker's local resources and obtain the result
-                ChunkBuildResult result = job.task.performBuild(this.pipeline, this.bufferCache, job);
+                ChunkBuildResult result;
+                try {
+                    // Perform the build task with this worker's local resources and obtain the result
+                    result = job.task.performBuild(this.pipeline, this.bufferCache, job);
+                } catch (Exception e) {
+                    // Release resources attached to the failed task
```
coderbot16: Perhaps it would be good to move `releaseResources` to a `finally` block, in order to reduce code duplication?

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/ChunkBuilder.java
```diff
@@ -342,8 +342,18 @@ public void run() {
                     continue;
                 }
 
-                // Perform the build task with this worker's local resources and obtain the result
-                ChunkBuildResult result = job.task.performBuild(this.pipeline, this.bufferCache, job);
+                ChunkBuildResult result;
+                try {
```
coderbot16, line 347: Style: One blank line between the start of a new block and a previous non-comment line

```suggestion

                try {
```

coderbot16 (approved):
Looks good to me!

jellysquid3 (approved):

PR #620, opened 2021-04-05
fix: Use auto GUI scale as GUI scale limit (#259)

### LINKED ISSUE #259: GUI size limited to 4

MeeniMc:
Instead of using a fixed value of 4 for the maximum GUI scale in the Sodium options page, the auto GUI scale (largest calculated size) will be used. 

Allowing the GUI scale to extend above 4 is helpful on high-dpi displays, where significantly larger GUI scales are necessary to match the physical size of equivalent low-dpi displays. 

This matches vanilla behavior and fixes #259.

Original work by @Bytzo (closed pr #390), updated for Sodium-next by @MeeniMc

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -68,7 +68,9 @@ public class SodiumGameOptionPages {
                         .setName("GUI Scale")
                         .setTooltip("Sets the maximum scale factor to be used for the user interface. If 'auto' is used, then the largest scale factor " +
                                 "will always be used.")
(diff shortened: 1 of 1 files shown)


DragonEggBedrockBreaking:
Bytzo has now explained why #390 was closed and why he didn’t go to 1.16.x/next, so should this remain open?

MeeniMc:
In my opinion this PR is still an improvement across the board.

1. The steps to make it break are not common-use case: one needs to open the Sodium config, change the window size to a point where the slide would need to be updated
2. When it breaks, it simply reverts to prior behavior, it doesn't break catastrophically. That is, if you resize the window down, the slider will have a maximum value of 9 instead of 4 (a harmless inconsistency), if you resize the window up, it will still show a maximum of 4 (the prior behavior).

For these reasons I think this PR should still be merged.

MeeniMc:
Superceded by PR #1013

PR #630, opened 2021-04-11
Upgrade to Gradle 7

nick-botticelli:
* The JCenter repository ([which will stop serving packages on February 1, 2022](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/)) is deprecated in Gradle 7, so it is removed in 0aa7275. I did not replace this with another repoistory (e.g. `mavenCentral()`) since there seem to be no dependencies that relied on it.
* `modCompile` is deprecated in Gradle 7, so it is replaced with `modImplementation` in c67d506.
* Gradle wrapper is updated to 7.0 in 3299a4f.

Personal testing showed that the command `gradlew clean build --warning-mode all --no-daemon` resulted in no real performance difference between 6.8.3 and 7.0 (37s and 38s). In best case scenario, [performance should increase for incremental builds](https://docs.gradle.org/7.0/release-notes.html#performance-improvements).

The built JAR runs fine on Minecraft 1.16.5 on my end.

I am open to any suggestions to improve this PR.

Environment:
* Windows 10 Enterprise (20H2)
* Intel® Core™ i7-8565U

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,6 +1,5 @@
 pluginManagement {
     repositories {
-        jcenter()
         maven {
             name = 'Fabric'
             url = 'https://maven.fabricmc.net/'
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     //to change the versions see the gradle.properties file
     minecraft "com.mojang:minecraft:${project.minecraft_version}"
     mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
-    modCompile "net.fabricmc:fabric-loader:${project.loader_version}"
+    modImplementation "net.fabricmc:fabric-loader:${project.loader_version}"
 }
 
 if (project.use_third_party_mods) {
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


DragonEggBedrockBreaking (approved):

settings.gradle
```diff
@@ -1,6 +1,5 @@
 pluginManagement {
     repositories {
-        jcenter()
```
i509VCB, line 3: So the million dollar question, are there any dependencies on jcenter we have to worry about?

DragonEggBedrockBreaking: nick already said:
> I did not replace this with another repoistory (e.g. mavenCentral()) since there seem to be no dependencies that relied on it.

Genau6502:
Doesn't loom need an update to 0.7-SNAPSHOT?

DragonEggBedrockBreaking:
> Doesn't loom need an update to 0.7-SNAPSHOT?

Loom 0.6-SNAPSHOT is enough to build on JDK 16, and is only necessary for a new specific tasks, for example the iris fork of sodium needs it because of the `publishToMavenLocal` task. As of the last announcement, 0.7-SNAPSHOT is technically beta, and if you go to [modmuss' website](https://modmuss50.me/fabric.html), Loom 0.6-SNAPSHOT is still the recommended.

jellysquid3:
Why is this issue being referenced dozens of times on other repositories? :confused:

jellysquid3:
Looks like @RDKRACZ is spamming many other repositories to update Gradle/Loom while referencing this issue in the process. I've gone ahead and blocked/reported them, which should put an end to the noise being added to this PR. We obviously don't encourage this kind of nonsense and I have no idea why they're going about spamming other people.

jellysquid3 (approved):
Looks good to me.

TelepathicGrunt:
> Looks like @RDKRACZ is spamming many other repositories to update Gradle/Loom while referencing this issue in the process. I've gone ahead and blocked/reported them, which should put an end to the noise being added to this PR. We obviously don't encourage this kind of nonsense and I have no idea why they're going about spamming other people.

The reason the person was doing that was because they wanted to build mod jars with java 16 (especially ones that have fixes in repo but not release yet) so that they can run Minecraft with Java 16 for performance boosts because they read in a gist that newer Java's have better garbage collectors. 

However, while they believed there were doing good and helping out a ton of people, they went and overdid it by mistake. But some folks educated the person on how to now clone repos and do the changes locally and can build with Java 16 instead of asking for the modder to do so. The person does understand what they did was wrong now so there should be no more repeats of that spam. I just wanted to provide context as to what happened in case anyone was still confused as to what the heck happened. Nothing intentionally malicious but keeping them block is still a valid choice.

PR #666, opened 2021-05-22
fix: Do not cull entities that are visible through chunks

bytzo:
### Summary
Currently, entities that are visible through chunks are not visible when they are inside of a culled chunk and the "Use Entity Culling" option is enabled.

This is because Sodium with the "Use Entity Culling" option will attempt to cull entities located inside of culled chunks, since they cannot be seen. However, entities that are visible through chunks are always visible to the player(such as entities with outlines or nametags), regardless if the chunk it is located in is visible or not. Sodium does not understand this, resulting in entities visible through chunks not being visible in some scenarios when the chunk they are in is culled.

This pull request adds an additional check to ensure entities visible through chunks are always marked as visible, in order to prevent them from being culled.

### Changes
- Moved all entity culling related mixins into the `features.entity.cull` mixin package.
- Added the method `shouldCullEntity`, which is true if the entity should be culled. Same result as `!isEntityVisible`, except will return `false` for entities with outlines or labels.
- The `shouldCullEntity` method is used for entity culling instead, preventing entities visible through chunks from being culled.
- When rendering entities, non-visible entities with labels and no outlines only render the label.
- When rendering entities, non-visible entities with outlines only render the entity and outline (and label if needed).

- Are the slight gains from this worth the extra mixins?

### Screenshots (intentionally see-through)
<h4>Without entity culling</h4>

<h4>With entity culling</h4>

<h4>With entity culling + this PR</h4>


FlashyReese (approved):

bytzo:
Marking as draft because there are more scenarios where entities are visible in culled chunks, such as with entity nametags. In this situation, it makes more sense to only cull the entity while still displaying the nametag.

src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
```diff
@@ -392,6 +392,11 @@ public boolean isEntityVisible(Entity entity) {
             }
         }
 
+        // Ensure entities with the glowing effect are not culled, since they are always visible to the player.
+        if (entity.isGlowing()) {
```
TheGlitch76: this check should run before the actual culling code

bytzo: Thanks for the feedback! However, I'm not sure if the order of visibility checks should matter much in this situation.

Putting the glowing visibility check earlier would allow entities to skip the chunk visibility check if they are glowing (since the method returns earlier), however there is no measurable performance improvement from doing so as checking for chunk visibility is not at all demanding. Even in ideal conditions (>1000 entities, all glowing), there is no measurable difference from making the change on my system.

I had placed this check at the end because it seems to be the least likely to happen. Usually, I try to order from most likely to least likely, which should allow the method to return earlier (this almost always provides no performance benefit- but I like to do it anyways :P)

Actually, now that I've thought about this a bit more, you may have mistaken this method for being something like `shouldCullEntity()` rather than `isEntityVisible()`, in which case you would be correct and the order does matter. However, since we only want to determine if the entity is actually visible (by returning true anywhere in the method), putting it before the chunk visibility check should not be necessary.

I could be completely missing something else here though, since I am relatively new to programming in Java. If so please let me know.

bytzo:
Spotted one last issue, marking as draft again.

bytzo:
Alright, that *should* be it. Sorry about all the issues.

src/main/resources/sodium.accesswidener
```diff
@@ -4,5 +4,7 @@ accessible class net/minecraft/client/model/ModelPart$Vertex
 accessible class net/minecraft/client/model/ModelPart$Quad
 accessible class net/minecraft/client/texture/Sprite$Interpolation
 
+accessible method net/minecraft/client/render/entity/EntityRenderer hasLabel (Lnet/minecraft/entity/Entity;)Z
```
FlashyReese: I think these access transformers can be avoided by just using mixins. You should be able to create an interface to bridge `EntityRenderer#hasLabel` and `EntityRenderer#renderLabelIfPresent` methods similar to `EntityLightSampler`.

src/main/java/me/jellysquid/mods/sodium/mixin/features/entity/cull/MixinWorldRenderer.java
```diff
@@ -0,0 +1,51 @@
...
+import me.jellysquid.mods.sodium.client.render.SodiumWorldRenderer;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.render.VertexConsumerProvider;
+import net.minecraft.client.render.WorldRenderer;
+import net.minecraft.client.render.entity.EntityRenderDispatcher;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.entity.Entity;
+
+@Mixin(WorldRenderer.class)
+public class MixinWorldRenderer {
```
FlashyReese: Can't we just mix this directly in `EntityRenderDispatcher`?

src/main/resources/sodium.accesswidener
bytzo: Thank you for telling me this, I was not aware that this was possible. Resolved with 62a42bf.

(Thanks, seriously. I was using access wideners all the time before this, and using Mixins is a much better solution I never had thought of. I clearly still have much to learn. )

src/main/java/me/jellysquid/mods/sodium/mixin/features/entity/cull/MixinWorldRenderer.java
bytzo: Yes that would be much better, thank you for the suggestion. Added with 62a42bf. Injecting into `EntityRenderDispatcher` also allows hitbox, shadow, and fire rendering to be canceled if needed, which was added in 62a42bf.

FlashyReese (approved):

jellysquid3:
I'm going to be delaying merging this as the code introduced by this is fairly complex and likely subject to change in Minecraft 1.17.

bytzo:
This PR is quite a mess, due to me making several mistakes and changing how always visible entities are culled halfway through. Currently, I am debating on two options:

1. Simple
Entities visible through chunks are marked to be always visible. This is how the PR was originally. The code required to do this is very little.
2. Complex
Entities visible through chunks are partially culled. Only the parts that are visible through chunks will render, such as the nametag or outline. This is how the PR is now. The code required to do this is more complex, requires a bunch of mixins, and more likely to break with newer versions of the game.

I think I will change this PR back to the simple approach, since there is little benefit in the complex one.

bytzo:
See #824.

PR #667, opened 2021-05-22
new: Add option to toggle fast random (#366)

### LINKED ISSUE #366: Option to make lilypads render with the same rotation as in vanilla minecraft
The rotation that lilypads are rendered in is inconsistent with the vanilla version of minecraft. This, for example, makes it difficult to use lilypads for hiding (or finding) secret codes that use this location-specific rotation. I have been told this discrepancy is caused by Sodium using a different random source for randomized block rotations.

It would be useful to have the option to disable this specific optimization.

A more general option to choose which random source to use would also help.

bytzo:
Sodium uses a different random source in its block renderer which is faster than vanilla. However, as a side effect, block rotations may be different when compared to vanilla, which can alter gameplay if players rely on the rotation of randomly rotated blocks. To solve this, an option to toggle the faster random function is added.

The name and description of this option may or may not need improvement. Suggestions for the wording and name of this option are appreciated. Currently, it is:

> **Use Fast Random**
> If enabled, a fast random function will be used for block rendering. This can affect the rotation of randomly rotated textures when compared to vanilla.

If this pull request is merged, translations would need to be updated.

This pull request resolves #366.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -305,6 +305,16 @@ public class SodiumGameOptionPages {
                         .build()
                 )
                 .add(OptionImpl.createBuilder(boolean.class, sodiumOpts)
+                        .setName("Use Fast Random")
+                        .setTooltip("If enabled, a fast random function will be used for block rendering. This can affect the rotation of " +
+                                "randomly rotated textures when compared to vanilla.")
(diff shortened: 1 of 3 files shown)


chexo3:
Maybe it would be better to use the vanilla random solely for edge cases like this? That way we can still benefit from the speed increase of Sodium’s random in cases that have less of an effect on things the player can see.

bytzo:
With recent additions to the Sodium options screen, this PR would cause some controls to go off of the screen on smaller displays. Also might want to consider random block models, like the random position of tall grass. Closing for now.

PR #672, opened 2021-06-04
Smaller blendedColors array in BiomeColorCache

FionaTheMortal:
Removed the padding by `radius` of the `blendedColors` array. Accessing these values would have caused an out of bounds error in `getColor(...)` anyway.


src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -32,18 +36,23 @@ public BiomeColorCache(ColorResolver resolver, WorldSlice slice) {
 
         this.dim = 16 + ((this.radius + 2) * 2);
 
+        this.blendedColorsMinX = origin.getMinX() - 2;
+        this.blendedColorsMinZ = origin.getMinZ() - 2;
+
+        this.blendedColorsDim = 16 + 2 * 2;
```
jellysquid3: `blendedColorsDim` should be moved to a static variable (such as `BLENDED_ARRAY_WIDTH`) since it's constant. This should also allow the JVM to better optimize the `getBlendedColor` call.

jellysquid3 (commented):
Minor nit-pick, but the solution here looks good to me otherwise.

jellysquid3:
Manually merged with https://github.com/CaffeineMC/sodium-fabric/commit/56dc005a9a4aea669a237e4bcf093057292d88ee because GitHub wasn't playing ball. Thanks for your contribution!

Issue #351, opened 2020-10-15
Game keeps crashing when using ReplayMod (sometimes not)


zxcdevlop:
Sodium doesn't work with ReplayMod. Whenever I try to render a video, the game freezes, until I have to force quit it. Sometimes I don't even have to run ReplayMod at all! It happens to me randomly when I'm playing single-player. Please, fix this problem. I want to continue using it.

altrisi:
This is a known issue with ReplayMod, currently it's not possible to render a video with Sodium installed, but you can still record it without problems. In theory, ReplayMod should notify you, but it seems that may not work.
See ReplayMod/ReplayMod#150
You just have to temporarily remove Sodium when rendering the video, but you can have it while recording.

If the game crashes without ReplayMod or in another context, then that's another issue.
Are you using the latest version of both?

dbechrd:
I've encountered this ReplayMod issue with other mods installed, even when not having Sodium installed. I had to disable all other mods to fix ReplayMod. This is almost certainly not Sodium's issue to fix.

Johni0702:
> I've encountered this ReplayMod issue with other mods installed, even when not having Sodium installed. I had to disable all other mods to fix ReplayMod. This is almost certainly not Sodium's issue to fix.

No, this issue is specific to ReplayMod+Sodium and we already know exactly what the cause is and what needs to be done to fix it (see the issue on ReplayMod's repo linked above).

There may of course be other compatibility issues, those are unrelated to this one though. If they do match the OP's description, then please report them cause I've only every heard of this exact issue with Optifine (which I have worked around a long time ago) and Sodium (which this issue exists for).

IMS212:
Fixed since Sodium 0.4.4.

PR #676, opened 2021-06-07
new: API to block until all dirty chunks have been updated (#351) 

Johni0702:
Split into two parts because I'm a lot less certain about the way the api framework should look than actually implementing the functionality.

# new: Framework to provide an API to third party mods

The API is separated into its own source set and we build a dedicated jar from that, so mods can depend on it without
depending on the entirety of Sodium (less to download/remap and no accidentally using the implementation classes):
```
dependencies {
    def sodiumVersion = "0.2.0+rev.972c7c1-dirty"
    modCompileOnly "me.jellysquid.mods:sodium-api:${sodiumVersion}"
    if (project.ext.use_sodium_in_dev) {
        modRuntime "me.jellysquid.mods:sodium:${sodiumVersion}"
    }
}
```
For the api jar to be properly remapped, we need to include a dummy fabric.mod.json. Mods should not include the API jar
at runtime (would be weird to have a sodium-api mod present without sodium, and picking the most recent one via the
version will not work for snapshot versions), so that json really is only used for loom.

Not sure what the best way to implement `SodiumApi.get()` is because the api source set cannot reference impl classes
and interfaces cannot contain mutable fields. The best I could come up with is having a semi-internal inner Impl class
with the mutable field.
The API implementation classes are generally (more noticeable in the following commit) separated from other
implementation classes because they - much like the Mixin classes - represent an interface to the outer world, and need
to notify the thingl RenderDevice about entering/leaveing the managed code, i.e. should not be called from other
internal code.

Bumps fabric-loom to 0.7 because 0.6 does not support the Gradle 7 + maven-publish combination (see
FabricMC/fabric-loom@9873153). Not going straight to 0.8
because that requires Java 16 to run.

#  new: API to block until all dirty chunks have been updated (#351) 

For background as to what this is for, see https://github.com/ReplayMod/ReplayMod/issues/150#issuecomment-623741826 and the reply.
I would expect this to be fairly uncontroversial, only real downside is that it's adding a bit of code which needs to be maintained but I'll gladly do that given the alternative is to hack into internals from the other mod via mixins.
The only change to behavior which applies even when the API is not in use, is the small addition of a `idleThreads` counter which is incremented/decremented before/after a worker thread goes to wait on the `jobNotifier`. This is required to make sure there are no more tasks currently being processed (cause those will neither be in the `buildQueue` nor the `uploadQueue`).

# Unrelated but noteworthy

Not included in or directly related to this PR (because it introduces a non-trivial change to existing behavior) but something I noticed while adding the `idleThreads` counter: The existing `isBuildQueueEmpty` method is used to replace a vanilla method which afaict is supposed to return whether all processing is done (i.e. the same thing which the new `isIdle` method returns), so `isBuildQueueEmpty` should probably be replaced with `isIdle` to more closely resemble the behavior of the vanilla method.

Also not included but something I noticed: There is a super tiny race condition in `getNextJob` where another job is added to the queue and a notify is sent, both after it is polled but before the wait. Rather easy to fix by just checking the queue a second time after entering the synchronized block before going to sleep, I just haven't opened a PR for it because that's guaranteed to conflict with this one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,10 @@
+import net.fabricmc.loom.task.RemapJarTask
+import net.fabricmc.loom.task.RemapSourcesJarTask
+
 plugins {
-    id 'fabric-loom' version '0.6-SNAPSHOT'
+    id 'fabric-loom' version '0.7-SNAPSHOT'
     id 'org.ajoberstar.grgit' version '4.1.0'
+    id 'maven-publish'
 }
 
 archivesBaseName = "${project.archives_base_name}-mc${project.minecraft_version}"
@@ -10,6 +14,21 @@ group = project.maven_group
 minecraft {
     refmapName = "mixins.sodium.refmap.json"
     accessWidener = file("src/main/resources/sodium.accesswidener")
+}
+
+sourceSets {
+    api {
+        java {
+            compileClasspath += main.compileClasspath
+        }
+    }
+
+    main {
+        java {
+            compileClasspath += api.output
+            runtimeClasspath += api.output
+        }
+    }
 }
 
 dependencies {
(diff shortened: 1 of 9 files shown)


MenacingPerson:
Is this still not merged? Can we atleast have the status of what needs to be done?

Leo40Git:
This will only be merged after 0.2 releases, since Jelly has enacted a _feature freeze_ so we can release 0.2.

Johni0702:
Changed to draft cause I'm really unhappy with the whole API part. I had a [small discussion on the Canvas tracker](https://github.com/grondag/canvas/issues/282) cause it has a very similar compatibility issue and I quite like the API grondag suggested, so I will probably convert this PR to provide the same, far nicer and simpler API, once that's stabilized.

Kroppeb:
I feel like the best way to achieve your goal, is to just force all render sections to be marked as important. That way Sodium will wait for them to finish before rendering a frame.

Johni0702:
> I feel like the best way to achieve your goal, is to just force all render sections to be marked as important. That way Sodium will wait for them to finish before rendering a frame.

No, it doesn't and it shouldn't. It will not block for important rebuilds if they are more than a few chunks from the player. That will also miss any chunks which have already been queued and those which are currently building (unless you're suggesting to re-mark them but then you're doing double the work).
I think I've got a pretty good understanding of Sodium's chunk renderer by now and don't think the actual implementation needs changing, it's just the whole API framework which I don't like.

jellysquid3:
I am admittedly not sure what the "standard" way for providing APIs as a Fabric mod is here. However, your solution here for blocking is straight forward and obvious, no complaints.

PR #690, opened 2021-06-16
fix #686: Clear cloud framebuffer when render distance changed on Fabulous graphics

### LINKED ISSUE #686: Setting render distance to 3 or lower in fabulous graphics makes clouds stick in place on screen
### Version Information

mc1.16.5-0.2.0-rc6

### Expected Behavior

Setting the render distance to 3 or lower is meant to hide clouds.

### Actual Behavior

The separate fabulous framebuffer for clouds is not cleared, leading to the same effect as #79 

### Reproduction Steps

Enable fabulous graphics and clouds
Set your render distance higher than 3
Look at clouds
Set your render distance to 3 or 2
Move the camera

### Other Information

Screenshot from a discord report:
https://cdn.discordapp.com/attachments/778685768041627719/853454610462933012/2021-06-12_19.02.31.png

brennanmcmicking:
This same issue appeared when the user would disable clouds on Fabulous graphics (#79) and was fixed in pull #80 by LucilleTea. Applying the same code to whenever a user changed their render distance resolves #686.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -35,7 +35,16 @@ public class SodiumGameOptionPages {
                         .setTooltip("The view distance controls how far away terrain will be rendered. Lower distances mean that less terrain will be " +
                                 "rendered, improving frame rates.")
                         .setControl(option -> new SliderControl(option, 2, 32, 1, ControlValueFormatter.quantity("Chunks")))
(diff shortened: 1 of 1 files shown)


jellysquid3:
I don't understand the solution here. Why does this problem only occur when changing the render distance? Does vanilla still render clouds even if the render distance is <=3 chunks?

coderbot16:
Vanilla does not render the skybox or clouds on render distances less than 4. The bug appears to be that the clouds framebuffer is only cleared when the render distance is 4 or greater. See GameOptions#getCloudRenderMode and the place that it's referenced in WorldRenderer.

PR #695, opened 2021-06-17
Curseforge now natively supports fabric

Aresiel:


jellysquid3:
Thanks for catching this.

PR #701, opened 2021-06-19
Fix worldheight changes.

Kroppeb:
Based on FlashyReese's work, but with the small fix to make stuff render when the player is above the worldheight.


jellysquid3:
This should be resolved now.

PR #710, opened 2021-06-25
fix: Prevent rendering sky when fog distance is reduced

### LINKED ISSUE #35: Fog occlusion hides chunks incorrectly with blindness effect
### Expected Behavior
Chunks within vision should not be culled

### Actual Behavior
Having fog occlusion enabled will cause chunks to disappear when the player is very close, but not inside them, while the blindness effect is on the player.

### Reproduction Steps
Apply blindness effect, be near a chunk border, have fog occlusion enabled.

### Attachments
https://cdn.discordapp.com/attachments/651120262129123330/726131134970462228/2020-06-27_03-05-46-1.mp4

### System Information
- Java Version: AdoptOpenJDK 14
- CPU: i7 3770
- GPU: GTX 970

### LINKED ISSUE #438: Visible chunks under water are being culled by Fog Occlusion
### Expected Behavior
Chunks that are visible underwater or visible while blind are not culled, since they are visible.

### Actual Behavior
Chunks far enough away are culled, causing it to look strange underwater compared to with that option disabled.

### Reproduction Steps
1. Turn Fog Occlusion on
2. Go underwater and observe. Chunks far from the player are not rendered.
3. Turn Fog Occlusion off
4. Observe again. Chunks are rendered properly.

### LINKED ISSUE #612: Improve fog occlusion
This could be a good fix for #438.

Have the client detect the current render distance, and instead of un-rendering underwater chunks that are usually not visible, display a flat color (see edited image.) I used paint.net to simulate the solid color instead of black. (2nd image)

I haven't considered any alternatives.

**Additional context**
Fog Occlusion On
Fog Occlusion On (Edited)

bytzo:
### Overview

When using fog occlusion, it is possible for the sky to be visible through culled chunks, ruining the illusion that there is fog. This is because vanilla relies on chunk fog to hide the sky when fog distance is reduced, however Sodium's chunk occlusion will cull these chunks as they are completely obscured by fog.

fog_occlusion_issue

To solve this issue without completely disabling chunk occlusion, the sky will no longer render when the fog distance is reduced from the default, since more dense fog means you cannot see the sky past it. Without the sky being rendered, the background of the world becomes the exact same as the fog color, so chunks culled by fog occlusion will now be indistinguishable from the fog. This behavior deviates from vanilla and can cause visual differences, however it solves some visual inconsistencies and matches Minecraft Bedrock Edition (it also seems to hide the sky when the fog distance is reduced). The most notable visual differences from vanilla include the sky not being visible from underwater and the End dimension's noise sky not being visible during the boss fight, both due to the fog being reduced.

This fixes #35, fixes #438, and fixes #612. It also fixes [MC-152504](https://bugs.mojang.com/browse/MC-152504) and [MC-230905](https://bugs.mojang.com/browse/MC-230905).

### Screenshots

| Sodium | Sodium + this pull request |
| - | - |
| flat_blindness_before | flat_blindness_after |
| underwater_before | underwater_after |
| end_before | end_after |
| ocean_before | ocean_after |


MeeniMc:
Just tried this, I love it for the massive visual improvement it brings in when under water (w.r.t. normal Sodium)! 

The loss of the sky rendering in the end is a bit more disruptive than from under water. The  'thick-fog'  effect removes the 'random noise' end sky from view completely. It does however solve the jagged transition between rendered / non-rendered chunks that Sodium also causes in this dimension otherwise, just wish one could have pie and eat it too. 

The 'thick-fog' behavior doesn't result in any discernible rendering difference w.r.t. vanilla in the nether.

wb1016:
used about a week. no bugs directly related encountered.
had some issues with a utility mod that features FreeCam(moving view point without moving actual player position), but it is because they use non-standard rendering. 
according to log, when i enter to liquid in freecam mode, the utility mod write error messeage on the log and halts game.
not sure if it only this branch affects but still it is their issue, not sodium.

PR #716, opened 2021-06-27
Non-Inclusive Fabric API Compatibility

Leo40Git:
Ignore the kinda iffy name, this PR _should_ allow FAPI compatibility without actually including any FAPI modules ~~and getting the speedrunning community in a tizzy.~~

Currently supported:
- `fabric-rendering-fluids-v1`
- `fabric-rendering-data-attachment-v1`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,10 +16,12 @@ minecraft {
 }
 
 configurations {
-    modIncludeImplementation
+    modFAPI
 
-    include.extendsFrom modIncludeImplementation
-    modImplementation.extendsFrom modIncludeImplementation
+    if (project.use_fapi)
+        modImplementation.extendsFrom modFAPI
+    else
+        modCompileOnly.extendsFrom modFAPI
 }
 
 dependencies {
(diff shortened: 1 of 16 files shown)


MulverineX:
There is already a project working on this: https://github.com/comp500/Indium
For now it's a seperate mod that integrates with Sodium, and while it's **not** perfect, it's pretty far along.

jellysquid3:
It's my understanding this PR is in regards to the support Sodium recently gained for some FRAPI modules. Though what exactly it's doing is beyond me.

Leo40Git:
this _will_ essentially be a "rewrite" of sorts of the new compat stuff so it won't depend om FAPI beimg included

keyword being will because this is just a placeholder so i don't forget

mrmangohands:
For the record we're generally fine with the bundled module approach :P

Leo40Git:
We have actual code now, hooray!

Unfortunately this hasn't been tested yet because `sun.misc.Unsafe` isn't resolving during compilation (what).

Leo40Git:
Okay, I've tested it - fluid rendering seems to work in both cases.

PR #719, opened 2021-06-28
fix: Don't try to join the results of cancelled futures on 1.17

coderbot16:
This fixes a crash on 1.17 when futures for building nearby chunks are cancelled.

While it's not optimal, it provides an immediate fix for the crash that is hopefully unlikely to cause other issues.

Thanks to Kroppeb for pointing out the issue.


arceuss:
i can finally live in peace without restarting minecraft 50k times after a crash now

Issue #394, opened 2020-11-16
Waterlogged stairs cause z-fighting


LucilleTea:
### Expected Behavior
Water should not be visible through the back of the stair.

### Actual Behavior
The water and the stair z-fight. This happens regardless of the Compact Vertex Format, or Chunk Renderer, or any other setting in the advanced tab (and does not happen without sodium).

### Reproduction Steps
Waterlog some stairs, move >30 blocks away from them, facing their back (for me, Jokbon's screenshot doesn't look quite as far).

### Attachments
![2020-11-16_12 00 38](https://user-images.githubusercontent.com/64509677/99203720-5de16e80-2803-11eb-8598-339e865e7345.png)

![jokbon](https://media.discordapp.net/attachments/602805020861267968/777695754885529620/2020-11-16_02.44.39.png?width=1667&height=937)

### System Information
- Java Version: AdoptOpenJDK Java 15 64-bit
- CPU: i7 3770
- GPU: GTX 1070
- OS: Windows 10
Using Sodium 0.1.1-SNAPSHOT in my case, Jokbon seems to be on the stable release.

CalXee:
Isn't this a vanilla bug? Are more sides of stairs Z fighting, or are they Z fighting more severely?

LucilleTea:
I could not reproduce it in the same situation without sodium running, like I said. It may have happened in previous versions or in different circumstances.
Edit: Just confirmed with true vanilla 1.16.4, i don't see the issue there either. So that rules out Indigo fixing it.

CodingRays:
While dealing with a similar bug in iris (https://github.com/IrisShaders/Iris/issues/242) i found the code that causes this. 
In sodiums `FluidRenderer.render` function while testing if sides are exposed the `isFluidExposed` function only tests if the next block contains the same fluid as the current block. Vanillia minecraft has an additional test for self occlusion which is implemented in `FluidRenderer.method_290710`.

MeeniMc:
Issue is resolved in 0.4.10 (presumably since pr above merged)

programaus:
> Issue is resolved in 0.4.10 (presumably since pr above merged)

This actually still persists

![Image](https://github.com/user-attachments/assets/1e649801-5754-4cc6-92e6-3e2f1d4e1fbf)
1.20.1 fabric, 0.5.11, indium 1.0.34

douira:
[...]

PR #720, opened 2021-06-29
Add self occlusion test for waterlogged blocks

CodingRays:
I added a self occlusion test for fluids inside waterlogged blocks for all directions except the up direction. This is already implemented in vanilla minecraft.

This should somewhat remedy #394 however for blocks that only partially fill the face it will still be an issue. This has also been an issue in iris see https://github.com/IrisShaders/Iris/issues/242


jellysquid3 (approved):

src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java
```diff
@@ -71,8 +71,18 @@ public FluidRenderer(LightPipelineProvider lighters, BiomeColorBlender biomeColo
...
 
-    private boolean isFluidExposed(BlockRenderView world, int x, int y, int z, Fluid fluid) {
-        BlockPos pos = this.scratchPos.set(x, y, z);
+    private boolean isFluidExposed(BlockRenderView world, int x, int y, int z, Direction dir, Fluid fluid) {
+        // Up direction is hard to test since it doesnt fill the block
+        if(dir != Direction.UP) {
+            BlockPos pos = this.scratchPos.set(x, y, z);
+            BlockState blockState = world.getBlockState(pos);
+            VoxelShape shape = blockState.getCullingShape(world, pos);
+            if (VoxelShapes.isSideCovered(VoxelShapes.fullCube(), shape, dir.getOpposite())) {
```
Altirix, line 80: kelp uses the fullCube shape so the occlusion test causes #769 #782

if it doesnt affect anything in iris this could be fixed by doing `if (blockState.getBlock() != Blocks.KELP_PLANT && VoxelShapes.isSideCovered(VoxelShapes.fullCube(), shape, dir.getOpposite())) {`

Altirix: `if (blockState.isOpaque() && VoxelShapes.isSideCovered(VoxelShapes.fullCube(), shape, dir.getOpposite())) {` looks to be more ideal. as this should then work for any mods that add in transparent fullCube blocks.

(also this still wont occlude slabs where the water is z fighting i think the fluid renderer would need to be aware of the blocks voxelshape to do much more)

PR #723, opened 2021-06-30
Convert issue templates into forms

altrisi:
This PR adds some of the brand new [Github Issue Forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) for the Sodium repo.

Issue forms are similar to templates in a way, but they are a lot more end-user friendly, since they have proper inputs, textboxes, helper texts, placeholders, selections and checkboxes, and everything gets converted to a nice issue when being submitted.

Not only that, they also allow making some fields required, which is something quite useful IMO to prevent people just skipping the whole template and just writing something like "the game crashes" at the bottom. It also doesn't include the helper texts when the issue is submitted, only the titles and content, making it faster to read even if the user forgot to remove those parts of the templates.

This PR converts the bug report and crash report templates into issue forms, having the same or more fields than those had.

The only real downside from those is that you can't customize the header's size and every input has a header. But I think requiring users to provide useful information outweighs those problems (of course users can still just put junk in those required fields, but it's some filtering at least).

You can test those issue forms on the [issue tracker of my sodium fork](https://github.com/altrisi/sodium-fabric/issues/new/choose), feel free to spam it with tests. Github also provides some kind-of preview if in the Files Changed tab you click on the `View File` button in the top right.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,112 @@
...
+        look at the file name in your `mods` folder.
+      placeholder: 'Example: mc1.16.4-0.1.1+rev.4b3964a'
+    validations:
+      required: true
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected Behavior
+      description: What did you expect to happen?
+      placeholder: 'Example: The Piston does not extend.'
```
SirGavith: Probably should be 'Example: The Piston extends'

altrisi: Fixed by changing it to the previous example.

jellysquid3:
These look great. Thank you for putting it all together. I would expect this to greatly decrease the number of invalid reports we see.

PR #727, opened 2021-07-02
new: Batch GUI rendering

lunbun:
In vanilla Minecraft, each texture of a status bar (health bar, hunger bar, etc...) is drawn separately, which can result in up to 40 draw calls. This PR batches all status bar rendering into a single draw call.

On my system, vanilla takes around 0.3 milliseconds to render status bars, but with this, it drops to around 0.016 milliseconds.


Leo40Git:
I can see mods breaking very easily with this change, and this seems like a bit of a micro-optimization to me.

xanderstuff:
I can confirm this optimization has a **non-negligible** performance increase. 
### Here is my experiment I performed:

I compiled and compared **this branch**'s performance (commit 3d67782) with **sodium 0.3.0+build.458** from GH actions. To more closely measure the difference in time it takes to render the status bars, I recorded max fps observed (minimum frame time) both for creative mode and survival mode (creative mode hides status bars - I am unsure if there are other rendering differences between creative mode and survival mode).
I stood in one place, without rotating my camera, and waited at least 30 seconds to see the highest fps. I switched gamemodes quickly with F3+F4 and cleared chat with F3+D. I used a multiplayer server so that disconnecting still left the world running. When entering a world, I slowly rotated a full 360 degrees to see/load all chunks, then moved to specified cam rotation. I was above ground in the overworld. I also set difficulty to peaceful to eliminate hostile mobs.

### Variables I made sure were constant between each measurement:
minecraft in windowed mode, render distance set to 8, ALT+F3 screen open, coords the same, camera rotation the same, 50% entity render distance, F3 screen shows "E: 0/53" and "C:253" (+/-2), health and hunger at full.
other info:
windows 7, MC 1.17, GPU: GTX 560Ti, 391.35 driver, CPU: AMD FX8150
seed: -81696447260766844, coords: 1708, 85, 4489, facing: -142, 27

### Results:
**Sodium 0.3.0+build.458:**
creative mode: max 296 fps (min 3.38ms frame time)
survival mode: max 249 fps (min 4.02ms frame time)
    **creative->survival frame time delta: +0.64ms**

**Sodium, lunbun:batch-status-bars, commit 3d67782:**
creative mode: max 301 fps (min 3.32ms frame time)
survival mode: max 286 fps (min 3.50ms frame time)
    **creative->survival frame time delta: +0.18ms**

note: I exited and reopened the world once and got consistent results.
I switched between creative/survival at least 4 times each, waiting at least 30 seconds for max fps, and got consistent results.
I repeated this whole experiment twice, (so build458, batch-status-bars, then build 458 again, then batch-status-bars again), and got consistent results.


So basically this optimization seems to improve whatever extra's rendered between creative mode and survival mode by (0.64ms - 0.18ms) = 0.46ms for me, which **represents vaguely 10%ish of my total frame times** in this test scene (IMO a non-negligible amount). This is similar to lunbun's claims.

rdrpenguin04:
I think this should be investigated with mods that update the status bar before being outright dismissed; it seems like a useful feature if it is proven to not affect mods.

This also seems like a good opportunity to expose an API to add things to the status bar without needing copious mixins necessarily; then, Sodium could hijack the API instead of code that other mods would be likely to hijack themself.

lunbun:
After further testing, it seems that the performance improvements this gives varies a lot. On my integrated GPU, this only increases the framerate from about 99 to 103. However, on my dedicated GPU, this increases the framerate from about 174 to 215.

It may be possible to include this optimization while being compatible with other mods by not directly overriding status bar rendering, which can be seen in 54491bc2e46c2a3def57e627ceea2849af4f58c3. This does break the contributing rules a little bit though since the mixins are not self-contained, so a better solution may be required.

xanderstuff:
however, 3d67782 does indeed have issue with mod compatibility. 

`appleskin-fabric-mc1.17-2.0.0` works just fine with `Sodium 0.3.0+build.458`,
but when I use this branch of sodium (3d67782), it will go to main menu, and when I enter a world it crashes:
`java.lang.IllegalStateException: Already building!` (see rest of crash report. I am using notenoughcrashes as well).
[crash-2021-07-03-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/6758050/crash-2021-07-03-client.txt)

Rendering stuff is a bit over my head, so I can't comment on ways to fix this.

xanderstuff:
ok so the good news is 54491bc doesn't crash with `appleskin-fabric-mc1.17-2.0.0`. The bad news is it's horribly broken.


(there should be 16/20 food, and 6.2/20 saturation, according to `/data get entity @s foodSaturationLevel`)

And rendering looks correct/normal after removing appleskin.

@lunbun could you do some more investigating to try and get appleskin to work? 
https://github.com/squeek502/AppleSkin/tree/1.17-fabric
https://www.curseforge.com/minecraft/mc-mods/appleskin/files/3358016

lunbun:
Appleskin adds additional textures that are being used to render the status bar, but this optimization requires that only the vanilla textures (or textures loaded with resource packs) are used to render status bars. With the current way that textures are batched, getting appleskin to work would require basically disabling the whole optimization.

lunbun:
I'm not sure what appleskin is supposed to look like normally, but it looks like its rendering correctly in commit 5f845c237202efb4e92dcd6a124b5f0b706ebdc8 without having to disable the optimization.

Leo40Git:
Sorry for being dismissive the first time around, this looks good.

xanderstuff:
I haven't gotten the chance to try out your "Batch render fps pie 8390309" commit, so idk how much of a difference it makes on my hardware, but would something similar be good for the fps and tps graphs as well? (ALT+F3)
If so, may want to check compatibility with [servertick-fabric](https://github.com/sammko/servertick-fabric) ([modrinth](https://modrinth.com/mod/servertick))

lunbun:
I believe that vanilla already batches the fps and tps graphs.

andreasdc:
I confirm that this change is huge improvement

jellysquid3:
I have implemented this and many other batching related optimizations for GUIs based on the user feedback for this PR. Most of this is now available in `0.4/experiments` and includes:
- Status bar icon batching (health, armor, bubbles, etc)
- Item render, durability bar, count, and glint batching for both the HUD and inventories
- Text rendering batching for some situations (i.e. sign rendering)
- And some other small stuff that I can't recall.

The implementation is substantially different from what's described in the PR for mod compatibility reasons & so we can batch additional parts of the UI, but it is really building upon the ideas this PR originally introduced and then some. As such, thank you for your contribution, as otherwise it would not have led to this line of development.

Regarding the FPS pie graph optimization, this is very interesting: But I have to wonder if it's worth optimizing a debug feature that a very small numbers of users know about, and are almost never playing the game with open.

andreasdc:
> I have implemented this and many other batching related optimizations for GUIs based on the user feedback for this PR. Most of this is now available in `0.4/experiments` and includes:
> 
> * Status bar icon batching (health, armor, bubbles, etc)
> * Item render, durability bar, count, and glint batching for both the HUD and inventories
> * Text rendering batching for some situations (i.e. sign rendering)
> * And some other small stuff that I can't recall.
> 
> The implementation is substantially different from what's described in the PR for mod compatibility reasons & so we can batch additional parts of the UI, but it is really building upon the ideas this PR originally introduced and then some. As such, thank you for your contribution, as otherwise it would not have led to this line of development.
> 
> Regarding the FPS pie graph optimization, this is very interesting: But I have to wonder if it's worth optimizing a debug feature that a very small numbers of users know about, and are almost never playing the game with open.

Don't ban people on discord for nothing first of all

PR #753, opened 2021-07-12
Update refrences to java to support Java 16 and 1.17

ghost:


amnotbananaama:
Thank you for the PR. At the moment, 1.17.x builds are for development purposes only. 

Users trying to compile bleeding-edge versions of Sodium themselves should be aware already that they need to use Java 16.

When Sodium is officially released for 1.17.x, users will need to compile with Java 16, not Java 8+ as your PR indicates. For these reasons, I'm going to close it at this time

PR #785, opened 2021-07-19
Fix dirty issue by ignoring macOS files (like .DS_Store)

douira:
macOS generates `.DS_Store` and other files automatically in many folders when Finder is used. Ignoring them in gitignore resolves the issue where builds on macOS are considered dirty. (Running `git status` will reveal the presence of an untracked `.DS_Store`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,9 @@ bin/
 # fabric
 
 run/
+
+# macOS
+
+.DS_Store
+.AppleDouble
+.LSOverride


sylv256 (approved):

caoimhebyrne (approved):

douira:
As others have noted in the Iris Discord channel, OS-specific exclusions *should* be handled through a global exclusion file setting in git. However, I think adding the most common files to gitignore in projects themselves helps reduce headaches from fiddling with dirty git indexes. That's why this PR (and the equivalent PR to iris) exist.

Leo40Git (approved):
These should be standard. I'm surprised IntelliJ and the like don't have a "hey you forgot to add OS-specific BS to your gitignore, wanna add it?" notification or something.

PR #791, opened 2021-07-23
new: Implement FREX Flawless Frames API (#351) (1.16)

Johni0702:
> Implements the "Flawless Frames" FREX feature using which third-party mods can instruct Sodium to sacrifice
 > performance (even beyond the point where it can no longer achieve interactive frame rates) in exchange for
 > a noticeable boost to quality.
 >
 > In Sodium's case, this means waiting for all chunks to be fully updated and ready for rendering before each frame.
 >
 > See https://github.com/grondag/frex/pull/9

Supersedes https://github.com/CaffeineMC/sodium-fabric/pull/676.

Implementation of the blocking update is identical to the old PR but the whole API part has been replaced with [this FREX one](https://github.com/grondag/frex/pull/9) (doesn't actually require FREX, it's just defined there as a standard which multiple mods can implement) which is much simpler, nicer and more general at the same time.

Unlike the previous API, this one requires that all visible chunks be fully updated (the previous one only required all **pending** ones to be updated, and the caller would have to loop until the full chunk graph is visible). So the looping is now done in Sodium (see `MixinWorldRenderer`). This loop isn't ideal (we re-setup many thing which could not have changed) but it's good enough to get the whole thing working and performance is secondary for Flawless Frames.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
@@ -1,6 +1,7 @@
 package me.jellysquid.mods.sodium.client;
 
 import me.jellysquid.mods.sodium.client.gui.SodiumGameOptions;
+import me.jellysquid.mods.sodium.client.util.FlawlessFrames;
 import me.jellysquid.mods.sodium.client.util.UnsafeUtil;
 import net.fabricmc.api.ClientModInitializer;
 import net.fabricmc.loader.api.FabricLoader;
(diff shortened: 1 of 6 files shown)


Johni0702:
It's been almost a year. Any chance this is ever getting merged? (there are also branches for all the other MC versions, just didn't bother opening PRs for them yet)

Johni0702:
Just gonna bump this every month now until I get an answer... This has been used in production for more than a year now, with no issues whatsoever. What is the concern that holds you back from clicking that merge button? Especially for the newer versions where there have been a lot of sodium releases in the mean time which could have included it.

Issue #280, opened 2020-08-24
Fix vanilla fast leaves improper culling


Jetsparrow:
There is a long-standing performance regression in vanilla, from 1.12:
[Leaves are not culled when set to fast](https://bugs.mojang.com/browse/MC-179383) 
Specifically, even completely concealed leaves with opaque textures are not culled.
This means that fast leaves don't offer the performance benefit they should.
This can be verified by building leaves adjacent to a wall, and then going inside in spectator mode.
![2020-08-24_17 12 43](https://user-images.githubusercontent.com/37241560/91055816-d6ce9100-e62d-11ea-9d4d-770c340a5d2f.png)

Mojang seem unwilling to fix it. [The same issue was reported back in 2015 and resolved without a fix.](https://bugs.mojang.com/browse/MC-85132)

A fix would greatly improve performance on lower-end devices in forested areas.

OpenJDK; Lithium 0.5.1; Phosphor 0.6.0+build.7; Sodium 0.1.0;

These screenshots demonstrate the slowdown. All screenshots taken right after reloading graphics (f3+t)

Single biome taiga map, 24 chunks view distance, no random ticks or mob spawns. 
![2020-08-24_16 24 14](https://user-images.githubusercontent.com/37241560/91053671-d503ce00-e62b-11ea-9597-d38a4bf1239b.png)
![2020-08-24_16 24 41](https://user-images.githubusercontent.com/37241560/91053686-dcc37280-e62b-11ea-8956-f5705f4ede9c.png)
 
Same map, but with leaves replaced by stone, which provably culls properly.
![2020-08-24_16 42 18](https://user-images.githubusercontent.com/37241560/91053700-e056f980-e62b-11ea-9570-d0925e5e6852.png)
![2020-08-24_16 42 41](https://user-images.githubusercontent.com/37241560/91053706-e351ea00-e62b-11ea-94fa-82444b127e8a.png)

Coffee-for-Cats:
I don't remember if this configuration is present in Sodium, but optine has "smart" as leaves configuration.
I think it would make this problem less significant.

Coffee-for-Cats:
Okay, and we have this comment too: https://github.com/CaffeineMC/sodium-fabric/issues/233#issuecomment-668656707

sh-lucas:
Transparency is probably not an issue: https://github.com/CaffeineMC/sodium-fabric/issues/233#issuecomment-668656707
However, cutting the amount of vectors would help

[...]

PR #810, opened 2021-07-31
Cull inside faces of leaves on Fast graphics

Leo40Git:
As the title says, this PR makes the inside of leaves blocks (I.E any block that inherits from `LeavesBlock`) be culled, but only on Fast graphics.


src/main/resources/sodium.mixins.json
```diff
@@ -38,6 +38,7 @@
     "features.entity.smooth_lighting.MixinEntityRenderer",
     "features.fast_biome_colors.MixinBackgroundRenderer",
     "features.fast_biome_colors.MixinClientWorld",
+    "features.fast_leaf_culling.MixinLeavesBlock",
```
jellysquid3: This isn't about making leaf culling "fast", it's about fixing a bug. We should probably toss this and the other injection for RenderLayers into`features.render_layers.leaves` so that it can be disabled by config if problems arise (there are other mods which try to do this very thing...)

src/main/java/me/jellysquid/mods/sodium/mixin/features/fast_leaf_culling/MixinLeavesBlock.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Mixin(LeavesBlock.class)
+public class MixinLeavesBlock extends Block {
+    public MixinLeavesBlock() {
+        super(Settings.of(Material.AIR));
+        throw new AssertionError("Mixin constructor called!");
+    }
+
+    @Override
+    public boolean isSideInvisible(BlockState state, BlockState stateFrom, Direction direction) {
```
jellysquid3: `@Override` -> `@Overwrite` with reason, please. This makes it more explicit what's going on.

src/main/java/me/jellysquid/mods/sodium/mixin/features/fast_leaf_culling/MixinLeavesBlock.java
```diff
@@ -0,0 +1,29 @@
...
+        if (!(SodiumClientMod.options().quality.leavesQuality.isFancy(MinecraftClient.getInstance().options.graphicsMode))) {
```
jellysquid3: Minor nitpick: We shouldn't use an inverted if-statement if we're just taking one of two branches... It makes the code easier to read if the logic is instead "If fancy leaves enabled, use default culling logic".

jellysquid3 (changes requested):
Small changes requested. Otherwise the implementation looks good. I'm happy to merge this in for Sodium 0.3.3.

src/main/java/me/jellysquid/mods/sodium/mixin/features/fast_leaf_culling/MixinLeavesBlock.java
Leo40Git: The problem is, there's no `isSideInvisible` to overwrite in `LeavesBlock`!

Leo40Git:
What should the impact for this option be? Admittedly I don't recall if I tested it.

src/main/java/me/jellysquid/mods/sodium/mixin/features/fast_leaf_culling/MixinLeavesBlock.java
jellysquid3: Ah... you're right. That's my fault. Apologies.

jellysquid3:
It's highly variable and depends on the amount of foliage, so I feel "Medium" is a good compromise. Rendering less faces, even if they're mipped and far away, still provides a significant boost at times.

src/main/java/me/jellysquid/mods/sodium/mixin/features/render_layer/MixinRenderLayers.java
```diff
@@ -29,4 +36,11 @@
         BLOCKS = new Reference2ReferenceOpenHashMap<>(BLOCKS);
         FLUIDS = new Reference2ReferenceOpenHashMap<>(FLUIDS);
     }
+    @Inject(method = "getBlockLayer(Lnet/minecraft/block/BlockState;)Lnet/minecraft/client/render/RenderLayer;", at = @At(value = "RETURN"), cancellable = true)
```
jellysquid3: Can we move this injection into `features.render_layer.leaves` package as well? I don't want to mix bug fixes with the existing optimizations here.

jellysquid3 (approved):

Jetsparrow:
Solves https://github.com/CaffeineMC/sodium-fabric/issues/280

Granted, this was measured using a different method of leaf culling, but the effect should be huge. 
Very happy to see this merged

PR #814, opened 2021-08-02
Add Jitpack config to use Java 16

Leo40Git:
This allows other mods to access Sodium at compile-time, thus allowing compatibility!


FlashyReese (approved):

Leo40Git:
I guess this isn't necessary anymore now that we have a proper Maven, huh?

jellysquid3:
We don't really intend to support Jitpack, as it's highly problematic even when it works. In the future, releases of Sodium will be published to our own Maven.

Issue #817, opened 2021-08-04
Rendering of resource pack models does not occur at certian angles


Altirix:
### Version information

sodium-fabric-mc1.17.1-0.3.0+build.5.jar

### Expected Behavior

**Vanilla rendering**

cullleaves-2.2.0.jar
fabric-api-0.36.0+1.17.jar
![2021-08-04_14 04 10](https://user-images.githubusercontent.com/19439141/128240357-813a879a-266a-4a33-8f09-18c774298073.png)

### Actual Behavior

**Sodium rendering**

cullleaves-2.2.0.jar
fabric-api-0.36.0+1.17.jar
indium-1.0.0.jar
sodium-fabric-mc1.17.1-0.3.0+build.5.jar
![2021-08-04_14 05 12](https://user-images.githubusercontent.com/19439141/128241162-dd9fbc90-70cd-4e8e-ae72-db37aada7792.png)

### Reproduction Steps

**Version - 1.17.1**

**Required Mods:**
cullleaves-2.2.0.jar
fabric-api-0.36.0+1.17.jar
indium-1.0.0.jar
sodium-fabric-mc1.17.1-0.3.0+build.5.jar

Enable smart leaves in resource packs
observe as vanilla renders all models in view distance while sodium only renders models in the current chunk.

### Java version

Java 16

### CPU

R7 5800x

### GPU

RTX 3080

### Additional information

refers to https://github.com/TeamMidnightDust/CullLeaves/issues/11

this is the only resource pack ive noticed this behaviour on. other resource packs with models render just fine.

Altirix:
Thinking about it. culling the leaves uses isSideInvisible which is for glass in vanilla. so maybe this is related to the translucency sorting?

Rajbir98:
this is why I play in fast graphics

Altirix:
turns out this is present on all packs. it seems to affect all models that fall under the conditions to trigger it. im not sure what exactly causes it but you can easily see it on the Y axis also 3rd person view.

https://cdn.discordapp.com/attachments/752637878462578778/872921788349300797/Base_Profile_2021.08.05_-_19.38.14.09.DVR.mp4

Altirix:
#732

MeeniMc:
still present as-of 0.4.10, related to #732

pajicadvance:
Fixed in 0.5.0.

Issue #732, opened 2021-07-03
Bushy leaves culled when they are still visible


400Ferrets:
### Version: 

> - sodium-fabric-mc1.17-0.3.0+build.469

### Expected Behavior:

> - Bushy leaves from vanilla tweaks texture pack with them enabled stick around whenever you can see them

### Actual Behavior:

> - Previously mentioned bushy leaf bits on the bottoms of blocks disappear even though you can still see them

### Reproduction Steps:

1. Be using bushy leaves in vanilla tweaks texture pack
2. Locate tree
3. a couple blocks away from tree jump up and down and watch your flickering bushy leaves

### System Information:

> - Java Version: 16.0.1 64bit
> - CPU: 4x Intel(R) Xeon(R) CPU E5-1607 v2
> - GPU: AMD R9 270

### Other Information:

In attached photo the LEFT side shows screenshots using sodium, and the RIGHT side show screenshots using vanilla 1.17
![bushyleafculling](https://user-images.githubusercontent.com/68585283/124352169-5b7e2a00-dbc4-11eb-8e40-b94b951f61df.png)

used texture pack, you can also make a new one at [Vanilla Tweaks](https://vanillatweaks.net/picker/resource-packs/) make sure to select bushy leaves
[VanillaTweaks_r609822.zip](https://github.com/CaffeineMC/sodium-fabric/files/6758277/VanillaTweaks_r609822.zip)

jellysquid3:
Does disabling "Use Block Face Culling" in Video Settings > Advanced fix this problem?

400Ferrets:
> Does disabling "Use Block Face Culling" in Video Settings > Advanced fix this problem?

It does not.

pajicadvance:
Fixed in 0.5.0.

PR #822, opened 2021-08-06
Fix to allow block face culling to be disabled

Altirix:
sodium had an option to toggle block face culling but it was never used in sodium 0.3. disabling block face culling is required for resource packs that use models

#817
#732

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RegionChunkRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RegionChunkRenderer.java
@@ -1,6 +1,7 @@
 package me.jellysquid.mods.sodium.client.render.chunk;
 
 import com.google.common.collect.Lists;
+import me.jellysquid.mods.sodium.client.SodiumClientMod;
 import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexAttributeBinding;
 import me.jellysquid.mods.sodium.client.gl.buffer.GlBufferUsage;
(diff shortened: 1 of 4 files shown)


jellysquid3:
Oh, you're right. I'm not sure how this happened. Looks like a blunder on my part.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RegionChunkRenderer.java
```diff
@@ -127,38 +128,47 @@ private boolean buildDrawBatches(List<RenderSection> sections, BlockRenderPass p
...
-            ChunkRenderBounds bounds = render.getBounds();
-
             long indexOffset = state.getIndexSegment()
                     .getOffset();
 
             int baseVertex = (int) state.getVertexSegment()
                     .getOffset();
 
-            this.addDrawCall(state.getModelPart(ModelQuadFacing.UNASSIGNED), indexOffset, baseVertex);
+            if (SodiumClientMod.options().advanced.useBlockFaceCulling) {
```
jellysquid3: This should be moved to a `private final` variable, and initialized once when `RegionChunkRenderer` is created. If this setting changes, the renderer will be re-created, so it should be safe.

jellysquid3 (changes requested):
Minor nitpick, otherwise looks fine.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RegionChunkRenderer.java
Altirix: I've done this change in my latest commit. thank you

PR #823, opened 2021-08-06
change: Combine cloud visibility and quality options

bytzo:
This pull request combines the cloud visibility toggle and cloud quality option into one control. It also changes Sodium to use the vanilla cloud option instead of its own, since it is no longer necessary to store them in a different way.

Currently, a user may navigate to the Quality tab and assume there is no option to disable clouds. This is because vanilla uses one option for cloud visibility and quality, while Sodium does not. By implementing this change, the option to disable clouds will be with the cloud quality option, matching vanilla and helping avoid confusion.


jellysquid3:
I agree with your points here. This is a large source of confusion for players and there's no reason for us to have the option anymore... and I'm always a fan of trimming out fat in the mod.

jellysquid3 (approved):
Looks good to me. Thanks for the contribution. I will look into merging this in the near future.

PR #824, opened 2021-08-06
fix: Ensure entities with outlines or nametags are always visible

bytzo:
Currently, entities that can be seen through blocks (such as ones with the glowing effect or nametags) are not visible when in culled chunks. This is because Sodium culls entities in culled chunks.

This pull request adds an extra condition that marks entities as visible when they have the glowing effect or nametags, which should prevent them from being culled in culled chunks. This also repackages `me.jellysquid.mods.sodium.mixin.features.entity.smooth_lighting` to `me.jellysquid.mods.sodium.mixin.features.entity.cull`, since the Mixin has nothing to do with smooth lighting anymore.

This PR is the original implementation of #666, updated for 1.17. See #666 for more info and [this comment](https://github.com/CaffeineMC/sodium-fabric/pull/666#issuecomment-894066555) on why the more complex changes were dropped.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -319,6 +319,11 @@ public class SodiumWorldRenderer implements ChunkStatusListener {
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


jellysquid3:
I've merged this as 8944df6f.

PR #868, opened 2021-08-29
Fix kelp blocks to make them render water again

Altirix:
fixes edge case in #720 

Before:
image

After:


K0-RR:
Isn't this already fixed in v0.3.1? Part of the changelog states"
"Fixed an issue where some blocks wouldn't render their waterlogged state against glass."

Altirix:
nah, this didnt just affect glass. the water will not render if any block is placed to that side of the kelp. you can see this is a few youtubers videos while they make water elevators

mrpenguinb:
The issue can be encountered when a door or ladder is nearby
Kelp

PR #871, opened 2021-08-30
added extra file handling

Altirix:
`try (JsonReader reader = new JsonReader(new FileReader(path.toFile()))) {`

without using JsonReader could cause a MalformedJsonException if the file was corrupt

`if (!file.canWrite() && file.exists()){throw new IOException("Sodium Config is read-only.");}`

without checking if the config is readonly the Files.move will cause AccessDeniedException


Altirix:
eabd900 and onwards changes throwing a RuntimeException to instead overwrite the corrupt config to allow the game to boot and will display this message in the settings page 
image

jellysquid3:
This has been fixed with https://github.com/CaffeineMC/sodium-fabric/commit/298fe1662b288c4b418dee856c47ecd647cd113c.

PR #885, opened 2021-09-04
Create ja_jp.json

CKenJa:
Added Japanese language files.

The translation of vignette, "shadow at the edge of the field of view(視界端の影)" is not accurate, but it is easy to understand and is the same in Optifine, so I translated it as such.
Similarly, can I change unfamiliar words such as fog occlusion and culling to words that are easier to understand?

I'm not very good at writing English, so I may be saying something strange.


amnotbananaama:
Thank you for the PR! **Unfortunately, we will not be accepting translation PR's directly, as we cannot verify them ourselves.** That being said, I do appreciate your effort!

Translations will be pulled automatically from [OneSky](https://jellysquid.oneskyapp.com), which is a free service for crowdsourced translations that anyone can contribute to. 

Right now, the translations on OneSky are more than a year old. They will be updated in the next few days to use the latest `en_us` strings. You can use [this](https://jellysquid.oneskyapp.com/collaboration/translate/project/project/366422/language/4) link for a direct link to the Japanese translation.

You don't have to—but I highly recommend—that you join [the official discord](http://caffeinemc.net/discord). This will give you access to important announcements about Sodium and translations, and allows you to give feedback/coordinate with other translators.

Thank you!

CKenJa:
I understand. I should have contacted on discord beforehand.

Issue #859, opened 2021-08-27
Cull powder snow faces in more cases


muzikbike:
Currently powder snow uses cullface to remove both internal and external faces as to form a continuous mass with adjacent powder snow blocks. This hardcoded behaviour, while fully functional for this purpose, does not account for culling of external faces by non-powder snow blocks, resulting in these faces always being rendered if adjacent to any non-powder snow block even if otherwise completely impossible to see normally.

I suggest that sodium also culls the external faces (although not the internal faces) as if powder snow was any other block, in situations where other blocks with cullface would have the faces culled. (Basically do the same as water and lava, with the distinction that powder snow shouldn't cull its internal faces against blocks, even though water and lava do.)

![2021-08-28_00 42 18](https://user-images.githubusercontent.com/52297970/131198692-68573477-ca21-40d1-898c-8a6a59efd30d.png)
![2021-08-28_00 42 37](https://user-images.githubusercontent.com/52297970/131198695-06953cf2-6a2b-463c-93b0-b08a36350588.png)
![2021-08-28_00 42 51](https://user-images.githubusercontent.com/52297970/131198696-98b219e6-11b1-4257-9c2a-5544773075e3.png)

muzikbike:
Seems fixed now, although the culling is a bit excessive - I'm not sure if that's fixable (dedicated ticket upcoming).

muzikbike:
Reported in #1163.

Issue #866, opened 2021-08-29
Ice z-fighting


ghost:
### Version information

sodium-fabric-mc1.17.1-0.3.1+build.6

### Expected Behavior

Ice should not have weird z-fighting while swimming underwater.

### Actual Behavior

Ice has weird z-fighting. See video below:

https://user-images.githubusercontent.com/75946024/131260403-efb0323b-41be-4f3b-8010-6175e01b10ac.mp4

### Reproduction Steps

1. Find frozen_ocean biome.
2. Swim underwater and look at the icebergs. 
3. Observe weird z-fighting at ice far away.

### Java version

16.0.1.9.1

### CPU

8x Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz

### GPU

NVIDIA GeForce GTX 1080/PCIe/SSE2 GL version 3.2.0 NVIDIA 471.41, NVIDIA Corporation

### Additional information

I tested this with nothing but fabric api and sodium 0.3.1. Issue still observed.

amnotbananaama:
I *believe* this is a duplicate of #38. Someone correct me if I am wrong please

Altirix:
> 
> 
> I _believe_ this is a duplicate of #38. Someone correct me if I am wrong please

nah, just had a look, its Z fighting seems to affect ice and tinted glass, also noticed it wont render water or other transparent blocks that are placed next to the tinted glass. if you place normal glass or coloured glass you dont have the running water rendered of the face

https://github.com/CaffeineMC/sodium-fabric/blob/68ca12bcbd57eddfaf47aa66492464540c2c7581/src/main/java/me/jellysquid/mods/sodium/client/render/pipeline/FluidRenderer.java#L329

but seems water not rendering thru the glass/ice types is something else 
![image](https://user-images.githubusercontent.com/19439141/132939614-ce099956-b057-4d7b-b73f-3aa2d4ef00b4.png)

Issue #89, opened 2020-07-15
Cull the top face of water in waterlogged blocks where neccessary


muzikbike:
For example in the cases of waterlogged top slabs and stairs, this face can never be seen so there's no point in wasting resources rendering it.

![2020-07-15_22 53 44](https://user-images.githubusercontent.com/52297970/87608483-f5b93980-c6f7-11ea-8a29-625ad2109786.png)
![2020-07-15_22 53 47](https://user-images.githubusercontent.com/52297970/87608486-f6ea6680-c6f7-11ea-985b-5d5e83f8c397.png)

muzikbike:
Appears to be fixed:
![2022-02-03_11 17 47](https://user-images.githubusercontent.com/52297970/152333177-a809fa41-6cef-4ab0-8f13-e1591f7e9327.png)

PR #892, opened 2021-09-10
Enhancements for block face & liquid occlusion & Fixes for water Z-fighting with transparent blocks

### LINKED ISSUE #858: Allow top face culling for potted cacti, bamboo and azaleas
Vanilla has this disabled as a kludge fix for MC-195461. Sodium could probably do a better job, re-enabling cullface for these blocks and fixing the original bug properly in the first place.

https://bugs.mojang.com/browse/MC-208189
https://bugs.mojang.com/browse/MC-226540

Altirix:
**Before**

**After**

fixes #858 #859

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/occlusion/BlockOcclusionCache.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/occlusion/BlockOcclusionCache.java
@@ -44,13 +44,13 @@ public class BlockOcclusionCache {
             if (selfShape == VoxelShapes.fullCube() && adjShape == VoxelShapes.fullCube()) {
                 return false;
             }
-            
+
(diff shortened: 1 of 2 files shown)


Altirix:
After with addition of #89 






After fixes for #866


just noticed i squashed over the commit where i swapped out isfluidexposed to isfluidopaque, can revert back anyway if that is prefered as im not too sure myself anymore as im guessing its probs stylistic or mod compat to go with exposed and java is probs optimising out the invert on the return and inverts on the test anyway and if it doesnt the benefit is probs undetectable anyway.

jellysquid3 (approved):

jellysquid3:
Great work, and apologies for the late response here. Changes look OK to me and didn't show issues in testing.

PR #910, opened 2021-09-17
Clean up simple compiler warnings: Unused imports, untyped generic.

ccrvincent:
Was just poking around and noticed these warnings.  There were a few "resource leak" warnings I didn't touch.


jellysquid3 (approved):
Fair enough. :slightly_smiling_face:

PR #921, opened 2021-09-27
Updated README.md

anmolsekhon590:
Fixed grammatical errors and added enhancements in the README Markdown file


silversquirl:
None of these changes seem to actually fix any issues. The grammar is fine as-is

anmolsekhon590:
The use of 'which' is incorrect in line 63 because there is a defining clause in the sentence. In that case, the usage of 'that' is preferred. Other changes can be classified as enhancements.

jellysquid3:
It's small, but the changes here are an improvement I think.

PR #940, opened 2021-10-16
Fix building and running with Java 17

IMS212:
Alongside updating to the latest Gradle RC, this upgrades the visibility of the nextLong method to match 17's random and thus not error.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-rc-1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
 zipStoreBase=GRADLE_USER_HOME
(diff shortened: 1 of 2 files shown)


ImperatorStorm:
wouldn't it be best to wait until gradle 7.3 is stable?

IMS212:
(I've confirmed this doesn't break Java 16 support, at least in development. It shouldn't break as we're upgrading the visibility, not downgrading it.)

IMS212:
> wouldn't it be best to wait until gradle 7.3 is stable?

I haven't experienced any issues with 7.3 so far, and in my opinion Java 17 is quite a nice update to have.

jellysquid3:
I'd rather wait for a stable release of Gradle since J17 compatibility is not too pressing.

joshcangit:
Indeed, this would work with gradle **7.2** as well.
I didn't know it was a Java 17 incompatibility.
The `build.gradle` is unchanged so it should also work with Java 16.

Anyway, the actual fix seems to be mainly the changes in XoRoShiRoRandom.java which solves the build failure.

IMS212:
Gradle 7.3 has just released, this should be ready to merge.

jellysquid3 (approved):

PR #948, opened 2021-10-21
better support for resource packs and Block face culling

Altirix:
### Proposed Changes
Switches Block face culling from true or false to, Disabled, Min, Max.

Min is the only new addition, it disabled block face culling for `CUTTOUT_MIPPED` which is used by leaves, better leaves & smart leaves are fixed by this without needing to disable block face culling entirely. this still is not perfect as this disables the effect on grass blocks too

leaves mixin priority changed to 100 to let cull leaves take priority when installed

warning: this will force sodium to reset its config as the value for blockfaceculling will be invalid


Altirix:
Fixes #732 #817 If players use Blockface Culling: Minimum

MeeniMc:
This PR does fix the linked issues.

Here are some performance figures I collected, this is a jungle biome w/ 32 render distance. Presenting stabilized FPS looking at the same scene at the same angle after loading the full mesh (that is, player doing a 360 before returning to measurement orientation).

CL means Cull-Leaves addon; CL resource-pack only is with CL 'OFF' and the internal resource pack loaded; CL+bushy leaves is CL 'ON' and both CL RP and Bushy Leaves from Vanilla Tweaks loaded. 

| Mode | disabled | min | max | remark
| --- | --- | --- | --- | --- |
| Baseline Sodium | 128 | 136 | 150 | leaves have no back face anyway |
| Sodium + CL resource-pack only | 91 | 95 | 134 | leaves have back faces in min and disabled mode |
| Sodium + Bushy leaves resource-pack | 25 | 39 | 55 | fps unstable (mesh too large) |
Sodium + CL +Bushy leaves resource-pack | 120 | 122 | 157 | only the '3D' part of the leaves show on backfaces in min or disabled |

PR #955, opened 2021-10-28
Fully translated into Russian

ghost:


amnotbananaama:
Thank you for the PR! **Unfortunately, we will not be accepting translation PR's directly, as we cannot verify them ourselves.** That being said, I do appreciate your effort!

The plan is to use a crowdsourcing service which will allow users—such as yourself—to submit translations for Sodium.

You don't have to—but I highly recommend—that you join [the official discord](http://caffeinemc.net/discord). This will give you access to important announcements about Sodium and translations, and allows you to give feedback/coordinate with other translators.

Thank you!

PR #961, opened 2021-11-07
Move mixin config to CaffeineConfig

### LINKED ISSUE #915: Sodium does not apply config overrides if config file is not present
## Expected Behavior

Sodium should apply config overrides from mods' `fabric.mod.json` after creating the default blank config on the first ever load.

## Actual Behavior

When loading for the first time, Sodium will not apply config overrides from mods' `fabric.mod.json`. This can result in crashes if a mod disables certain Sodium mixins due to mixin conflicts.

I believe this code needs a `config.applyOverrides()`.

https://github.com/CaffeineMC/sodium-fabric/blob/68ca12bcbd57eddfaf47aa66492464540c2c7581/src/main/java/me/jellysquid/mods/sodium/common/config/SodiumConfig.java#L182-L191

## Reproduction

This crash can be reproduced using the following "minimal" real world mod setup:
 - Fabric API 0.40.1
 - Sodium 0.3.2
 - Cloth Config 5.0.38
 - Colormatic 2.2.10 (this mod disables some mixins due to a redirect conflict)

which results in the following crash: 
[initial_crash.log](https://github.com/CaffeineMC/sodium-fabric/files/7205643/initial_crash.log)

It is likely that any mod that disables Sodium mixins in order to prevent mixin conflicts is affected by this issue.

A workaround is to simply load the game again, so that Sodium reads a config file.

altrisi:
See https://github.com/CaffeineMC/caffeine-meta/issues/2#issuecomment-962614248

Almost sure this also fixes #915.

Not ready for merging.


douira:
This issue can probably be closed now as the referred issue has been closed.

douira:
Even though this issue was closed because of target branch removal, it should've been closed anyways because the issue it was solving was solved long ago. Just fyi

PR #964, opened 2021-11-11
Update to 1.18

IMS212:
No known issues,


src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
```diff
@@ -321,12 +313,7 @@ public Biome getBiomeForNoiseGen(int x, int y, int z) {
      * Gets or computes the biome at the given global coordinates.
      */
     public Biome getBiome(int x, int y, int z) {
-        int relX = x - this.baseX;
-        int relY = y - this.baseY;
-        int relZ = z - this.baseZ;
-
-        return this.biomeCaches[getLocalSectionIndex(relX >> 4, relY >> 4, relZ >> 4)]
-                .getBiome(this, x, relY >> 4, z);
+        return this.world.getBiome(new BlockPos(x, y, z));
```
PepperCode1: Assuming `WorldSlice` instances are used on one thread at once, this should probably be replaced with a `BlockPos.Mutable` field being `.set` to the new position.

CalXee:
In my experience, loading chunks causes dark chunks quite often, and it's not just the spawn chunks (if you are referring to the thing I think you are referring to) (unless it's caused by Iris)
And PLEASE don't forget to add Chunk Builder (Initially known as Priority Updates) and Simulation Distance into the sodium GUI.

jellysquid3:
>  And PLEASE don't forget to add Chunk Builder (Initially known as Priority Updates)

Sodium completely replaces the chunk building code. We can't just use Minecraft's option here, it's implementation-specific and doesn't apply to Sodium in any way.

That said, in the experiments branch for Sodium 0.4 we have an "Always Defer Chunk Updates" option which gives something similar to what Mojang ended up adding later on in the MC 1.18 development cycle.

IMS212:
I'll remove it, then. In the meantime, I've been looking into what's causing the black sections and I have no idea. All that I can tell is that 1.18 added a light update queue.

src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
IMS212: What is better about this?

IMS212:
The dark chunks seem to have been fully fixed as of newest commit.

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
```diff
@@ -38,6 +39,15 @@ public static OptionPage general() {
                         .setBinding((options, value) -> options.viewDistance = value, options -> options.viewDistance)
                         .setImpact(OptionImpact.HIGH)
                         .setFlags(OptionFlag.REQUIRES_RENDERER_RELOAD)
+                        .setFlags(OptionFlag.REQUIRES_RENDERER_RELOAD)
```
MESLewis: Duplicate line

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
```diff
@@ -206,6 +216,13 @@ public static OptionPage quality() {
                         .setBinding((opts, value) -> opts.quality.enableVignette = value, opts -> opts.quality.enableVignette)
                         .setImpact(OptionImpact.LOW)
                         .build())
+                .add(OptionImpl.createBuilder(boolean.class, vanillaOpts)
+                        .setName(new TranslatableText("options.autosaveIndicator"))
+                        .setTooltip(new TranslatableText("sodium.options.autosave_indicator.tooltip"))
+                        .setControl(TickBoxControl::new)
+                        .setBinding((opts, value) -> opts.showAutosaveIndicator /* Autosave indicator*/ = value, opts -> opts.showAutosaveIndicator)
```
MESLewis: Inline comment seems redundant

src/main/java/me/jellysquid/mods/sodium/mixin/features/chunk_rendering/MixinWorldRenderer.java
```diff
@@ -11,10 +11,7 @@
 import net.minecraft.client.world.ClientWorld;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Matrix4f;
-import org.spongepowered.asm.mixin.Final;
-import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Overwrite;
-import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.*;
```
MESLewis: I think star imports are discouraged

src/main/resources/assets/sodium/lang/en_us.json
```diff
@@ -28,6 +29,7 @@
   "sodium.options.entity_shadows.tooltip": "If enabled, basic shadows will be rendered beneath mobs and other entities.",
   "sodium.options.vignette.name": "Vignette",
   "sodium.options.vignette.tooltip": "If enabled, a vignette effect will be rendered on the player's view. This is very unlikely to make a difference to frame rates unless you are fill-rate limited.",
+  "sodium.options.autosave_indicator.tooltip": "If enabled, shows an \"Saving world\" indicator on the bottom right of your screen when auto-saving.",
```
MESLewis: I think 'a' instead of 'an' is better

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
IMS212: That comment was added back when the autosave indicator field was unmapped, so yes, it is now redundant.

PR #970, opened 2021-11-15
Add a "Vanilla" tab to the settings panel.

Icedude907:
This pull request adds a 'Vanilla' tab to sodium's settings panel, opening the vanilla game's menu screen.

While `Shift+P` opens the vanilla settings screen normally, this isn't immediately apparent to new users and seems like a hacky solution. It's convenient though so I've kept it.

You could argue that the button is better placed in the 'Advanced' panel, but I decided against that to keep the option easily visible for new users.

In addition, mods like Iris mixin their own tabs with custom actions, and support for this in the main project should make this process easier & less likely to cause mixin conflicts.

My implementation probably leaves much to be desired, so advice or assistance is appreciated.

VanillaButton


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
```diff
@@ -17,24 +32,15 @@
...
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.EnumSet;
-import java.util.HashSet;
-import java.util.List;
-import java.util.stream.Stream;
 
 public class SodiumOptionsGUI extends Screen {
-    private final List<OptionPage> pages = new ArrayList<>();
+    private final List<OptionTab> tabs = new ArrayList<>();
```
Madis0, line 37: Why did you refactor OptionPage to OptionTab?

Madis0 (commented):
If you added a button to Sodium settings, perhaps it makes sense to add one to vanilla settings as well, to get back?

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
Icedude907: I was conceptually thinking that a "page" is the settings listed in a "tab", kind of how a page in a web browser is encapsulated within the tab.
I also needed to differentiate between an entry which held settings or acted as a button (which triggers the vanilla settings panel), and going with the word page would mean the subclasses would be named `OptionPageButton` and `OptionPagePage`.

Icedude907:
> If you added a button to Sodium settings, perhaps it makes sense to add one to vanilla settings as well, to get back?

Due to how the game handles submenus, hitting the escape key or `Done` in the vanilla video settings will send you back to the sodium settings if you came from there.

douira:

PR #971, opened 2021-11-16
Detach shaders from program after linking

MESLewis:
According to https://www.khronos.org/opengl/wiki/Shader_Compilation#Cleanup shaders no longer need to be attached to a program after linking. By detaching the shaders they will get cleaned up immediately when deleted, instead of waiting for their attached program to be deleted as well.


duplexsystem:
What all have you tested this on? A lot drivers dont follow the khronos spec well...

MESLewis:
I have an AMD Radeon RX 5700 XT with driver version 21.10.2
I haven't seen any information that this is a pain point for implementations.

jellysquid3 (approved):
Talked about this on Discord some, changes look good.

PR #975, opened 2021-11-18
[WIP] Instanced Model Rendering

burgerindividual:
This PR merges the functionality of the previously external project [Baked Entity Models](https://github.com/KilnGraphics/baked-minecraft-models) into Sodium. This is still an **extremely rough** draft, and it will require lots of work before I would consider it shippable with Sodium. Some TODOs will be listed at the end, some of which I feel are required before merging.

When the new option "Use Model Instancing" is enabled, supported models will be packed into VBOs and instanced. This removes unnecessary uploads to the GPU by only sending transformations for existing vertices. This also removes many unnecessary draw calls by packing many similar entities into batches. This can drastically improve framerate when many entities are visible.

In my tests, the time entities took to render was cut by a factor of about 2-3x, though I will do more scientific tests later.

**Requires OpenGL 4.4 or `ARB_shader_storage_buffer_object`, `ARB_shading_language_packing`, and` ARB_buffer_storage`.**
(Later on, this will probably be able to be limited to only `ARB_shading_language_packing`. More info in TODOs)

**TODO:**
? = Potentially not needed or not sure
- [ ] Fix commit history (requires force pushing or re-creating branch)
- [ ] Add to ThinGL to avoid using raw OpenGL for functions that don't already have a ThinGL replacement
- [ ] Use ThinGL shaders, vertex buffers, index buffers, pipelines, etc.
- [ ] Add support for buffer textures as a fallback for SSBOs so the extension isn't required
- [ ] Don't hardcode the usage of buffer storage
- [ ] Use shader injection to create shaders at runtime rather than using an existing set of shaders (ANTLR)
- [ ] Use dynamically created vertex formats rather than static ones
- [ ] Fix non-instanced types rendering in the wrong order (potentially unfixable if separate vanilla and sodium phases required)
- [ ] ? Have a fallback batching method if instancing doesn't work
- [ ] Instance models that only have differing textures (ex. players and skins)
- [ ] Batch entity fire
- [ ] Batch entity shadows
- [ ] Use less static garbage
- [ ] Integrate instancing into SodiumWorldRenderer and/or more closely mesh world and entity rendering code
- [ ] ? Extend and replace some vanilla objects rather than mixing in (Ex. EntityRendererDispatcher)
- [ ] Have a fallback path for when the persistent buffers exceed their limit rather than native crashing
- [ ] Fix some block entities that have incorrect textures
- [ ] Fix some entities from not being rendered at the correct times (ex. player model in inventory, player heads)


jellysquid3:
Closing as this is unfinished, has major technical hurdles, and is presently unmaintained.

PR #980, opened 2021-11-20
fix: add missing "Close" translation string

DragonEggBedrockBreaking:


Madis0:
It was removed in https://github.com/CaffeineMC/sodium-fabric/pull/764/ as it uses the vanilla `gui.done` now.

jellysquid3:
Latest revision no longer refers to this string.

PR #985, opened 2021-11-21
Update MixinDebugHud.java

crosby-moe:
Stops game from crashing if non-vital redirect fails

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/mixin/features/debug/MixinDebugHud.java
+++ b/src/main/java/me/jellysquid/mods/sodium/mixin/features/debug/MixinDebugHud.java
@@ -1,6 +1,5 @@
 package me.jellysquid.mods.sodium.mixin.features.debug;
 
-import com.google.common.collect.Lists;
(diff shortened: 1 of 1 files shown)


coderbot16:
In what configuration would this fail? It would be appreciated to have a link to a relevant issue report or if you said what mod this causes issues with.

Making this redirect optional with no fallback is undesirable because it means that the F3 screen will not contain sodium debug info.

crosby-moe:
wouldn't it be better to inject on return then? would make it able to have multiple mixins at once on the same method

im making a mod that also redirects that particular method but I have a fallback for where both my mod and sodium are present

embeddedt:
I'm also a bit confused as to why this is a redirect and not an inject that allocates a new list? Stacking multiple mixins together is exactly the kind of use case an inject is intended for.

However, for this to work, I believe both mods would need to use injects and not redirects.

crosby-moe:
Agreed
Ill make a new commit changing the redirect into an inject on return later today when I have the time

PR #993, opened 2021-11-25
Minor locale changes

Madis0:
- Removes now-unused "Particle Quality" string
- Shortens "Leaves Quality" to "Leaves"
- Matches "VSync" writing in FPS limit tooltip with the setting name


Madis0:
#1668

PR #1000, opened 2021-11-28
Use Vector API to optimize some parts of the game.

Vonr:
---
name: Use Vector API to optimize some parts of the game.
about: Project Panama's Vector API is currently in its 3rd incubator. While this is still unstable, I thought it was approaching the point that it was worth spending some effort to lay down some groundwork for future optimizations.
This Pull Request should not be merged into the main branches until the Vector API is out of incubation as the API can still change drastically while in incubation and currently requires the end user to add JVM flags to their game.
title: ''
labels: enhancement
assignees: ''

---

### Proposed Changes

This Pull Request will aim to optimize certain parts of the game to use Project Panama's Vector API.

Project Panama's Vector API is an API that gives developers more direct access to SIMD instructions through the use of vectors. This can drastically improve performance when used correctly.

That said, this Pull Request is currently unlikely to use the Vector API to its fullest potential and will require the contributions of the community. There are currently also no benchmarks to prove that it is faster so it's a bit of a risk merging this.

Hopefully, even if this Pull Request ends up not being accepted, somebody else can contribute a better implementation of the general idea.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/model/light/smooth/SmoothLightPipeline.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/model/light/smooth/SmoothLightPipeline.java
@@ -1,5 +1,6 @@
 package me.jellysquid.mods.sodium.client.model.light.smooth;
 
+import jdk.incubator.vector.FloatVector;
 import me.jellysquid.mods.sodium.client.model.light.LightPipeline;
 import me.jellysquid.mods.sodium.client.model.light.data.LightDataAccess;
 import me.jellysquid.mods.sodium.client.model.light.data.QuadLightData;
@@ -88,9 +89,16 @@ public class SmoothLightPipeline implements LightPipeline {
         float brightness = this.lightCache.getWorld().getBrightness(face, shade);
         float[] br = out.br;
 
-        for (int i = 0; i < br.length; i++) {
-            br[i] *= brightness;
-        }
+//        for (int i = 0; i < br.length; i++) {
+//            br[i] *= brightness;
+//        }
+
+        // What's the use of this? br is never accessed after multiplying it by brightness.
+        br = FloatVector.fromArray(
+                FloatVector.SPECIES_128,
+                br,
+                0
+        ).mul(brightness).toArray();
     }
 
     private void applyComplex(AoNeighborInfo neighborInfo, ModelQuadView quad, BlockPos pos, Direction dir, QuadLightData out, int flags) {
--- a/src/main/java/me/jellysquid/mods/sodium/client/util/color/ColorABGR.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/util/color/ColorABGR.java
@@ -1,5 +1,9 @@
 package me.jellysquid.mods.sodium.client.util.color;
 
+import jdk.incubator.vector.FloatVector;
+import jdk.incubator.vector.IntVector;
+import jdk.incubator.vector.VectorOperators;
+
 /**
  * Provides some utilities for packing and unpacking color components from packed integer colors in ABGR format, which
  * is used by OpenGL for color vectors.
@@ -17,14 +21,32 @@ public class ColorABGR implements ColorU8 {
      * @param a The alpha component of the color
      */
     public static int pack(int r, int g, int b, int a) {
-        return (a & 0xFF) << 24 | (b & 0xFF) << 16 | (g & 0xFF) << 8 | (r & 0xFF);
+        return IntVector.fromArray(
+                IntVector.SPECIES_128,
+                new int[]{a, b, g, r},
+                0
+        ).lanewise(VectorOperators.AND, 0xFF)
+                .lanewise(
+                        VectorOperators.LSHL,
+                        IntVector.fromArray(
+                                IntVector.SPECIES_128,
+                                new int[]{24, 16, 8, 0},
+                                0
+                        )
+                ).reduceLanes(VectorOperators.OR);
+//        return (a & 0xFF) << 24 | (b & 0xFF) << 16 | (g & 0xFF) << 8 | (r & 0xFF);
     }
 
     /**
      * @see ColorABGR#pack(int, int, int, int)
      */
     public static int pack(float r, float g, float b, float a) {
-        return pack((int) (r * COMPONENT_RANGE), (int) (g * COMPONENT_RANGE), (int) (b * COMPONENT_RANGE), (int) (a * COMPONENT_RANGE));
+        float[] c = FloatVector.fromArray(
+            FloatVector.SPECIES_128,
+            new float[]{r, g, b, a},
+            0
+        ).mul(COMPONENT_RANGE).toArray();
+        return pack((int) c[0], (int) c[1], (int) c[2], (int) c[3]);
     }
 
     /**
--- a/src/main/java/me/jellysquid/mods/sodium/client/util/color/ColorABGR.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/util/color/ColorABGR.java
@@ -41,12 +41,12 @@ public class ColorABGR implements ColorU8 {
      * @see ColorABGR#pack(int, int, int, int)
      */
     public static int pack(float r, float g, float b, float a) {
-        float[] c = FloatVector.fromArray(
+        IntVector v = FloatVector.fromArray(
             FloatVector.SPECIES_128,
             new float[]{r, g, b, a},
             0
-        ).mul(COMPONENT_RANGE).toArray();
-        return pack((int) c[0], (int) c[1], (int) c[2], (int) c[3]);
+        ).mul(COMPONENT_RANGE).convert(VectorOperators.F2I, 0).reinterpretAsInts();
+        return pack(v.lane(0), v.lane(1), v.lane(2), v.lane(3));
     }
 
     /**
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
@@ -99,8 +99,8 @@ public class BiomeColorCache {
             }
         }
 
-        int[] ca = c.div(area).toArray();
-        return ColorARGB.pack(ca[0], ca[1], ca[2], 255);
+        c = c.div(area);
+        return ColorARGB.pack(c.lane(0), c.lane(1), c.lane(2), 255);
     }
 
     private int getColor(int x, int z) {
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
@@ -1,5 +1,7 @@
 package me.jellysquid.mods.sodium.client.world.biome;
 
+import jdk.incubator.vector.IntVector;
+import jdk.incubator.vector.VectorSpecies;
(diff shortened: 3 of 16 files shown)


jellysquid3:
Have you verified that these code paths weren't already optimized by the compiler to take advantage of vectorization in the first place? Many of the places you've patched (namely the lighting code and color blending functions) were written to take advantage of the optimizing compiler and shouldn't need manual vectorization here. Last time I checked the generated machine code this seemed to be true, but that could have changed.

I think before I'm willing to accept any PR like this, some things need to happen first:
- We need benchmarks and evidence that these methods are hot enough to justify the increased complexity of code. I'm only aware of a few locations (such as animated texture interpolation and biome blending) where this matters all that much.
- Java needs to support these features without compiler/runtime options, and whichever version of Minecraft we're targeting needs to support it as well. You already noted this, but it should be mentioned again.
- Allocations need to be avoided at all costs in some of these code paths. I'm not sure how the Vector API works in regards to this, but the array constructors seem to suggest that memory is being allocated unless the JVM can optimize it away somehow.

jellysquid3:
I didn't mean to close this. Just clicked the wrong button.

Vonr:
> We need benchmarks and evidence that these methods are hot enough to justify the increased complexity of code. I'm only aware of a few locations (such as animated texture interpolation and biome blending) where this matters all that much.

Unfortunately, I do not have any evidence that the methods are hot enough but my more recent optimizations target bits of code that get called many times such as net.minecraft.util.math.Quaternion

> Java needs to support these features without compiler/runtime options, and whichever version of Minecraft we're targeting needs to support it as well. You already noted this, but it should be mentioned again.

Fully agree on this, This PR should stay open (unless it gets rejected) until the Vector API is supported fully.

> Allocations need to be avoided at all costs in some of these code paths. I'm not sure how the Vector API works in regards to this, but the array constructors seem to suggest that memory is being allocated unless the JVM can optimize it away somehow.

As far as I am aware, the Vector API provides a way to nearly guarantee that the JVM will vectorize the code that you write using it. What I think this means is that the Vector API should only slow down the game if your CPU is massively outdated and does not have SSE capabilities which is almost impossible.
I'm not too sure if memory is being allocated by using it, but from my naive observations it seems that overall memory usage while playing the game has indeed gone up by about 100MB so that is a very real worry. That said, I'm not too sure whether there is a way to set up a vector without allocating memory, and the array constructors are mostly replacing variables so it shouldn't take up that much more memory. All in all, I don't know enough about this topic to really fix it right now, but I'll try to contact the Panama team in my free time.

src/main/java/me/jellysquid/mods/sodium/client/model/light/smooth/SmoothLightPipeline.java
```diff
@@ -88,9 +89,16 @@ private void applySidedBrightness(QuadLightData out, Direction face, boolean sha
...
         float[] br = out.br;
 
-        for (int i = 0; i < br.length; i++) {
-            br[i] *= brightness;
-        }
+//        for (int i = 0; i < br.length; i++) {
+//            br[i] *= brightness;
+//        }
+
+        // What's the use of this? br is never accessed after multiplying it by brightness.
```
magneticflux-, line 96: Originally, `br` is modified in-place as `out.br`, which I assume is used later from something accessing `QuadLightData` (i.e. the renderers themselves).
Your code should instead assign `out.br` to preserve this behavior.

Vonr: out.br is final though

src/main/java/me/jellysquid/mods/sodium/client/util/color/ColorABGR.java
```diff
@@ -17,14 +21,32 @@
...
      * @see ColorABGR#pack(int, int, int, int)
      */
     public static int pack(float r, float g, float b, float a) {
-        return pack((int) (r * COMPONENT_RANGE), (int) (g * COMPONENT_RANGE), (int) (b * COMPONENT_RANGE), (int) (a * COMPONENT_RANGE));
+        IntVector v = FloatVector.fromArray(
+            FloatVector.SPECIES_128,
+            new float[]{r, g, b, a},
+            0
+        ).mul(COMPONENT_RANGE).convert(VectorOperators.F2I, 0).reinterpretAsInts();
+        return pack(v.lane(0), v.lane(1), v.lane(2), v.lane(3));
```
dubiousconst282, line 49: Assuming lane() compiles down to pextrd (3 cycles, 1 throughput), this will probably be slower than the scalar version. A better way to do this (and pack()) is with a byte shuffle (1 cycle + memory load), so you can reorder the lower byte of each int into the one lane.

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -86,13 +89,18 @@ private int calculateBlendedColor(int posX, int posZ) {
...
-                r += ColorARGB.unpackRed(color);
-                g += ColorARGB.unpackGreen(color);
-                b += ColorARGB.unpackBlue(color);
+                c = c.add(
+                        IntVector.fromArray(
+                                IntVector.SPECIES_128,
+                                new int[]{ColorARGB.unpackRed(color), ColorARGB.unpackGreen(color), ColorARGB.unpackBlue(color), 0},
+                                0
+                        )
+                );
```
dubiousconst282, line 98: Same as above, inserting/extracting individual scalars into a vector is inefficient, 1 broadcast + shuffle would probably suffice here.

dubiousconst282 (commented):
_Disclaimer: I'm not familiar with vectors in Java or it's JIT, and this may not apply to architectures other than x86/x64._

In general, inserting and extracting vector lanes is expansive, so you won't gain much speed if you can't load scalars into a vector all at once (i.e. they aren't ordered linearly in memory). If you can't avoid it, `fromArray()/toArray()` will be likely faster than four or eight lane() insertions/extractions.

Other than benchmarking, I'd suggest looking at the assembly generated by the JIT, and the [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html).

FreezePhoenix:
Did some performance testing, this substantially degraded performance.

The vanilla code appears to get automatically vectorized anyways, when analyzing the output of the JIT compiler.

Vonr:
I'll probably redo everything from scratch, basing off 1.18.1/dev but I don't think I can do much due to limited knowledge in this area. I would still like someone else to overtake this effort if at all.

FreezePhoenix:
This initiative is unlikely to be worth it until Project Valhalla (User defined primitives and value-based classes) is completed, or the Vector API gets improved further. As it currently stands, there is no easy way to load values into a vector without allocating a new array (which takes a significant amount of time) or a bunch of single operations. Additionally, if one were to make a class of the following signature:
```java
class Quaternion {
    float[] numbers;
}
```
Then the `numbers` gets allocated each time you construct Quaternion, through a `*newarray` operation. When the class is defined as follows:
```java
class Quaternion {
    float x;
    float y;
    float z;
    float w;
}
```
no such allocation is made. However, Quaternions are still heavy to pass around, and behave as a wrapper. If a Quaternion were to behave as simply an array of floats with getters or setters, then data locality would be preserved, and only one constructor would be called. 

I am aware that there are classes that do this in Kotlin (They are called Value classes, and they act as a wrapper for a single object or value, and the Kotlin compiler tries to inline the object or value whenever possible. However, I have come across a limitation: Java treats the Value class as though it is the class being wrapped. For example:
```kotlin
interface Quaternion

@JvmInline
value class QuaternionValue private constructor(@JvmField val numbers: FloatArray): Quaternion {
    constructor() : this(FloatArray(4))
    constructor(x: Float, y: Float, z: Float, w: Float): this(floatArrayOf(x, y, z, w))
    constructor(x: Int, y: Int, z: Int, w: Int): this(floatArrayOf(x.toFloat(), y.toFloat(), z.toFloat(), w.toFloat()))
    companion object {
        @JvmStatic
        @JvmName("QuaternionValue")
        fun QuaternionValue(x: Int, y: Int, z: Int, w: Int): Quaternion {
            return QuaternionValue(x.toFloat(), y.toFloat(), z.toFloat(), w.toFloat())
        }
    }

    fun that(floatArray: FloatArray) {

    }
    fun that(floatArray: QuaternionValue) {

    }

    val x: Float
        get() = numbers[0];
    val y: Float
        get() = numbers[1]
    val z: Float
        get() = numbers[2]
    val w: Float
        get() = numbers[3]

    fun scale(multiplier: Float) {
        FloatVector.fromArray(FloatVector.SPECIES_128, numbers, 0).mul(multiplier).intoArray(numbers, 0);
    }

    fun copy(): QuaternionValue {
        return QuaternionValue(numbers.clone());
    }
}
```
This class can be constructed easily from Kotlin. However, trying to call one of the constructors (Specifically anything other than the default constructor) from Java fails.

Additionally, while it's obvious that this would have clear benefits if MC were written purely in Kotlin, since the Kotlin compiler handles invocations of constructors and the like, this is not the case. Mixin also doesn't have great interop with Kotlin, and I have not experimented with ASM + Kotlin.

Overall I think that this pull request / issue should be put on a back burner, at least until some improvements come to the API that it relies on.

FreezePhoenix:
> _Disclaimer: I'm not familiar with vectors in Java or it's JIT, and this may not apply to architectures other than x86/x64._
> 
> In general, inserting and extracting vector lanes is expansive, so you won't gain much speed if you can't load scalars into a vector all at once (i.e. they aren't ordered linearly in memory). If you can't avoid it, `fromArray()/toArray()` will be likely faster than four or eight lane() insertions/extractions.
> 
> Other than benchmarking, I'd suggest looking at the assembly generated by the JIT, and the [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html).

It seems to be the opposite - the JVM has a hard time optimizing out any `new float[] { ... }` and instead keeps the allocations. From my experience, loading around 4 scalars into a vector one-by-one is more efficient than `fromArray` (Assuming you don't already have an array).

> Allocations need to be avoided at all costs in some of these code paths. I'm not sure how the Vector API works in regards to this, but the array constructors seem to suggest that memory is being allocated unless the JVM can optimize it away somehow.

I've done some testing, and the JVM appears to be unable to optimize out `new` usages, whether or not for arrays. This appears to be an outstanding problem in the JVM, but it also seems like it's here to stay.

> We need benchmarks and evidence that these methods are hot enough to justify the increased complexity of code. I'm only aware of a few locations (such as animated texture interpolation and biome blending) where this matters all that much.

I've done some benchmarking, using the methods that Vonr has been using to construct Vectors from `new float[] { ... }` proves to be around 4x slower than the vanilla implementation of it. Using `.withLane`, the fastest I can get is still slightly slower than plain scalar instructions.

I have found one exception, which is extremely interesting: `Quaternion.conjugate`.

src/main/java/me/jellysquid/mods/sodium/client/model/light/smooth/SmoothLightPipeline.java
FreezePhoenix: Regarding this: Since `br` was in the original code a reference, assigning to values inside `br` changes values in `out.br`.

Even though an object may be final, it is still mutable (unless it's fields are marked as final).

So, you should instead do: `.mul(brightness).intoArray(br)`

FreezePhoenix:
I have recently found a way to load vectors into the register without allocating memory. However, I find it to only be effective for `float` and `int`, not `double` or `long`. More experimentation is required, however, in order to confirm the latter observation.

PR #1001, opened 2021-11-28
Handle DashLoaders use of Unsafe.allocateInstance()

alphaqu:
DashLoader uses Unsafe.allocateInstance in spots where we cannot use the constructor as we have cached all of the fields to save computation on initialization. Sodium has a couple of conflicts were in the past DashLoader needed to disable a couple of important Sodium features to stay compatible. This pr solves all of the important conflicts.


jellysquid3:
I'm not accepting mod-specific workarounds for problems like these. You shouldn't be initializing objects with `Unsafe.allocateInstance` unless you intend to initialize all the fields manually, and trying to barricade ourselves to prevent bugs caused by our constructors not running isn't an acceptable workaround.

alphaqu:
The issue is that I do initialize all of the fields. The issue comes when sodium is present where there are mixin fields that are not there normally. This requires a check either on my side to see if sodium is present or for sodium to be resilient and be able to handle the field being null.

PR #1050, opened 2021-12-13
Respect server view distance, use new fog calculation

IMS212:
Seems to fix the multiple fog distance issues.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -158,7 +158,7 @@ public class SodiumWorldRenderer {
 
         this.useEntityCulling = SodiumClientMod.options().performance.useEntityCulling;
 
-        if (this.client.options.viewDistance != this.renderDistance) {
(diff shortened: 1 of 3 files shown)


IMS212:
Fog calculation is finally correctly fixed.

dopadream:
good fucking riddance

MeeniMc:
Tested, it works as intended; changing the server viewdistance also updates the fog distance

KaspianDev:
Im not a big fan of this, this makes the fog too close to me on servers with small view distance

IMS212:
Being a fan of this doesn’t matter; the question is if it matches Vanilla. We will not break what vanilla does just so a few people are happier.

dopadream:
It's literally how fog behaves in vanilla.

KaspianDev:
i'd like to have an option to modify it like in old versions of the game

IMS212:
That may be added by a mod like Sodium Extra, but should not be part of the official mod.

jellysquid3 (approved):

jellysquid3:
We don't intend to meaningful deviate from how vanilla renders things. Fog is a gameplay feature and we especially don't want to change that. This is supposed to be an optimization mod, not a kitchen sink of gameplay options.

PR #1066, opened 2021-12-19
Add jitpack.yml so that JitPack can build with the correct java version

Justsnoopy30:
This allows projects like Iris to be able to use JitPack to compile against sodium development branches instead of needing a local jar file.

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,2 @@
+jdk:
+  - openjdk17


jellysquid3 (approved):

IMS212:
This doesn't seem to work, I believe it must be put like this:
before_install:
  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
  - source install-jdk.sh --feature 17

jellysquid3:
I really do not want to let someone's arbitrary Bash script from a web URL, not versioned or verifiable in any way, to run in our CI environment.

jellysquid3:
What is the use case for JitPack here, anyways? Do you just need snapshot builds of a given tag/branch on GitHub? Surely we can create a CI task to upload builds to our Maven...

coderbot16:
IMS: ~~Jitpack build passed just fine on 1.18: https://jitpack.io/com/github/CaffeineMC/sodium-fabric/1.18.x~dev-540d27bf70-1/build.log~~ Never mind???

JellySquid: The use case is mainly for depending on arbitrary in-development GitHub builds, perhaps from forks if needed for testing with Minecraft snapshots.

It's not meant for release versions, but just temporarily here and there during development. Otherwise, we depend on Sodium builds provided by Modrinth.

Issue #105, opened 2020-07-18
Sodium's options fill the entire screen unnecessarily 


DevBefell:
**Is your feature request related to a problem? Please describe.**
It is inconvenient that the gui is on the top left instead of the middle.
**Describe the solution you'd like**
Make the option GUI Centered/Option to change.
**Describe alternatives you've considered**

**Additional context**

jellysquid3:
This is a bug in Sodium's positioning code, and will be addressed.

DevBefell:
So it is not suppose to be on the top left and it is a bug?

douira:
Is this still relevant? This issue is 4 years old

DevBefell:
As far as I know, I never closed as it was kinda a bug that went into a feature, it was a "bug" that was worked on but ultimately didn't make the cut, and I left this as is as if developers want to reference back, it is here.   The current versions of Sodium all have their GUI's positioned in the top right as opposed to the center so this issue is basically resolved. 
But with the discussions of GUI changes/api for third party mods using the UI it could be kept until then,  see how/when this is counted completed/resolved.

douira:
I think this issue can be closed on account of left-alignment being a feature. There will be separate issues for ui redesigns or options apis.

PR #1108, opened 2022-01-07
Center the option GUI

MeeniMc:
MC GUI's are generally centered. Sodium options are departing from that model by being left adjusted.

This also reduces distance between the DONE button and the rest of the options.

This applies only when gui-scale is not set to AUTO, in AUTO it looks the same as before.

Signed-off-by: MeeniMc <68366846+MeeniMc@users.noreply.github.com>

## Some screens to illustrate: 

With patch, gui-scale=4x, 4K screen: center of the screen is occupied, less distance between the options and the 'apply/done' buttons

---

With patch: gui-scale=AUTO, pane 1: the layout has been verified to still prevent overlap between DONE/APPLY and the options in pane 1, and the option tooltips in pane 3 (this is the reason the DONE/APPLY have not been centered with this PR). 

---

<details> 
  <summary>For reference only, how it looks on 4k without this PR</summary>
Before, gui-scale=4x, 4K screen: lots of empty space in the center of the screen.

</details>

## Alternatives

This is an alternative to pr #41, except that it doesn't suffer from overlap of buttons in AUTO mode

## Related issues

Fixes issue #105

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumOptionsGUI.java
@@ -158,9 +158,12 @@ public class SodiumOptionsGUI extends Screen implements ScreenPromptable {
         this.rebuildGUIPages();
         this.rebuildGUIOptions();
 
-        this.undoButton = new FlatButtonWidget(new Dim2i(this.width - 211, this.height - 30, 65, 20), Text.translatable("sodium.options.buttons.undo"), this::undoChanges);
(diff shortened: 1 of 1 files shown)


Pectacular:
Wouldn't it be better if the options menu was centred vertically too and not only horizontally?

MeeniMc:
> Wouldn't it be better if the options menu was centred vertically too and not only horizontally?

Vanilla menus are centered horizontally, but are generally top-aligned (see the Options menu layout with a small GUI size).

MeeniMc:
Updated to remove merge conflicts. Do I need to do something to get this merged, or just put it in the trash bin?

MeeniMc:
Discussed on discord: postpone resolution of the underlying issue to config screen refactor see #2308

PR #1110, opened 2022-01-08
Changes to mapped buffers and streaming buffers

burgerindividual:
* Makes mapped buffers and streaming buffers pointer-based
* New option for explicitly-flushed buffer mapping
* Completely revamp the syncing system in MappedStagingBuffer


burgerindividual:
Closing due to bad code and changes were already pushed to fix the issues.

PR #1149, opened 2022-01-20
Remove stuff about GitHub actions builds from readme, since it's disabled

JustinTimeCuber:


mfn003:
It's enabled

JustinTimeCuber:
Huh, didn't notice, looks like it got re-enabled about a month ago

PR #1236, opened 2022-03-20
fix: Make boxes outside of the world height always visible

### LINKED ISSUE #1081: firework particles (and probably particles in general) are not made above the maximum height of the world
### Version information

Sodium alpha 6 - 1.18.1

### Expected Behavior

We should see the particles

### Actual Behavior

The particles are not rendered

### Reproduction Steps

Video:

https://user-images.githubusercontent.com/68275908/147370011-3a154743-07e2-455e-9494-23564c57e1fa.mp4

### Java version

Java 17

### CPU

Ryzen 1600

### GPU

GTX 1070

### Additional information

_No response_

bytzo:
### Summary

This pull request changes the [`isBoxVisible()`](https://github.com/CaffeineMC/sodium-fabric/blob/d50338a9a6c01614f5f941bebbcd4b47e61939fd/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java#L349) method to first check if the specified box is outside the world height. If it is, the method will return `true`. This fixes #1081.

### Details

Currently, `isBoxVisible()` always returns `false` for boxes above or below the world height, since no visible chunk sections exist there. This can cause issues with Sodium's entity and particle culling, since they both use this value to determine if something should be culled or not. If using this value directly, entities and particles above or below the world height will always be culled, even if the player can see them.

Although a workaround for this issue already exists for entities (see 885222a476b7054ae428031d72c0b47fa579872a), particles still have the issue of being incorrectly culled (see #1081). The workaround for entities checks if an entity is above or below the world height, then avoids culling it if so. While the same kind of logic could be implemented with particle culling, it might make more sense to instead implement the logic within the `isBoxVisible()` method. This way, the result of `isBoxVisible()` is more true to its name and a special condition does not need to be added with its every use.

### Additional notes

This pull request changes the conversion of block to chunk coordinates within `isBoxVisible()` to use `ChunkSectionPos.getSectionCoord()` for readability.

By making these changes, an issue where entities below 0 or above 255 are never culled is fixed.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -326,18 +326,12 @@ public class SodiumWorldRenderer {
             return true;
         }
 
-        Box box = entity.getVisibilityBoundingBox();
-
-        // Entities outside the valid world height will never map to a rendered chunk
-        // Always render these entities or they'll be culled incorrectly!
-        if (box.maxY < 0.5D || box.minY > 255.5D) {
-            return true;
-        }
-
         // Ensure entities with outlines or nametags are always visible
         if (this.client.hasOutline(entity) || entity.shouldRenderName()) {
             return true;
         }
+
+        Box box = entity.getVisibilityBoundingBox();
 
         return this.isBoxVisible(box.minX, box.minY, box.minZ, box.maxX, box.maxY, box.maxZ);
     }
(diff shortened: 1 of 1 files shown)


MeeniMc (approved):

jellysquid3 (approved):

jellysquid3:
Looks good to me. This solution is far more robust than the (very silly) hack that I had placed there before.

jellysquid3:
Thanks for the contribution once again.

PR #1259, opened 2022-04-05
Fix memory leak in ChunkBuilder.

Giraffe1966:
This PR fixes a memory leak that I could reproduce after going through multiple nether portals with a high render distance. This leak appears to be caused by the fact that ChunkBuilder.localContexts#remove() is never called by the main thread. Even though the ThreadLocal object itself is garbage collected when all references to a ChunkBuilder disappear, the main thread maintains a strong reference to the value held by the localContexts ThreadLocal, preventing the value from ever being garbage collected since the main thread never dies. This fixes that issue by ensuring that ChunkBuilder.localContexts#remove() is called by the main thread when the ChunkBuilder is being removed (e.g.  when a player goes through a nether portal).


burgerindividual:
This PR had conflicts with the latest branch due to the package name change, so I ported it in commit 3af94becf3c01e78a263df9fa44843f7cc9aaa9c and cosigned you. Thanks for the PR!

PR #1265, opened 2022-04-09
Add back vanilla shader color bug fix

violet-sh:
This adds back a previously included bug fix which fixes color corruption noticeable in beacon beams and block breaking.


jellysquid3:
Thanks.

PR #1305, opened 2022-05-15
refactored createBuilder to use a factory pattern

jan-leila:


burgerindividual:
I don't really see a reason for this change. In my opinion, the existing version looks a bit better, so I'm going to close this.

PR #1331, opened 2022-06-14
Added enableSodiumColorBlending method to fluid and block classes to get a value indicating if to enable color blending for this object.

### LINKED ISSUE #895: Blockentity with color not rendered correctly
### Version information

sodium-fabric-mc1.17.1-0.3.0+build.5

### Reproduction Steps

1.Install Sodium and another mod that has blockentity with a block color provider
2.The blockentity is rendered in a strange way

### Java version

Java 16

### CPU

Intel Core i5-10400F

### GPU

NVIDIA GeForce GTX 1650 SUPER

### Additional information

_No response_

### LINKED ISSUE #1323: LinearColorBlender breaks any case where a precise position is required in a BlockColorProvider.
### Version information

mc1.19-0.4.2+build.16

### Expected Behavior

I've implemented a new block that requires its precise position to get data to calculate the color.
It needs its precise actual position to get its associated block entity, then get an item stored inside the block entity.

```java
ColorProviderRegistry.BLOCK.register(TunBlock::getContainedFluidColor, TUN);
```

```java
public static int getContainedFluidColor(BlockState state, BlockRenderView view, BlockPos pos, int i) {
    if (view.getBlockEntity(pos) instanceof TunBlockEntity tunEntity) {
        if (tunEntity.hasContained()) {
            return containable.get(tunEntity.getContained()).getColor(state, view, pos, i);
        }
    }

    return -1;
}
```

### Actual Behavior

With sodium installed, the color provider receives the neighbors block positions but not the actual block position.

This is caused by [LinearColorBlender](https://github.com/CaffeineMC/sodium-fabric/blob/e550fad8b38825f822d0039210f557babca15f3c/src/main/java/net/caffeinemc/sodium/render/terrain/color/blender/LinearColorBlender.java#L54) that requires the neighbors block colors to blend them.

The result, in by specific case, is that the color provider cannot get the block entity, and the result obtained is null, so it returns the empty color.

Calling trace

calling

### Reproduction Steps

Create a new block, then register a color provider for the block.
Start a debug session, with a breakpoint inside the method for getting the color.
Check the BlockPos parameter.
Check the actual block pos of the block (the block position, not the player position!) with F3 in the game.

Remember to add sodium into the mods folder into your run folder!

Example: 
```java
ColorProviderRegistry.BLOCK.register(MyBlock::getPartColor, MYBLOCK);
```

```java
public static int getPartColor(BlockState state, BlockRenderView view, BlockPos pos, int i) {
    if (view.getBlockEntity(pos) instanceof MyBlockEntity entity) {  //<<< Place the breakpoint here.
        //Calculating color from the block-entity...
    }

    return -1;
}
```

### Java version

Java 17

### CPU

Intel(R) Core(TM) i7-9750H

### GPU

NVIDIA RTX 2080

### Additional information

_No response_

ghost:
This is useful if someone creates a mod with blocks requiring to get a precise color from the current position, without blending.

An example from my mod.

With blending enabled:  `boolean enableSodiumColorBlending() { return true; }`

With blending disabled:  `boolean enableSodiumColorBlending() { return false; }`

The usage is just to override the method into the `block` or `fluid` class.

```java
@Override
public boolean enableSodiumColorBlending() {
    return false;
}
```

As default returns true, so the color blending is enabled by default.

This adds a way to fix #1323 and fix #895

### Note
This way, are the modders who need to implement compatibility by overriding the method, but requires to be documented in the readme or in a wiki.

### DIFF
--- a/src/main/java/net/caffeinemc/sodium/mixin/features/block/MixinBlock.java
+++ b/src/main/java/net/caffeinemc/sodium/mixin/features/block/MixinBlock.java
@@ -1,0 +1,9 @@
+package net.caffeinemc.sodium.mixin.features.block;
+
+import net.caffeinemc.sodium.render.terrain.color.blender.SodiumColorBlendable;
+import net.minecraft.block.Block;
+import org.spongepowered.asm.mixin.Mixin;
+
+/** Injects the {@link SodiumColorBlendable} interface. */
+@Mixin(Block.class)
+public class MixinBlock implements SodiumColorBlendable {}
--- a/src/main/java/net/caffeinemc/sodium/mixin/features/block/MixinFluid.java
+++ b/src/main/java/net/caffeinemc/sodium/mixin/features/block/MixinFluid.java
@@ -1,0 +1,9 @@
+package net.caffeinemc.sodium.mixin.features.block;
+
+import net.caffeinemc.sodium.render.terrain.color.blender.SodiumColorBlendable;
+import net.minecraft.fluid.Fluid;
+import org.spongepowered.asm.mixin.Mixin;
+
+/** Injects the {@link SodiumColorBlendable} interface. */
+@Mixin(Fluid.class)
+public class MixinFluid implements SodiumColorBlendable {}
--- a/src/main/java/net/caffeinemc/sodium/render/terrain/color/blender/LinearColorBlender.java
+++ b/src/main/java/net/caffeinemc/sodium/render/terrain/color/blender/LinearColorBlender.java
@@ -4,6 +4,7 @@ import net.caffeinemc.sodium.render.terrain.color.ColorSampler;
 import net.caffeinemc.sodium.render.terrain.quad.ModelQuadView;
 import net.caffeinemc.sodium.util.packed.ColorARGB;
 import net.caffeinemc.sodium.util.color.ColorMixer;
+import net.minecraft.state.State;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.world.BlockRenderView;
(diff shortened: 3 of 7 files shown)


Juuxel:
I'd note as someone whose mod suffers from the bug (Juuxel/PaintersBlocks#3) that an interface on Block and overriding a method to return false isn't necessarily the best solution. First, Sodium shouldn't be blending random block colours anyway without actually knowing if they're foliage etc.

Second, modders might accidentally assume the interface is always present in e.g. a super call, which will cause a classloading crash. (Unless you just add the method without `@Override` to your class and avoid `super` but that seems a bit ugly too)

But yeah, it'd be great to finally have some fix for the issue in any case :+1:

LucilleTea:
A fix for the underlying issue was made [here on the 1.17.x/0.4-experiments branch](https://github.com/CaffeineMC/sodium-fabric/commit/634b11ed989482b9aab59ba45acc92ba31b97648), though there's no feedback on whether it worked in [the associated issue](https://github.com/CaffeineMC/sodium-fabric/issues/895). If it did work, it should be a better solution than asking modders to manually opt out of the blending.

ghost:
> A fix for the underlying issue was made [here on the 1.17.x/0.4-experiments branch](https://github.com/CaffeineMC/sodium-fabric/commit/634b11ed989482b9aab59ba45acc92ba31b97648), though there's no feedback on whether it worked in [the associated issue](https://github.com/CaffeineMC/sodium-fabric/issues/895). If it did work, it should be a better solution than asking modders to manually opt out of the blending.

https://github.com/CaffeineMC/sodium-fabric/issues/895#issuecomment-1165041235

ghost:
Closed because of already existing better solution.

PR #1374, opened 2022-07-03
Update actions so that artifacts are uploaded.

piuvas:
Also updates dependencies, is there a reason not to? 

💜


burgerindividual:
Artifacts from actions are disabled purposefully to prevent bug reports from versions that aren't supported and/or are known to be buggy. I will update all of the actions dependencies though, now that you've reminded me.

PR #1375, opened 2022-07-04
Block lighting fixes

### LINKED ISSUE #226: Black sides of a block when using 3D textures
### Expected Behavior
Blocks show animation normally when using 3D textures from certain Resource Packs such as Sphax BDcraft with More 3D Blocks addon.

### Actual Behavior
Certain blocks (furnaces, for example) instead of working animation become black on all sides.

### Reproduction Steps
1. Install Sodium
2. Install Sphax BDCraft and More 3D Blocks addon
3. Place furnace and fill it with coal and ore.

### System Information

- Java Version: Latest.
- CPU: 8Gb is afforded to Minecraft.

PepperCode1:
- Check for luminance as well when optimizing lightmap retrieval; Fix #226
  - Before, only `state.hasEmissiveLighting(world, pos)` was checked to determine if the lightmap should be retrieved for a certain state, but this caused blocks that were not emissive but did emit light (such as a lit furnace or full respawn anchor) to receive the incorrect lightmap value. Checking the state luminance in addition to emissivity fixes this issue.
- Use vanilla fix for underwater lighting
  - `state.getFluidState() != EMPTY_FLUID_STATE` -> `!state.shouldBlockVision(world, pos)`
  - This change makes light calculation match vanilla's `BlockModelRenderer` which may avoid future issues and provide better compatibility. It also serves the same purpose as before, meaning underwater lighting still works correctly.

### DIFF
--- a/src/main/java/net/caffeinemc/sodium/render/terrain/light/data/LightDataAccess.java
+++ b/src/main/java/net/caffeinemc/sodium/render/terrain/light/data/LightDataAccess.java
@@ -62,21 +62,21 @@ public abstract class LightDataAccess {
         BlockState state = world.getBlockState(pos);
 
         float ao;
+        boolean em;
 
         if (state.getLuminance() == 0) {
             ao = state.getAmbientOcclusionLightLevel(world, pos);
+            em = state.hasEmissiveLighting(world, pos);
         } else {
             ao = 1.0f;
+            em = true;
         }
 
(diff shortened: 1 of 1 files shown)


MeeniMc:
Would it be possible to fasttrack this to 0.4.6 rather than /next?

burgerindividual:
Next shouldn't take *too* long to come out at a full release I think.

PR #1384, opened 2022-07-11
Block lighting fixes and optimizations II

PepperCode1:
- Fix flat lighting on inset quads on blocks that do not emit light and are not emissive
- Fix dark quads on light-emitting blocks behind tinted glass
- Restructure SmoothLightPipeline to use the fastest code path possible
- Add some vanilla checks for better parity
- Add IS_PARALLEL quad flag
- Add FC (full cube) to light cache words

### DIFF
--- a/src/main/java/net/caffeinemc/sodium/render/terrain/BlockRenderer.java
+++ b/src/main/java/net/caffeinemc/sodium/render/terrain/BlockRenderer.java
@@ -66,7 +66,7 @@ public class BlockRenderer {
             }
 
             if (!cull || this.occlusionCache.shouldDrawSide(state, world, pos, dir)) {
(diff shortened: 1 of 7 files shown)


Altirix:
image

this can cause slabs and stairs to become emissive if they are placed between a source block and the player.

i have no idea if this has any negative effects but i was not able to find any. this is what i was able to come up with to prevent this from occurring.

https://github.com/CaffeineMC/sodium-fabric/compare/dev...Altirix:sodium-fabric:flat_light_emissives

found a few cases where its different, Lava cauldrons encased in tinted glass will have dark sides, the fence post that is encased in tinted glass and lit from below has dark sides (same as 0.4.10)

Issue #872, opened 2021-08-30
Certain chunks disappearing and reappearing at certain angles, persists in Sodium MC 1.17.1. 


DTM06:
### Version information

mc1.17.1-0.3.2+build.7, mc1.16.3-0.1.0

### Expected Behavior

![vanilla](https://user-images.githubusercontent.com/88454555/131344099-f4265afd-278b-41ac-9c18-50fd875860be.gif)

Taken in Vanilla 1.17.1. As seen in the GIF, chunks persist no matter what angle the player is facing. 

**Vanilla F3 Debug Menu :** 

![vanillaf3](https://user-images.githubusercontent.com/88454555/131344304-d89908b1-b476-4028-926c-90cf1ec59964.png)

### Actual Behavior

![sodium](https://user-images.githubusercontent.com/88454555/131344400-37427f36-4b1b-4817-9662-3f586ad82a58.gif)

Taken in Sodium 1.17.1 (0.3.2+build.7). As seen in the GIF, chunks seem to disappear at certain angles. These are the same two scenes, one in the overworld, one in the nether. I've seen this in many other instances, but only selected these two scenes in order to present this issue, Excuse the low resolution GIF, this was necessary in order to meet GitHub upload size requirements.

**Sodium F3 Debug Menu :** 

![sodiumf3](https://user-images.githubusercontent.com/88454555/131345168-856be450-393c-42b2-990d-91be04edb331.png)

### Reproduction Steps

As I am uncertain of the specific circumstances in which this issue occurs, I have provided a world download of one of the scenes seen in the above GIFs. I've tried playing around in my video settings to no avail, however I have still provided my usual settings for maximum accuracy. 

1. Mirror my exact video settings as seen in the following screenshots (taken in mc1.17.1-0.3.2+build.7) : https://imgur.com/a/DqGyBrl
2. Download the following world file (1.17, but is backwards compatible with 1.16) : 
[Debug.zip](https://github.com/CaffeineMC/sodium-fabric/files/7076950/Debug.zip)
3. Read the lectern at world spawn, then right-click the button in front of you.

If all goes to plan, you should be seeing something similar to what I have described. 

### Java version

1.17.1 : Java 16.0.1 64-bit (Bundled), 1.16.5 : Java 1.8.0_51 64-bit (Bundled)

### CPU

Intel Core i5-10400F

### GPU

Radeon RX 5600 XT, Driver ver. 21.8.2

### Additional information

I am not running any mods apart from Sodium MC.
[...]

Issue #722, opened 2021-06-30
Subchunk culling issue


rjb3977:
### Version Information

`mc1.17-0.3.0+build.458`

### Expected Behavior

The subchunk renders when visible.

### Actual Behavior

The subchunk does not render when viewed at certain angles from certain chunks with blocks in just the right places.

### Reproduction Steps

Load the attached world and fly around between the two chunks as shown in the screenshots. The subchunk containing the gold block will disappear and reappear.

I believe the example in this world is minimal. Removing any individual stone block makes the subchunk render from all angles.

[Broken Occlusion.zip](https://github.com/CaffeineMC/sodium-fabric/files/6743866/Broken.Occlusion.zip)

### System Information

- OS: `Windows 10`
- Java Version: `16.0.1`
- CPU: `Ryzen 3900X`
- GPU: `Radeon RX 5700 XT`

### Other Information
Other mods:
- `sodium-fabric-mc1.17-0.3.0+build.458`
- `lithium-fabric-mc1.17-0.7.2`
- `light-overlay-6.0.0-fabric`
- `fabric-api-0.36.0+1.17`
- `cloth-config-5.0.34-fabric`
- `architectury-2.0.16-fabric`

I originally encountered the issue in a cave. I copied the setup with world edit, reduced it as much as I could, and moved it to the center of the map. I haven't tested rotating or reflecting it.

Screenshots taken from the attached world from the same two positions using both Sodium and Vanilla:

### Sodium
![sodium 1](https://user-images.githubusercontent.com/42881603/124028069-906e5f00-d9c1-11eb-82b2-1a8b984328e7.png)
![sodium 2](https://user-images.githubusercontent.com/42881603/124028074-92382280-d9c1-11eb-8f4d-b9718d8976c1.png)

### Vanilla
![vanilla 1](https://user-images.githubusercontent.com/42881603/124028078-9401e600-d9c1-11eb-88a7-3c5deb9a50ea.png)
![vanilla 2](https://user-images.githubusercontent.com/42881603/124028085-95331300-d9c1-11eb-8a9b-80cc1dd8c7b4.png)

rjb3977:
I realize now that the strange structure visible in the bottom right isn't actually necessary, and can be replaced with a flat wall. What matters is that the wall completely disconnects the section with the gold block from the section below the player in the second and fourth screenshots. Here's the same world with just three flat walls and the gold block.
[...]

PR #1393, opened 2022-07-20
Culling algorithm improvements

Kroppeb:
Probably fixes the following 2 issues (which I think are duplicates?): #722 and #872

### DIFF
--- a/src/main/java/net/caffeinemc/sodium/render/chunk/occlusion/ChunkOcclusion.java
+++ b/src/main/java/net/caffeinemc/sodium/render/chunk/occlusion/ChunkOcclusion.java
@@ -1,150 +1,266 @@
 package net.caffeinemc.sodium.render.chunk.occlusion;
 
-import it.unimi.dsi.fastutil.floats.FloatArrayList;
-import it.unimi.dsi.fastutil.ints.IntArrayFIFOQueue;
 import it.unimi.dsi.fastutil.ints.IntArrayList;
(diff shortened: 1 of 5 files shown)


src/main/java/net/caffeinemc/sodium/render/chunk/occlusion/ChunkOcclusion.java
```diff
@@ -1,136 +1,266 @@
...
+        }
 
-    private static int markDirection(int state, int dir) {
-        return (state | (1 << dir));
-    }
+        last++;
 
-    private static boolean hasDirection(int state, int dir) {
-        return (state & (1 << dir)) != 0;
+        if (first != 0 && last != sections.length) {
```
Leo40Git, line 258:
```suggestion
        if (first != 0 && last != sections.length - 1) {
```
At present, this "trim array" path is always run.

Kroppeb: yes and no, it's a off by one where I initialize.

Kroppeb: actually no, the `last++` fixes that.

Leo40Git: Oh, oops, sorry. This `if` just looks fishy.

Issue #1380, opened 2022-07-07
Please add the resolution slider to the settings


naughtok:
**Is your feature request related to a problem? Please describe.**
Occasionally I get the problem where I'll update my graphics drivers or something and Minecraft will start running at 60hz again instead of my native 165hz. When that happens, tabbing in and out of Minecraft is extremally slow and there is a black screen for 5-10 seconds because it has to switch refresh rates. In order to fix this problem with Sodium installed, I have to remove Sodium and all the mods that depend on it from my mods folder, open Minecraft just to switch it from "1920x1080@60hz" to "1920x1080@165hz," and then add the mods back and relaunch Minecraft. Its kind of a pain, and I would assume to be a simple fix.
**Describe the solution you'd like**
My solution would be to just add the resolution slider to the settings, it would be nice to have.

burgerindividual:
I'll mark this as planned for 0.5.0, but in the meantime, you can do Shift + P to bring up the vanilla menu.

FlashyReese:
There is a PR for it, https://github.com/CaffeineMC/sodium-fabric/pull/957. Although it might be a bit out of date.

naughtok:
> 

Oh I didn’t realize you could bring up the vanilla menu

voltflake:
What's the status of implementing this slider? 0.5.0 was released a long time ago, maybe bump up issue priority?

douira:
This issue and the other issue you commented on can be closed for being part of https://github.com/CaffeineMC/sodium-fabric/issues/46 which includes both of these aspects. https://github.com/CaffeineMC/sodium-fabric/issues/46 is implemented by the open PR https://github.com/CaffeineMC/sodium-fabric/pull/2296.
Edit: #46 has been closed with reasoning given in a comment there.

PR #957, opened 2021-10-30
Implement missing graphics sliders

Madis0:
Implements:

- Fullscreen Resolution slider (ported from Sodium Extra [with permission](https://discord.com/channels/748363294749753484/812215092510195732/903737556481474630) [(via)](https://discord.gg/TBGGM9VzH9); code slightly simplified)
- Distortion Effects slider (vanilla name and tooltip)
- FOV Effects slider (vanilla name and tooltip)
- Increases the width of option groups from 200 to 250 to make Fullscreen Resolution and Max Pre-Rendered Frames phrases fit better

Fixes #46, therefore finally giving Sodium settings parity with all vanilla sliders!

**Note:** Draft PR because
* FOV and Distortion Effects sliders have a slight rounding error compared to vanilla's. I tried using Math.round and Math.floor in several ways with no luck. 
*  "max pre-rendered frames" gets abbrevated on hover and "fullscreen resolution" doesn't... Should make it the opposite now but I haven't found the reason for abbreviation.

Help is appreciated.

PR #1409, opened 2022-08-01
Add missing resolution slider to video settings - Issue #1380

akrossu:
## Issue
- [Please add the resolution slider to the settings #1380](https://github.com/CaffeineMC/sodium-fabric/issues/1380)

## Description
- Adds a simple slider element in Video Settings to adjust the full-screen resolution.

## Known Issues
- ~~Does not dynamically set resolution without having to un-fullscreen and re-fullscreen.~~
- ~~Does not display resolution when changing settings but displays the settings index value such as 0 - 100 possible resolutions.~~

Thank you to this pr by [Madis0](https://github.com/Madis0)(https://github.com/CaffeineMC/sodium-fabric/pull/957) I was able to get a foot in the door with the issue.

I am aware this is planned for 0.5.0, however, I was very bored and wanted to make my first forked pull request


douira:

PR #1419, opened 2022-08-09
optimization: Cache queues + visibleSections

MCRcortex:
20% performance improvement at 32 render distance with always doing graph updates (to benchmark)


MCRcortex:
Burger rewrote the whole thing instead

PR #1422, opened 2022-08-14
Vertex formats refactor

FlashyReese:
These commits were cherrypicked and port from the `1.17.x/0.4-experiments` branch.
- Removes the matrix helpers and exposes those fields with access wideners.


IMS212:
This code is no longer applicable in dev.

PR #1503, opened 2022-11-07
Scale the debug pie with the rest of the GUI

adryd325:
On Hi-DPI displays the Debug pie is nearly unreadable
image

This is a vanilla issue, but I feel that this bug fix would fit well in Sodium.

Preview of the fix:
image

The commit could likely be cherry-picked onto 1.19.x/dev


MeeniMc:
As of 1.21.3 this is resolved in vanilla

RaphyTwin:
I have the same issue. I am using Sodium with 1.21.3

PR #1546, opened 2022-12-20
Optimized culling to include backtracing

MCRcortex:
No known issues however included an error of 1 section to very significantly reduce false positives for culling,
I have not noticed any other culling issues.

The implementation does a backtrace from a potential section back to the camera, if the backtrace finds that a section hasnt been enqueued into the bfs at that point it discards the section from being added. It exploits the properties of the bfs and linear rendering


jellysquid3:
Pulled into https://github.com/CaffeineMC/sodium-fabric/tree/1.19.3/culling-fixes so we can prototype on this more.

PR #1547, opened 2022-12-22
Fix Github Build Status badge in README

uku3lig:
Simply makes the Github Build Status badge in the Bleeding Edge Build section work again, after the breaking change introduced here: badges/shields#8671


IMS212:
Thanks!

PR #1571, opened 2023-01-11
Replace ranged for loop with while loop, ~20% perfomance uplift

mjwells2002:
this improves the rust software rasterizer by about 20% in draw_cube_800px test,


jellysquid3:
After investigating this on Discord, we discovered that the for-in-range loop seems to result in some kind of issue that makes the generated machine code much worse. Trunk uses a while-loop now.

PR #1587, opened 2023-01-24
Add optimized `Direction.getFacing` implementation

magneticflux-:
The old implementation looped over all directions and computed the dot product for each one, looking for the largest. We take advantage of the fact that we're just looking for the closest _basis_ vector (_not_ any arbitrary vector) and compare the input vector's components directly, taking care to preserve the order of precedence for the zero vector.


magneticflux-:
(This is a micro-optimization, don't expect framerate improvements. It affects rendering glinted compasses, baking complex block models, orienting the "block break" overlay, and finding the source direction of finished raycasts)

jellysquid3 (approved):
LGTM. I never benchmarked this but it seems pretty slow. Testing this in item frame rendering made a considerable improvement. Thanks for the contribution.

PR #1622, opened 2023-02-14
Optimized Stitching from Iris

PepperCode1:


jellysquid3 (approved):
LGTM. This is also exceptionally well documented.

PR #1626, opened 2023-02-16
Feature: ray culling + frustum cache

MCRcortex:
Implements a variant of rayculling from vanilla which only checks 5 chunks instead of all chunks so that performance is better at the loss of some accuracy


jellysquid3:
This has been open for a long time, but the conclusion we're at currently is that we're not going to implement ray-culling in the graph search due to the very high cost, difficulty in optimizing it, and poor accuracy. If we do improve the culling algorithm further, it will likely be with software rasterization.

PR #1636, opened 2023-02-22
Add error screen when Indium is missing and mods request a FRAPI implementation

Technici4n:
I assume there are reasons why Sodium does not include (i.e. jar-in-jar) Indium directly, so here is a compromise that I think would be acceptable. The general idea is:
- Mods declare that they want a FRAPI implementation by adding the `requires_renderer` custom value.
- Sodium detects that, and if Indium is missing, displays a warning screen to the user.

This is what the screen looks like at the moment:
image

Of course, the screen likely needs to be tweaked more to be optimal!

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/screen/MissingIndiumScreen.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/screen/MissingIndiumScreen.java
@@ -1,0 +1,97 @@
+package me.jellysquid.mods.sodium.client.gui.screen;
+
+import me.jellysquid.mods.sodium.client.SodiumClientMod;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.screen.Screen;
+import net.minecraft.client.gui.widget.ButtonWidget;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.text.Text;
+import net.minecraft.util.Util;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class MissingIndiumScreen extends Screen {
+    public static boolean shouldShow() {
+        if (FabricLoader.getInstance().isModLoaded("indium")) {
+            return false;
+        }
+
+        List<ModContainer> modsRequiringRenderer = FabricLoader.getInstance().getAllMods().stream()
+                .filter(mod -> mod.getMetadata().containsCustomValue("requires_renderer"))
+                .toList();
+
+        if (modsRequiringRenderer.size() > 0) {
+            SodiumClientMod.logger().error("Found %d mods that require Indium to be installed: %s".formatted(
+                    modsRequiringRenderer.size(),
+                    modsRequiringRenderer.stream().map(mod -> mod.getMetadata().getId()).collect(Collectors.joining(", "))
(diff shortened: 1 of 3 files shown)


src/main/java/me/jellysquid/mods/sodium/client/gui/screen/MissingIndiumScreen.java
```diff
@@ -0,0 +1,97 @@
...
+            return false;
+        }
+    }
+
+    private static final String TEXT_BODY_RAW = """
+            Some mods require an implementation of the Fabric Rendering API to be
+            installed. Such an implementation is provided by Indium.
+            Please download and install Indium for these mods to render correctly.
+
+            The list of such mods can be found in the log file.
```
Technici4n, line 44: Maybe the list of mods (or at least the first X) should be displayed to the user directly.

src/main/resources/fabric.mod.json
```diff
@@ -24,7 +24,8 @@
     ]
   },
   "custom": {
-    "fabric-renderer-api-v1:contains_renderer": true
+    "fabric-renderer-api-v1:contains_renderer": true,
+    "requires_renderer": true
```
Technici4n, line 28: This is just for illustration purposes and for testing, this should not go in Sodium of course!

src/main/java/me/jellysquid/mods/sodium/mixin/core/MixinMinecraftClient.java
```diff
@@ -17,9 +18,12 @@ public class MixinMinecraftClient {
 
     @Inject(method = "<init>", at = @At("RETURN"))
     private void postInit(RunArgs args, CallbackInfo ci) {
+        var parent = MinecraftClient.getInstance().currentScreen;
+
         if (SodiumClientMod.options().isReadOnly()) {
-            var parent = MinecraftClient.getInstance().currentScreen;
             MinecraftClient.getInstance().setScreen(new ConfigCorruptedScreen(() -> parent));
+        } else if (MissingIndiumScreen.shouldShow()) {
+            MinecraftClient.getInstance().setScreen(new MissingIndiumScreen(() -> parent));
```
Technici4n, line 26: If the config is corrupted AND indium is missing, it is possible to skip the indium screen. I am not sure how important this is, and whether it is worth addressing.

src/main/java/me/jellysquid/mods/sodium/client/gui/screen/MissingIndiumScreen.java
```diff
@@ -0,0 +1,97 @@
...
+
+        this.addDrawableChild(ButtonWidget.builder(TEXT_BUTTON_DOWNLOAD_INDIUM_CURSEFORGE, (btn) -> {
+            Util.getOperatingSystem().open("https://www.curseforge.com/minecraft/mc-mods/indium");
+        }).dimensions(32, this.height - 40, 120, 20).build());
+
+        this.addDrawableChild(ButtonWidget.builder(TEXT_BUTTON_DOWNLOAD_INDIUM_MODRINTH, (btn) -> {
+            Util.getOperatingSystem().open("https://modrinth.com/mod/indium");
+        }).dimensions(this.width / 2 - 120 / 2, this.height - 40, 120, 20).build());
+
+        this.addDrawableChild(ButtonWidget.builder(TEXT_BUTTON_IGNORE_WARNING, (btn) -> {
```
Technici4n, line 75: Maybe this button should be hidden by default, and only appear after some config flag is toggled, to ensure that users don't accidentally click on it?

PR #1650, opened 2023-03-04
change: increase build performance by remapping once

marcus8448:
Instead of remapping both the api jar and the mod jar (which includes all of the api classes anyway), we can just remap the mod jar, then extract the api classes out of it

This is slightly hacky, but avoids doing double the work

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,3 @@
-import net.fabricmc.loom.task.RemapJarTask
-
 plugins {
     id 'fabric-loom' version '1.0-SNAPSHOT'
     id 'org.ajoberstar.grgit' version '4.1.0'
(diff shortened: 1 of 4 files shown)


jellysquid3:
This is a pretty neat solution. My only question is whether we can still use this API JAR in a Maven publication while attaching the correct metadata. Do you happen to know?

marcus8448:
I've done some testing, and have realized that loom actually fails to remap the api due to a bug.
I've opened a PR to loom to fix that (FabricMC/fabric-loom#840 - it works with this), but alternatively I can add a dummy fabric.mod.json into the api jar (instead of using the `Fabric-Loom-Remap` metadata flag).
Is there any other kind of metadata that you're concerned about?

marcus8448:
Okay, now the api jar includes a basic `fabric.mod.json` (instead of using `Fabric-Loom-Remap: true`) so loom versions <1.1 can use the resulting api jar. Mods JiJ-ing sodium api should no longer be an issue as making `sodium` provide `sodium-api` should prevent fabric from loading the other jars.

> whether we can still use this API JAR in a Maven publication while attaching the correct metadata

The following snippet publishes the remapped api jar which can be used like any other published mod jar
```gradle
plugins {
    id 'maven-publish'
}

publishing {
    publications {
        mavenJava(MavenPublication) {
            artifact(apiJar) {
                builtBy apiJar
            }
        }
    }
    repositories {
    }
}
```

PR #1669, opened 2023-03-21
 Fix face culling precision errors

PepperCode1:
- Use doubles instead of floats to represent the camera position and render bounds


5HT2 (commented):
This looks good to me, unless there was any particular reason in the choice of floats before.

PR #1672, opened 2023-03-23
Restore `VertexConsumer.quad` optimization and resolve #1354

### LINKED ISSUE #1354: Animated textures: Sodium should not require modders to implement explicit compat
I implemented the following workaround to allow Modern Industrialization machine and fluid textures rendered using a Block Entity Renderer to be animated with Sodium: https://github.com/AztechMC/Modern-Industrialization/commit/3ada9dd2b69f573a3fd3f2aee8b4039cba26e22e.

However, I don't like this: I don't think it's reasonable to expect that mods will all implement similar support for Sodium. Would it not be possible to flag the contained sprite as active when a `BakedQuad` is added to a vertex consumer (at least the one used by block entity renderers)?

PepperCode1:
- Restore optimization in `MixinBufferBuilder`
- Add additional optimization to `MixinBufferBuilder`
- Mark sprites used by `BufferBuilder.quad` as active; Resolve #1354
- Mark sprites used by `DrawableHelper.drawSprite` as active
- Fix `ColorU8.byteToNormalizedFloat` and add int variant
- Rename `SpriteExtended` to `SpriteContentsExtended`

### DIFF
--- a/src/api/java/net/caffeinemc/mods/sodium/api/util/ColorU8.java
+++ b/src/api/java/net/caffeinemc/mods/sodium/api/util/ColorU8.java
@@ -36,6 +36,15 @@ public interface ColorU8 {
      * @return The floating point value of the integer component in the range of 0.0..1.0
      */
     static float byteToNormalizedFloat(byte value) {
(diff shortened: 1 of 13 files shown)


IMS212:
Looks good to me.

Issue #1156, opened 2022-01-29
Adjustable Max FPS by 1


PweSol:
Currently, we can adjust the max fps only by the step of 5. Having the ability to adjust it by 1 is much more useful. For example, it's great if you want to achieve the optimal G-Sync experience (explanation [here](https://blurbusters.com/gsync/gsync101-input-lag-tests-and-settings/5/) or [here](https://forums.blurbusters.com/viewtopic.php?f=5&t=3073)).

zaki-hussain:
I think if you change line 98 of src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java from `.setControl(option -> new SliderControl(option, 5, 260, 5, ControlValueFormatter.fpsLimit()))` to `.setControl(option -> new SliderControl(option, 5, 260, 1, ControlValueFormatter.fpsLimit()))` the step should become 1. It's just changing the third number to whatever you want the step to be.

PweSol:
It works! Thanks!
_Only had to disable Sodium version check in Iris to make the game launch with self-compiled builds._

Dekamir:
I hope this becomes the default. FreeSync needs approx. 3 FPS lower than the refresh rate in order to work properly.
This would greatly help screen tearing.

PweSol:
<del>Does anyone know how to do this on the latest version? Sodium has been changed quite significantly internally and I can't find anything similar in order to change it.</del>
It's now in src/main/java/net/caffeinemc/sodium/config/user/UserConfigCategories.java

MeeniMc:
The rejected PR contains rationale why this issue is a wontfix.

jellysquid3:
We are not going to change this behavior, since vanilla now enforces that the config values are within their defined ranges.

PR #1698, opened 2023-04-17
Allow more precise adjustment of Max Framerate option

Calinou:
### Proposed Changes

This allows catering to displays with variable refresh rate enabled. For example, when using a 120 Hz monitor, you can set the maximum framerate to 117, enable V-Sync and get low input lag without tearing. This works because the framerate remains just below the display's maximum VRR range, which prevents queued frames from accumulating.

Previously, you would either have to edit the settings file manually to achieve this, use a third-party FPS limiter (which is harder to set up and increases input lag), or stick to a lower FPS limit that is below the maximum VRR range (such as 110 FPS for 120 Hz).

The exact value to use in this scenario depends on your monitor refresh rate. Values to use for common refresh rates are:

- 60 Hz: 58 FPS
- 120 Hz: 117 FPS
- 144 Hz: 141 FPS
- 240 Hz: 235 FPS

For other refresh rates, you can calculate the FPS limit to use with `refresh_rate - (int(refresh_rate / 60.0) + 1)`.

See <https://blurbusters.com/howto-low-lag-vsync-on/> for more information.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -95,7 +95,7 @@ public class SodiumGameOptionPages {
                 .add(OptionImpl.createBuilder(int.class, vanillaOpts)
                         .setName(Text.translatable("options.framerateLimit"))
                         .setTooltip(Text.translatable("sodium.options.fps_limit.tooltip"))
-                        .setControl(option -> new SliderControl(option, 10, 260, 10, ControlValueFormatter.fpsLimit()))
+                        .setControl(option -> new SliderControl(option, 10, 260, 1, ControlValueFormatter.fpsLimit()))
                         .setBinding((opts, value) -> {
                             opts.getMaxFps().setValue(value);
                             MinecraftClient.getInstance().getWindow().setFramerateLimit(value);


burgerindividual:
I don't think this will work due to the validation that Minecraft does when parsing the settings file. The reason the fps slider's increment was increased to 10 from 5 was for this reason, so it would match vanilla.

Calinou:
> I don't think this will work due to the validation that Minecraft does when parsing the settings file. The reason the fps slider's increment was increased to 10 from 5 was for this reason, so it would match vanilla.

I've been wondering if vanilla could change this, but it would need some person with networking:tm: to be able to reach out to Mojang about this 🙂 (since it's technically not a bug)

PweSol:
I've personally been using this for a long time https://github.com/CaffeineMC/sodium-fabric/issues/1156 and never had any noticeable issues.

After certain Sodium or Mc version, the FPS lock stopped saving in the menu but it still locks the actual game. This is probably related to the parsing of the settings file as mentioned by burgerguy.

However, there are still some more complications due to how Mc handles something else internally related to the frame rate (I forgot the details). We already had a more in-depth discussion about it in some other pull request or issue, but I can't find it.

tldr: this won't be merged / won't work until someone figures out and writes a better system of how Mc handles this

IMS212:
This won't be merged, as this A: breaks vanilla compatibility and B: introduces many inconsistencies in how the FPS limiting works.

Vanilla has intentionally added this limit so users cannot break their settings; if Sodium allows invalid values, moving back to Vanilla will reset the settings and annoy users.

PR #1756, opened 2023-05-12
new: Add mixin rule override-ability, disable `core` overrides

FlashyReese:
Allows mixin override-ability to particular rules and disables `core` mixin rule overrides.

I'm unsure if I should port over Lithium's mixin config here or to [mixin-config](https://github.com/CaffeineMC/mixin-config) repository and then apply it here after it's on the CaffeineMC maven.


altrisi:
Another option without changing the config system is to move it into a different mixin config (unless there's some dependency with others that causes issues).

jellysquid3:
I'm going to take this out of the Sodium 0.5 milestone, but we'll re-visit it for when we switch to using CaffeineConfig.

PR #1774, opened 2023-05-22
update actions dependencies

piuvas:


piuvas:
i don't even know if prs like these even get carried over to 1.20/dev

jellysquid3:
This has been fixed in trunk since the PR was created, but thanks for submitting a patch either way.

PR #1790, opened 2023-06-01
Fixes chunk culling errors

MCRcortex:
Does what it says on the tin


MeeniMc:
This works, tested in 
* 1.20.rc.1
* new world with seed `4498904200215007879`
* `/tp @s 19999424.75 83.65 2169.66 -2129.16 44.55`
* 0.4.10+rev.4c3a2ce-dirty (aka 1.20 beta 2 from IMS in sodium-testing) has the bug
* this version doesn't have the bug

Didn't observe any meaningful performance difference on my hardware, FPS (objective) and smoothness (subjective) are the same with unlimited FPS.

jellysquid3:
Resolved with https://github.com/CaffeineMC/sodium-fabric/commit/8a8aad0df3ec36d5246d6a2a6efc1d34a7e092b1 using a slightly different technique.

PR #1794, opened 2023-06-06
Error handling improvements in chunk build threads

comp500:
I noticed in some recent issue reports that chunk build exceptions weren't being propagated to the main thread as intended in https://github.com/CaffeineMC/sodium-fabric/pull/615. This is because deferred (non-important) chunk builds aren't polled from the main thread using `join()` unlike important chunk rebuilds, and since https://github.com/CaffeineMC/sodium-fabric/commit/696da3c219b0d21febdea9dd73e07516f694248a Sodium always treats initial chunk builds as deferred, so if a particular error makes a chunk never successfully build, `join()` will never be called on it!

This PR therefore correctly propagates these exceptions, along with extra handling for crash reporting that adds Sodium rendering context, with the following changes:

- Added a queue to store chunk build failures in `ChunkBuilder`
- Modified `scheduleDeferred` to add `CompletableFuture` exceptions to this queue
- Created `handlePendingFailures` in `RenderSectionManager` to drain the failure queue and throw any found exception
- Modified `WorkStealingFutureDrain.findNext` to unwrap `CompletionException`s that contain `CrashException` so that the crash report context is preserved
- Wrapped the `ChunkRenderRebuildTask.performBuild` outer loop in a try-catch block to handle exceptions from chunk building and add chunk section / render context information
- Improved `RenderSection.toString` to provide more useful information
- Changed `ChunkBuilder.processJob` to use the Sodium logger rather than `printStackTrace`

For `scheduleDeferred` I changed it to use `whenComplete` as it doesn't wrap the exception in `CompletionException` (so any `CrashException` can be propagated directly) and it allows us to return `CompletableFuture<ChunkBuildResult>` matching the return type of `schedule`.

This is what the crash report context looks like:

```
-- Block being rendered --
Details:
	Block: Block{minecraft:gravel}
	Block location: World: (-16,-16,-48), Section: (at 0,0,0 in -1,-1,-3; chunk contains blocks -16,-64,-48 to -1,319,-33), Region: (-1,-1; contains chunks -32,-32 to -1,-1, blocks -512,-64,-512 to -1,319,-1)
	Chunk section: RenderSection at chunk (-1, -1, -3) from (-16, -16, -48) to (0, 0, -32)
	Render context volume: BlockBox{minX=-18, minY=-18, minZ=-50, maxX=1, maxY=1, maxZ=-31}
```

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
@@ -20,9 +20,9 @@ public class ChunkJobResult<OUTPUT> {
     }
 
     public OUTPUT unwrap() {
-        if (this.throwable instanceof CrashException) {
+        if (this.throwable instanceof CrashException crashException) {
             // Propagate CrashExceptions directly to provide extra information
-            throw (CrashException)this.throwable;
+            throw crashException;
         } else if (this.throwable != null) {
             throw new RuntimeException("Exception thrown while executing job", this.throwable);
         }
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
@@ -1,5 +1,7 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile.executor;
 
+import net.minecraft.util.crash.CrashException;
+
 public class ChunkJobResult<OUTPUT> {
     private final OUTPUT output;
     private final Throwable throwable;
@@ -18,7 +20,10 @@ public class ChunkJobResult<OUTPUT> {
     }
 
     public OUTPUT unwrap() {
-        if (this.throwable != null) {
+        if (this.throwable instanceof CrashException) {
+            // Propagate CrashExceptions directly to provide extra information
+            throw (CrashException)this.throwable;
+        } else if (this.throwable != null) {
             throw new RuntimeException("Exception thrown while executing job", this.throwable);
         }
 
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobTyped.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobTyped.java
@@ -1,5 +1,6 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile.executor;
 
+import me.jellysquid.mods.sodium.client.SodiumClientMod;
 import me.jellysquid.mods.sodium.client.render.chunk.compile.ChunkBuildContext;
 import me.jellysquid.mods.sodium.client.render.chunk.compile.tasks.ChunkBuilderTask;
 
@@ -48,6 +49,7 @@ public class ChunkJobTyped<TASK extends ChunkBuilderTask<OUTPUT>, OUTPUT>
             result = ChunkJobResult.successfully(output);
         } catch (Throwable throwable) {
             result = ChunkJobResult.exceptionally(throwable);
+            SodiumClientMod.logger().error("Chunk build failed", throwable);
         } finally {
             this.task.releaseResources();
         }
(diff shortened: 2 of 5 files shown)


jellysquid3:
Can you rebase this against the `/dev` branch? We changed a lot to do with the `ChunkBuilder` but the crash report functionality you added should still work. We also now store a "job output" queue which the main thread polls every frame, and it handles unwrapping any exceptions if they exist. And because of this, there is now only one location where the results get unwrapped, making it a lot easier to report these exceptions.

comp500:
Updated list of changes after rebase:

- ~~Added a queue to store chunk build failures in `ChunkBuilder`~~
- ~~Modified `scheduleDeferred` to add `CompletableFuture` exceptions to this queue~~
- ~~Created `handlePendingFailures` in `RenderSectionManager` to drain the failure queue and throw any found exception~~
- Modified ~~`WorkStealingFutureDrain.findNext`~~ `ChunkJobResult.unwrap` to unwrap `CompletionException`s that contain `CrashException` so that the crash report context is preserved
- Wrapped the `ChunkRenderMeshingTask.performBuild` outer loop in a try-catch block to handle exceptions from chunk building and add chunk section / render context information
- Improved `RenderSection.toString` to provide more useful information
- Updated `ChunkJobTyped.execute` to use the Sodium logger (there may be multiple chunk build errors before the main thread gets the results and aborts, so it's useful to log everything)

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
```diff
@@ -18,7 +20,10 @@ public static <OUTPUT> ChunkJobResult<OUTPUT> successfully(OUTPUT output) {
     }
 
     public OUTPUT unwrap() {
-        if (this.throwable != null) {
+        if (this.throwable instanceof CrashException) {
```
jellysquid3: Should use `if (this.throwable instanceof CrashException crashException)` since it does the instanceof and type-cast in one.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobTyped.java
```diff
@@ -48,6 +49,7 @@ public void execute(ChunkBuildContext context) {
             result = ChunkJobResult.successfully(output);
         } catch (Throwable throwable) {
             result = ChunkJobResult.exceptionally(throwable);
+            SodiumClientMod.logger().error("Chunk build failed", throwable);
```
jellysquid3: We've been trying to move away from the global logger and to using class-specific loggers in Sodium. Can you make the logger in `ChunkBuilder` package-private and use it here instead?

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/tasks/ChunkBuilderMeshingTask.java
```diff
@@ -139,6 +151,23 @@ public ChunkBuildOutput execute(ChunkBuildContext buildContext, CancellationToke
...
+        CrashReportSection crashReportSection = report.addElement("Block being rendered", 1);
+
+        BlockState state = null;
+        try {
+            state = slice.getBlockState(pos);
+        } catch (Exception ignored) {}
+        CrashReportSection.addBlockInfo(crashReportSection, slice, pos, state);
+
+        crashReportSection.add("Chunk section", render);
+        if (renderContext != null) {
```
jellysquid3: Style nitpick: `this` prefix needed for instance field accesses.

jellysquid3 (changes requested):

jellysquid3:
Changes look good to me outside of minor nitpicks; will merge when resolved.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/executor/ChunkJobResult.java
comp500: Ah, I forgot that was a feature! Been writing Java 8 for too long...

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/tasks/ChunkBuilderMeshingTask.java
```diff
@@ -101,6 +102,9 @@ public ChunkBuildOutput execute(ChunkBuildContext buildContext, CancellationToke
                             context.update(blockPos, modelOffset, blockState, model, seed);
                             cache.getBlockRenderer()
                                 .renderModel(context, buffers);
+                            if (Random.createLocal().nextInt(100000) == 0) {
+                                throw new RuntimeException("henlo");
+                            }
```
Octol1ttle: Debug leftovers?

comp500: Oops...

jellysquid3 (approved):

PR #1795, opened 2023-06-07
Avoid List#get in BufferBuilder#nextElement

Moulberry:
Calling List#get in BufferBuilder#nextElement has to perform a lot of pointer accesses.

BufferBuilder -> VertexFormat -> List -> Map -> Array -> Entry -> VertexFormatElement

Let's not do that, shall we?

Use a single array inside BufferBuilder instead

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/fast_advance/MixinBufferBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/fast_advance/MixinBufferBuilder.java
@@ -27,7 +27,7 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer implem
     private int currentElementId;
 
     @Unique
-    private VertexFormatElement[] formatElements = new VertexFormatElement[16];
+    private VertexFormatElement[] formatElements = new VertexFormatElement[32];
     @Unique
     private int formatElementCount = 0;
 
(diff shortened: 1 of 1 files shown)


src/main/java/me/jellysquid/mods/sodium/mixin/features/buffer_builder/fast_advance/MixinBufferBuilder.java
```diff
@@ -20,24 +26,36 @@ public abstract class MixinBufferBuilder extends FixedColorVertexConsumer implem
...
+
+    @Inject(method = "setFormat",
+        at = @At(
+            value = "FIELD",
+            target = "Lnet/minecraft/client/render/BufferBuilder;format:Lnet/minecraft/client/render/VertexFormat;",
+            opcode = Opcodes.PUTFIELD
+        )
+    )
+    private void onFormatChanged(VertexFormat format, CallbackInfo ci) {
+        this.formatElements = format.getElements().toArray(new VertexFormatElement[0]);
```
jellysquid3: I don't particularly like the memory allocation here, but I'm not sure whether this function is called enough to matter...

Maybe we can store a `VertexFormatElement[32]` in the `BufferBuilder`, and then re-use the array? OpenGL only allows a maximum of 32 vertex attributes. We'd just need to store the "actual" number of vertex elements alongside the array.

Furthermore, we could strip the `PADDING` elements so that `BufferBuilder.nextElement` doesn't need to check & skip them. That might result in better code generation.

jellysquid3:
Closing as this is no longer relevant in the latest version of Minecraft.

PR #1796, opened 2023-06-07
BufferBuilder: Avoid format lookup when format doesn't change

Moulberry:
Previously, onFormatChanged was invoked every time BufferBuilder#begin was called.
 
VertexFormatRegistryImpl#get then needs to acquire a lock, do a map lookup
and release a lock
 
Let's only do that when the format actually changes


jellysquid3 (approved):
LGTM. This is simple enough change and it makes sense, even if the performance impact is likely minimal.

PR #1807, opened 2023-06-12
Remove unnecessary RefCells

wahfl2:
I removed all of the RefCells with the addition of a Funsafe trait to "get_several_mut" from a HashMap (which should definitely be renamed).

### DIFF
--- a/native/core/src/graph.rs
+++ b/native/core/src/graph.rs
@@ -508,6 +508,10 @@ impl Graph {
                             .search_state
                             .enqueue(neighbor, GraphDirectionSet::single(direction.opposite()));
                     }
+                } else {
+                    context.origin
+                        .search_state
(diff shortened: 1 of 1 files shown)


jellysquid3:
Closing this, since we're using raw pointers now, and the code is a little different. Though your patch served as useful reason to investigate this much more closely. I think what you've written is obviously "safer" (as it verifies that none of the keys are overlapping), but I think we can reasonably assume they never will be, since this is an internal function of `SearchContext`, and it will always access unique keys.

Not using RefCells did slightly improve the code generation, since `SearchContext.adjacent(Direction)` doesn't need to de-structure the `RefCell` struct to get the underlying pointer. The in-memory representation of the array is essentially just a `[usize; 6]`, which allows the compiler to produce a single `MOV dst, src[base + (idx << 3)]` instruction (or some variant using memory addressing.)

PR #1820, opened 2023-06-17
Fix broken normals in WorldRenderer#drawBox

Moulberry:
The original author got the rows & columns the wrong way around.
This PR swaps the rows and columns. eg. m00, m10, m20 becomes m00, m01, m02.

This was done using using the following find & replace:
Find:    \.m(\d)(\d)
Replace: .m$2$1

Before (current): https://cdn.discordapp.com/attachments/867325617367810071/1119571223974772797/image.png
After (this PR): https://cdn.discordapp.com/attachments/867325617367810071/1119570983230111825/image.png


jellysquid3 (approved):

jellysquid3:
Good catch! I noticed the line rendering behavior was a little strange, but didn't have time to investigate further. Thanks for the patch.

PR #1842, opened 2023-06-23
fix: use vanilla render path and log warning when incompatible VertexConsumer is detected

embeddedt:
I am not an expert on rendering internals, can probably be improved.


src/api/java/net/caffeinemc/mods/sodium/api/vertex/buffer/VertexBufferWriter.java
```diff
@@ -22,6 +22,21 @@ static VertexBufferWriter of(VertexConsumer consumer) {
...
     }
 
+    /**
+     * Converts a {@link VertexConsumer} into a {@link VertexBufferWriter} if possible.
+     *
+     * @param consumer The vertex consumer to create a writer for
+     * @return An implementation of {@link VertexBufferWriter} which will write vertices into {@param consumer}, or null
+     * if the vertex consumer does not implement the necessary interface
+     */
+    static VertexBufferWriter tryOf(VertexConsumer consumer) {
```
Technici4n, line 34: `@Nullable`?

src/main/java/me/jellysquid/mods/sodium/client/render/vertex/VertexConsumerTracker.java
```diff
@@ -0,0 +1,16 @@
...
+import it.unimi.dsi.fastutil.objects.ReferenceOpenHashSet;
+import it.unimi.dsi.fastutil.objects.ReferenceSet;
+import me.jellysquid.mods.sodium.client.SodiumClientMod;
+import net.minecraft.client.render.VertexConsumer;
+
+public class VertexConsumerTracker {
+    private static final ReferenceSet<Class<? extends VertexConsumer>> BAD_CONSUMERS = new ReferenceOpenHashSet<>();
+
+    public static void logBadConsumer(VertexConsumer consumer) {
+        if(BAD_CONSUMERS.add(consumer.getClass())) {
```
Technici4n, line 12: Are there thread safety concerns here? If so then a `Collections.newSetFromMap(new ConcurrentHashMap<>())` can be used.

src/main/java/me/jellysquid/mods/sodium/mixin/features/block/MixinBlockModelRenderer.java
```diff
@@ -30,9 +34,14 @@ public class MixinBlockModelRenderer {
...
-    public void render(MatrixStack.Entry entry, VertexConsumer vertexConsumer, BlockState blockState, BakedModel bakedModel, float red, float green, float blue, int light, int overlay) {
-        var writer = VertexBufferWriter.of(vertexConsumer);
+    @Inject(method = "render(Lnet/minecraft/client/util/math/MatrixStack$Entry;Lnet/minecraft/client/render/VertexConsumer;Lnet/minecraft/block/BlockState;Lnet/minecraft/client/render/model/BakedModel;FFFII)V", at = @At("HEAD"), cancellable = true)
+    private void renderFast(MatrixStack.Entry entry, VertexConsumer vertexConsumer, BlockState blockState, BakedModel bakedModel, float red, float green, float blue, int light, int overlay, CallbackInfo ci) {
+        var writer = VertexConsumerUtils.convertOrLog(vertexConsumer);
+        if(writer == null) {
+            return;
+        }
+
+        ci.cancel();
```
FlashyReese, line 44: Isn't this just going to always cancel? Shouldn't this only cancel on when the writer is null?

src/main/java/me/jellysquid/mods/sodium/mixin/features/entity/fast_render/MixinModelPart.java
```diff
@@ -53,9 +54,15 @@ private void onInit(List<ModelPart.Cuboid> cuboids, Map<String, ModelPart> child
...
-    private void renderCuboids(MatrixStack.Entry matrices, VertexConsumer vertexConsumer, int light, int overlay, float red, float green, float blue, float alpha) {
-        var writer = VertexBufferWriter.of(vertexConsumer);
+    @Inject(method = "renderCuboids", at = @At("HEAD"), cancellable = true)
+    private void renderCuboidsFast(MatrixStack.Entry matrices, VertexConsumer vertexConsumer, int light, int overlay, float red, float green, float blue, float alpha, CallbackInfo ci) {
+        var writer = VertexConsumerUtils.convertOrLog(vertexConsumer);
+        if(writer == null) {
+            return;
+        }
+
+        ci.cancel();
```
FlashyReese, line 64: The comment above applies here as well.

src/main/java/me/jellysquid/mods/sodium/mixin/features/item/MixinItemRenderer.java
```diff
@@ -34,9 +38,14 @@ public class MixinItemRenderer {
...
-    private void renderBakedItemModel(BakedModel model, ItemStack itemStack, int light, int overlay, MatrixStack matrixStack, VertexConsumer vertexConsumer) {
-        var writer = VertexBufferWriter.of(vertexConsumer);
+    @Inject(method = "renderBakedItemModel", at = @At("HEAD"), cancellable = true)
+    private void renderModelFast(BakedModel model, ItemStack itemStack, int light, int overlay, MatrixStack matrixStack, VertexConsumer vertexConsumer, CallbackInfo ci) {
+        var writer = VertexConsumerUtils.convertOrLog(vertexConsumer);
+        if(writer == null) {
+            return;
+        }
+
+        ci.cancel();
```
FlashyReese, line 48: Here as well

src/main/java/me/jellysquid/mods/sodium/mixin/features/block/MixinBlockModelRenderer.java
embeddedt: No. Immediately above this line the function returns without cancelling when the writer is null.

FlashyReese: I thought `CallbackInfo#cancel` is effectively a premature method return.

embeddedt: The injection into the target method works like this:

```java
public void originalMethod(int x) {
    CallbackInfo ci = new CallbackInfo();
    this.handler$inject$rrr000(x, ci);
    if(ci.isCancelled())
        return;
}
```

so it does not matter whether you cancel at the beginning or the end inside the injected method.

embeddedt: In this case we cancel when we have a non-null writer as Sodium handles the rendering itself, but skip cancelling for a null writer as we want the vanilla logic below the injected handler to run.

jellysquid3 (approved):

PR #1851, opened 2023-06-27
Implement full support for Fabric Rendering API

Technici4n:
This PR implements full FRAPI support for Sodium. There are 3 pipelines: terrain, non-terrain, and item rendering. Terrain meshing was converted to always go through a FRAPI-aware pipeline. This is a heavily modified Indigo pipeline with new optimizations to use Sodium's chunk meshing caches to the largest possible extent. Non-terrain and item pipelines are only used for FRAPI models, and as such most code comes from Indigo.

To test this PR, compare the performance of the two meshing pipelines by enabling and disabling the `Enable FRAPI terrain pipeline` option under performance.

### Bundled Indigo
Everything under [`me.jellysquid.mods.sodium.client.frapi`](https://github.com/Technici4n/sodium-fabric/tree/frapi-support/src/main/java/me/jellysquid/mods/sodium/client/frapi) is directly copy-pasted from Indigo. This handles all the `QuadView`, `Mesh` and `RenderMaterial` encoding for us.

`QuadViewImpl` was modified to also store the `Sprite`, as Sodium uses it for animated texture optimization.

An exception is the `render` subpackage which contains Indigo code, but heavily modified for use in Sodium.

The AO calculations of Indigo were left out as Sodium already contains what is needed.

### BlockRenderer
[`BlockRenderer`](https://github.com/Technici4n/sodium-fabric/blob/frapi-support/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java) is the core class of this PR. It is completely rewritten to use a FRAPI-aware pipeline, with some optimizations for vanilla models however.

The way this works is that the quads end up in `processQuad` where they are processed and buffered one-by-one.

The key question is how much performance we lose from having to perform a bunch of chunks for light mode, material, etc... for each quad. From my testing, the new FRAPI pipeline has comparable performance to the original Sodium pipeline. There is 15-20% overhead from FRAPI, but this is compensated by not having to go though `WeightedBakedModel#getQuads` 7 times per face, so overall the performance should be comparable to current Sodium.

### Non-terrain block pipeline
This was ported from Indigo, then modified heavily to use Sodium's light pipeline. This is only used for specific calls to the `BlockModelRenderer`, and only for FRAPI models. The core class is [`NonTerrainBlockRenderContext`](https://github.com/Technici4n/sodium-fabric/blob/frapi-support/src/main/java/me/jellysquid/mods/sodium/client/frapi/render/NonTerrainBlockRenderContext.java).

### Item pipeline
This pipeline was ported from Indigo too, and is only used for FRAPI item models. The core class is [`ItemRenderContext`](https://github.com/Technici4n/sodium-fabric/blob/frapi-support/src/main/java/me/jellysquid/mods/sodium/client/frapi/render/ItemRenderContext.java).


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/BlockRendererFRAPI.java
```diff
@@ -0,0 +1,298 @@
...
+    private LightMode getLightingMode(BlockState state, BakedModel model) {
+        if (this.useAmbientOcclusion && model.useAmbientOcclusion() && state.getLuminance() == 0) {
+            return LightMode.SMOOTH;
+        } else {
+            return LightMode.FLAT;
+        }
+    }
+
+    private class Context extends AbstractRenderContext {
+        private void renderQuad(MutableQuadViewImpl quad) {
```
Technici4n: This is the entrypoint of the quad processing pipeline.

mrjasonn:
This really deserves to be used. It would be a very good feature. However, if it doesn't support Sodium 0.5, which we don't know as it's not released at the time of writing this, then it would need updating. Same with Fabric API and possibly Minecraft changes as well. @jellysquid3 if you haven't noticed this, this is actually decent. But it has to be stable enough that things won't break though which requires checking.

Technici4n:
It was noticed and discussed a bit already. It will not be in the 0.5 release, however it is likely that it will be in a subsequent one.

mrjasonn:
> It was noticed and discussed a bit already. It will not be in the 0.5 release, however it is likely that it will be in a subsequent one.

Then do you have any idea of which release it will be in? Will we need to wait until after the whole major 0.5 release or it will be in a release like 0.5.1? @Technici4n

Technici4n:
Unclear yet. For now you can use Indium.

Onahil:
pls respond i have read your post and im just confused about how i get rubidium to to work nothing against you im just dumb so pls just explain it to me like a monkey

TheUserCreated:
Rubidium is a fork not affiliated with sodium, this is the wrong place to ask about it.

Onahil:
so this makes sodium work, so how do i make sodium with connector

Onahil:
how does this make sodium work i am confused i may be and idiot so PLEASE explainhow i do this

TheUserCreated:
Sodium works with continuity when indium is installed. If you're using forge ports then ask elsewhere as those are unrelated to sodium. This is not a support forum, it's an issue tracker.

jellysquid3:
User was banned from the issue tracker for spamming issues & being utterly helpless.

jellysquid3:
That distraction aside, this is something we want to focus on & get implemented/merged for the next major version of Sodium.  It's long over-due.

Lolothepro:
@jellysquid3
image
Do you want to modify it, or do you still want to check the PR? (same for #2043)

Lolothepro:
ok

jellysquid3:
This has been rolled up into https://github.com/CaffeineMC/sodium-fabric/commit/d2df1838a36abf75198440d03f8f2c08085f395b.

PR #1855, opened 2023-06-30
Light pipeline fixes

### LINKED ISSUE #1852: [SCULK SENSOR] Sculk sensors emit too much light
### Version information

0.4.10+build.27

### Expected Behavior

image

### Actual Behavior

image

### Reproduction Steps

.

### Java version

Java 17

### CPU

Ryzen 3600

### GPU

GTX 1070

### Additional information

[Optifine](https://github.com/sp614x/optifine/issues/7445):
image

PepperCode1:
- Only use non-emissive lightmaps when calculating the minimum lightmap in `AoFaceData#calculateCornerBrightness`
- Ignore emissivity for offset pos in `FlatLightPipeline#getOffsetLightmap` to better match vanilla
- Update center lightmap logic to match vanilla
- Update OP light data flag to match javadoc; previously, it was inverted
- Move `BlockOcclusionCache` to the same package as `BlockRenderer` as that is the only place it is used
- Use normal orientation instead of flipped when summed AO values are equal
- Honor contract of `ModelQuadFlags.IS_ALIGNED` in `FluidRenderer`
- Update some comments

Fixes #1852. This is accomplished by the first point. When the fullbright lightmap from emissive blocks is used to calculate the minimum (non-zero) lightmap in a case where all other lightmap values are 0, the fullbright lightmap ends up being the minimum, and so the mean lightmap is also the fullbright lightmap. Now, the cached lightmap value excludes the emissive lightmap and a separate emissive flag is cached alongside it. The flag is checked and the fullbright lightmap is now only applied after the minimum lightmap has been calculated and applied. This separation also allowed for the second point.


jellysquid3 (approved):
I don't notice any issues with this, and the changes look good to me. We'll _really_ want to clean up the code in `LightDataAccess` at some point, but that doesn't need to be done in this PR.

jellysquid3:
Merged manually as 4f163b106132dc30496c6ec93590ffa26a5d8721 to resolve conflicts.

PR #1856, opened 2023-07-01
ChunkBuilder rewrite

MCRcortex:
Draft pr to rewrite the ChunkBuilder to be more stable and cause less lag spikes


jellysquid3:
This really needs a high-level overview as to what you're doing in the PR description... The code isn't documented either, and making a bunch of changes related to concurrent code is not exactly the most trivial thing to understand.

MCRcortex:
The pr changes the ChunkBuilder system to work as follows
ChunkBuilder has a pool of worker threads

When a ChunkRenderBuildTask is requested the ChunkBuilder places it at the back of the work queue if it is not an important rebuild or at the front of the queue if it is an important rebuild.
After enqueueing the task, it will release a permit on the jobCount semaphore

Each worker thread tries to acquire a permit from the jobCount semaphore, blocking until it obtains one, once it obtains a permit it dequeues a job from the queue (which is guaranteed to be there as only workers can dequeue from the inbound task queue), it then checks if the job has been canceled, if it has then it aborts the job. it then tries to aquire the executing status of the job atomically, if it obtains the executing role, it performs the chunk build, otherwise it aborts the execution.
After a job is finished, if it is a non important build it gets enqueued into the deferredResultQueue queue.

Now the thing with important builds is that they must be done before the render thread can continue, that is why they are done a bit differently. 
(Note the following is how the pr will be before its taken out of draft, not the current implementation of the iterator)
when the render thread gets to the phase of checking/uploading important chunk builds it loops over all the important chunk tasks, if it sees that one is finished, submit that, do this again until there are no finished important chunk builds. 
If there are no finished chunk builds, loop through all the important builds and try to obtain the execution role, (this will only happen if a worker thread is not executing the chunk rebuild task) if the role is obtained, execute the task on the render thread and submit it for rebuild and start back at step one of checking for finished builds, if all the builds are currently being executed spinlock on waiting for a build to complete

jellysquid3:
Resolved with https://github.com/CaffeineMC/sodium-fabric/commit/fb1eb2e70b7ce557508d0af934b2b291263464a0.

PR #1871, opened 2023-07-07
Fix fluid height check in fluid renderer

### LINKED ISSUE #1870: 1.20/dev: Fluids are not rendered correctly
### Version information

***

### Expected Behavior

image

### Actual Behavior

image

### Reproduction Steps

Easily reproducible in a cold biome.

### Java version

Java 17

### CPU

Ryzen 3600

### GPU

GTX 1070

### Additional information

_No response_

Grayray75:
Currently the fluid renderer calculates an incorrect height for blocks like shulker boxes or beacons.

Problem is that it incorrectly checks if a block is not opaque, when it should check if a block is not solid.

1.20.1 Vanilla:
```java
return !blockState.isSolid() ? 0.0F : -1.0F;
```

Current Sodium:
```java
if (!blockState.isOpaque()) {
    return 0.0f;
}
return -1.0f;
```

fixes #1870


IMS212:
Thanks!

PR #1873, opened 2023-07-10
fix: Use GUI render layer when drawing rectangles for options

FlashyReese:
# Bug Discovery Process

While investigating a specific scenario involving the following mods:

- `sodium-fabric-mc1.20-0.4.10+build.27`
- `owo-lib-0.11.1+1.20`
- `fabric-api-0.85.0+1.20.1`
- `isometric-renders-0.4.4+1.20`

When joining a world, the options rectangles were not being rendered correctly, as shown in the image below:

Before PR

# Bug Explanation
Starting from version 1.20, a significant change was made to the `DrawableHelper` utility. Previously, it was commonly subclassed and used by passing `MatrixStack` matrices parameters. However, in the latest update, `DrawableHelper` has been replaced by `DrawContext`, which is now directly passed as an object to rendering methods. This eliminates the need to manually construct our own. **In addition to this change, the rectangle rendering was not occurring in the dedicated GUI layer, which was part of the recent update. As a result, the rectangles were not rendering properly.**

# Bug Resolution

After implementing this PR, the options screens now display correctly, as depicted in the following image:

After PR

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/ControlValueFormatter.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/ControlValueFormatter.java
@@ -4,48 +4,48 @@ import net.minecraft.text.Text;
 
 public interface ControlValueFormatter {
     static ControlValueFormatter guiScale() {
-        return (v) -> (v == 0) ? Text.translatable("options.guiScale.auto").getString() : v + "x";
+        return (v) -> (v == 0) ? Text.translatable("options.guiScale.auto") : Text.literal(v + "x");
     }
 
     static ControlValueFormatter fpsLimit() {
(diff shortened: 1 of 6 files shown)


src/main/java/me/jellysquid/mods/sodium/client/gui/options/control/ControlValueFormatter.java
```diff
@@ -4,48 +4,48 @@
...
     }
 
     static ControlValueFormatter percentage() {
-        return (v) -> v + "%";
+        return (v) -> Text.literal(v + "%");
     }
 
     static ControlValueFormatter multiplier() {
-        return (v) -> v + "x";
+        return (v) -> Text.literal(v + "%");
```
IMS212: Is this right? You replaced `x` with `%`.

IMS212 (changes requested):
Looks good, just one question.

IMS212:
Thanks!

Issue #1872, opened 2023-07-09
Decorated pot bottom rendering as solid color


JustinTimeCuber:
### Version information

sodium-fabric-mc1.20-0.4.10+build.27

### Expected Behavior

The bottom of the decorated pot is textured correctly:

![2023-07-08_19 22 23](https://github.com/CaffeineMC/sodium-fabric/assets/46458276/0e045b40-3c1f-4898-b1b2-c8ab37be718e)

### Actual Behavior

The bottom of the decorated pot is a solid color, taken from the (5, 10) pixel of the side texture.

![2023-07-08_19 23 03](https://github.com/CaffeineMC/sodium-fabric/assets/46458276/7f674e44-28eb-47b1-9ebe-8d46f2731ed2)

![2023-07-08_18 25 46](https://github.com/CaffeineMC/sodium-fabric/assets/46458276/eb00eabb-d393-45f9-af56-7f0ef0f374f9)

### Reproduction Steps

1. Place a decorated pot in the air
2. View the block from below
3. Observe that the texture is not rendered correctly when Sodium is installed

### Java version

Java 17.0.7

### CPU

Intel Core i9-13900K

### GPU

GeForce RTX 3080 Ti

### Additional information

Observed while trying to fix the decorated pot model in [Chunky](https://github.com/chunky-dev/chunky/pull/1602), at first I thought it was a vanilla bug but it does not occur in a fresh install, and it does still occur with only Sodium and no other mods.

jellysquid3:
That's... interesting.

I have not even the slightest clue what is going wrong here.

Fnige:
Is there any models similar to the potted plant that could be looked at?

PR #1874, opened 2023-07-10
Fix decorated pot bottom rendering as solid color

Gaider10:
Fixes https://github.com/CaffeineMC/sodium-fabric/issues/1872


IMS212:
Thanks!

PR #1900, opened 2023-07-22
Add nvidia linux driver workaround to disable gl thread optimizations via environment variable 

MCRcortex:


src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/DriverWorkarounds.java
```diff
@@ -13,6 +13,10 @@ public static void beforeContextCreation() {
         if (Workarounds.isWorkaroundEnabled(Workarounds.Reference.NVIDIA_BAD_DRIVER_SETTINGS)) {
             NVIDIAWorkarounds.install();
         }
+
+        if (Workarounds.isWorkaroundEnabled(Workarounds.Reference.NVIDIA_BAD_DRIVER_LINUX)) {
```
jellysquid3: This should just use the existing `NVIDIA_BAD_DRIVER_SETTINGS` option and remove the check for Linux.

src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/DriverWorkarounds.java
```diff
@@ -13,6 +13,10 @@ public static void beforeContextCreation() {
...
+            NVIDIAWorkarounds.setLinuxDisableEnv();
```
jellysquid3: Rather than directly call into the Linux code, just have `NVIDIAWorkarounds.install` pick the code path depending on the operating system.

jellysquid3 (changes requested):

jellysquid3 (approved):

PR #1904, opened 2023-07-22
High contrast mode

IMS212:


douira:
High-contrast mode for our UI could be looked into, but this PR is entirely out of date now.

PR #1905, opened 2023-07-22
Fix Sodium option sliders not respecting vanilla clamping

embeddedt:
### Proposed Changes

Currently the Sodium option sliders for view & simulation distance use hardcoded minimum/maximum values instead of pulling these values from the vanilla option objects. This introduces a parity break from vanilla, as it normally clamps the maximum view distance to 16 if less than 1GB of RAM is allocated to the game. It also means that any mods modifying the default bounds vanilla places on these options (e.g. to lower minimum simulation distance) would need to patch Sodium for the change to apply.

This PR fixes that by attempting to retrieve the vanilla value for the minimum and maximum, and falling back to the Sodium-provided constant if not possible. I used wrapper functions to try and keep the code in the actual Sodium options as simple as possible.

If there is a technical reason why Sodium should not respect the vanilla bounding, this PR doesn't need to be merged, but it was a simple enough change that I felt I should just go ahead and submit anyway.

Attached is a screenshot showing the fix for the parity break described earlier (heap size set to 512MB, and maximum render distance is now clamped to 16).

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -19,22 +19,45 @@ import net.minecraft.text.Text;
 import org.lwjgl.opengl.GL;
 import org.lwjgl.opengl.GLCapabilities;
 
-import javax.sound.sampled.BooleanControl;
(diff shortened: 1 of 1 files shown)


mrjasonn:
Why not just make this add a memory warning instead? I mean Sodium has already very much optimised the rendering engine.

PR #1906, opened 2023-07-22
Show "No Error Context" option as unavailable under Wayland

Octol1ttle:
As of https://github.com/CaffeineMC/sodium-fabric/commit/fa594b4d88c7e8736e235492607d3fefc83272cd, "No Error Context" is force-disabled on Wayland, but the option is still togglable in video settings, which might mislead the user into thinking changing the option does something. This PR prevents the option from being togglable and crosses it out if the workaround is currently active to indicate that the option is not available.


jellysquid3 (approved):

jellysquid3:
Seems reasonable enough.

PR #1918, opened 2023-07-28
Restrict total number of chunk update threads based off heap size

embeddedt:
### Proposed Changes

With the right combination of optimization mods and configs it is possible to run the game on very small heap sizes (<=256MB). However, due to Sodium allowing users to configure the number of chunk builder threads as high as they would like, it is possible to starve the game of memory by having too many threads attempting to update chunks at once.

This PR solves that issue by introducing additional logic that caps the number of threads based off maximum heap size. Vanilla's chunk builder uses its own heuristic dependent on the `RenderLayer`s' expected buffer sizes. Here, we use simpler logic, and allow one thread for every 64MB of heap. On the default vanilla RAM allocation (2GB) this would permit 32 threads, which should still be plenty (Sodium doesn't even try to use more than 10 by default).

Example screenshot attached showing that the maximum number of threads is now 2 even on a 4-thread system due to the low heap size.

image

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -11,6 +11,7 @@ import me.jellysquid.mods.sodium.client.gui.options.control.SliderControl;
 import me.jellysquid.mods.sodium.client.gui.options.control.TickBoxControl;
 import me.jellysquid.mods.sodium.client.gui.options.storage.MinecraftOptionsStorage;
 import me.jellysquid.mods.sodium.client.gui.options.storage.SodiumOptionsStorage;
+import me.jellysquid.mods.sodium.client.render.chunk.compile.executor.ChunkBuilder;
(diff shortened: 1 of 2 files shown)


ghost:
I don't think this is needed, it's not a system any significant amount of users would encounter and if you're trying to run the game on heap sizes low enough to make it choke on higher thread counts (and therefore, for this logic to take effect) it's likely you know what you're doing anyway.

douira:
lgtm

embeddedt:
> if you're trying to run the game on heap sizes low enough to make it choke on higher thread counts (and therefore, for this logic to take effect) it's likely you know what you're doing anyway.

IMO it's bad practice for a mod to crash in an impossible-to-satisfy configuration, especially if that option is exposed in a way that's easy for a user to modify, and if it can easily be prevented with a small code change (as done here). It leaves a bad impression when a mod crashes even if it could be argued the user was at fault for using it wrong. I can also say from anecdotal experience that at least one person I know has experienced crashes and bad performance as a result of setting this value too high for the given heap size.

It's not going to be a perfect solution for modded (the heap could be large there but also be very full) but it's at least possible to more gracefully handle this solution for vanilla gameplay.

PR #1931, opened 2023-08-03
fix mixin config

offbeat-stuff:
The folder path did not match.


jellysquid3 (approved):

jellysquid3:
Oops. I'm glad someone caught it...

PR #1932, opened 2023-08-03
Revert to original center lightmap calculation

PepperCode1:
The current calculation matches vanilla. The vanilla calculation likely performs additional checks for inset quads so that they are not rendered too dark. However, Sodium does not need to perform these checks because it interpolates AO values for inset faces. Additionally, since the offset face is used during interpolation, the inset quad may appear darker than expected if the offset block is full opaque. The original calculation accounts for this possibility.


jellysquid3 (approved):
Makes sense to me. Code looks good.

PR #1945, opened 2023-08-05
Allow running with latest Not Enough Crashes versions

natanfudge:
NEC as of 4.4.6 doesn't show the crash screen when Sodium is installed, so the issue should be resolved. 
If there are any other issues feel free to contact me.


jellysquid3:
Does the version string need `+1.20.1`? I thought that metadata was ignored.

natanfudge:
I haven't tried it, I know +1.20.1 works well.

PR #1967, opened 2023-08-07
Miscellaneous Cleanup

PepperCode1:
- Remove `WeatherAccessor` as `Biome.Weather` is now a record
- Rename mixins to match target class name
- Rename mixins such that `Mixin` or `Accessor` affix is always a suffix
- Rename remaining occurrences of `TileEntity` to `BlockEntity`
- Fix `ChunkRenderList`'s `size` field not being incremented
- Remove `ChunkRenderList#getSize`
- Rename `ByteIterator#next` to `nextByteAsInt`


jellysquid3 (approved):

jellysquid3:
Changes look reasonable to me. The fewer mixins, the better.

PR #1972, opened 2023-08-09
Avoid sorting sections in `OcclusionCuller`

PepperCode1:
`OcclusionCuller.initOutsideWorldHeight` now uses diamond spiral iteration to avoid collecting and sorting sections before enqueuing them. See the comment above the method for more information.


jellysquid3 (approved):
Nice work. The distance-sort was a hack, and it produced an iteration order that was slightly different from a normal search. This code appears to fix that problem and is slightly faster.

Issue #1898, opened 2023-07-22
Chunks are occasionally incorrectly culled, leading to holes in the world


jellysquid3:
To be filled in.

jellysquid3:
This PR fixes the issue: #1790.

jellysquid3:
Fixed with commit https://github.com/CaffeineMC/sodium-fabric/commit/8a8aad0df3ec36d5246d6a2a6efc1d34a7e092b1.

PR #1974, opened 2023-08-10
Manually set IncomingDirections for neighbor chunks during searching

Noryea:
This commit should make fewer chunks be loaded (the number approximately equals to 0.4.x)

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
@@ -74,6 +74,7 @@ public class OcclusionCuller {
 
             if (adj != null) {
                 enqueue(queue, adj, 1 << GraphDirection.opposite(direction), frame);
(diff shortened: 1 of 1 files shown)


jellysquid3:
Why? This just reintroduces the bug we fixed with https://github.com/CaffeineMC/sodium-fabric/issues/1898. You are resetting the incoming directions for a chunk every time it is enqueued, but that breaks occlusion culling.

jellysquid3:
To explain: Each chunk stores visibility information which describes which _outgoing_ faces are visible for a given set of _incoming_ faces. When performing the search, we can encounter a situation where the same chunk is traversed into multiple times from different directions. We use `VisibilityEncoding.getOutgoingDirections` to create a bit-field union of the outgoing directions over the input set of incoming directions.

Noryea:
To be frank, I expected you to ignore this pull request(or close it without explaination), because it wasn't reasonable.
I did it just for a friend of mine, who uses a laptop with Intel HD Graphics 6000 and gets a _lower_ framerate on servers after installing Sodium 0.5.0...

Thank you for explaining the details, sorry to bother you

PR #1984, opened 2023-08-11
Improve graphics card probing on Linux

retrixe:
Discrete GPUs in hybrid graphics laptops often don't have display out ports, and `lspci` refers to them as "3D controllers" instead of "VGA compatible controllers." Oshi fails to detect these GPUs, so we replace its detection logic under Linux by enumerating PCIe devices in `/sys/bus/pci/devices` ourselves.

Since the Linux kernel doesn't provide the device names, we use `lspci`, which has a database of PCIe device names mapped to IDs, to retrieve the graphics card name. This code still functions without `lspci`, but the graphics card names will be missing.

We also try to retrieve the NVIDIA driver version through sysfs (another improvement over the previous probing logic).

Note: This doesn't cover NVIDIA's Tegra SoCs which don't have the GPU on the PCIe bus.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/probe/GraphicsAdapterVendor.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/probe/GraphicsAdapterVendor.java
@@ -10,11 +10,11 @@ public enum GraphicsAdapterVendor {
 
     @NotNull
     static GraphicsAdapterVendor identifyVendorFromString(String vendor) {
-        if (vendor.startsWith("Advanced Micro Devices, Inc.") || vendor.contains("(0x1002)")) {
+        if (vendor.startsWith("Advanced Micro Devices, Inc.") || vendor.contains("(0x1002)") || vendor.equals("0x1002")) {
             return AMD;
-        } else if (vendor.startsWith("NVIDIA") || vendor.contains("(0x10de)")) {
+        } else if (vendor.startsWith("NVIDIA") || vendor.contains("(0x10de)") || vendor.equals("0x10de")) {
             return NVIDIA;
-        } else if (vendor.startsWith("Intel") || vendor.contains("(0x8086)")) {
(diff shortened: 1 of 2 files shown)


src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/probe/GraphicsAdapterVendor.java
```diff
@@ -10,11 +10,11 @@ public enum GraphicsAdapterVendor {
...
     @NotNull
     static GraphicsAdapterVendor identifyVendorFromString(String vendor) {
-        if (vendor.startsWith("Advanced Micro Devices, Inc.") || vendor.contains("(0x1002)")) {
+        if (vendor.startsWith("Advanced Micro Devices, Inc.") || vendor.contains("(0x1002)") || vendor.equals("0x1002")) {
             return AMD;
-        } else if (vendor.startsWith("NVIDIA") || vendor.contains("(0x10de)")) {
+        } else if (vendor.startsWith("NVIDIA") || vendor.contains("(0x10de)") || vendor.equals("0x10de")) {
             return NVIDIA;
-        } else if (vendor.startsWith("Intel") || vendor.contains("(0x8086)")) {
+        } else if (vendor.startsWith("Intel") || vendor.contains("(0x8086)") || vendor.equals("0x8086")) {
```
Octol1ttle: Is there a reason this can't be replaced with `vendor.contains("0xDRVR")`? e.g. `if (vendor.startsWith("Advanced Micro Devices, Inc.") || vendor.contains("0x1002"))`

retrixe: I originally didn't replace it to be on the safer side, but I suppose it's a safe assumption that the vendor ID won't be part of some longer unrelated string

jellysquid3 (approved):
Seems reasonable to me... Well, your implementation, I mean. The situation itself is ridiculous.

Issue #1977, opened 2023-08-10
Issue templates should forbid the use of third-party paste services


jellysquid3:
Paste services contribute to significant "rot" on the issue tracker, for two reasons:

- Most paste services will automatically delete uploads after some time, and even when they provide an option to keep the upload forever, most users do not enable it.

- Paste services are (generally) not very reliable. Everyone has their favorites, most of them are hosted by unknown people, and few of them remain online for more than a few years.

When either of these happen, we lose important information that often cannot be obtained again. This unfortunately leads to issues being closed on the tracker because nobody has the crash reports any longer to debug it.

We should find a way to forbid URLs being used in the "Crash Report" and "Log File" section of the issue template to stop these problems. It might be considered a minor inconvenience, but you can already drag-and-drop the files into the template, and achieve the same effect (a hyperlink to an attachment hosted in our issue tracker.)

Maybe it is possible to use the issue template's "validation" parameters with a Regex expression or something to achieve this.

jellysquid3:
One small addendum: We should probably also take this time to prevent users from directly pasting the text into the issue template. It never renders correctly and it leads to a silly issue where hyperlinks are created to _other_ issues (see the issue history of #1). We want people to drag-and-drop the files in, not literally copy the text.

Maybe this can be achieved by forcing the Regex to match against any HTTPS URL, but only if the right domain is used?

Phoenix-Starlight:
As of this moment, issue templates only have the key `required` for `validations` meaning that there is no way to match/filter contents. We could look at workflows or consider creating a bot to do this.

PR #1994, opened 2023-08-12
Automatically close bug report issues that don't have valid log files

AstreaTSS:
This PR introduces a script that validates/verifies bug report issues to make sure they have a log file as a GitHub attachment. Any invalid issues will be closed with a comment explaining what they need to do - if they go back and fix the issues, then it'll automatically be re-opened.

This is achieved through a regex and a glorified Javascript... script that runs any time an issue is opened, edited, or reopened. This is a bit hacky (as with almost everything GitHub Actions), so please feel free to point out any adjustments you want to make. This PR can be tested on my fork just by creating an issue, so feel free to use it.

Special note is the closing/opening messages, which may need changing.

(This may or may not require making it so actions can read and edit things, as well as work with PRs - I'm not sure. Just note that I tested with everything enabled.)


douira:
I've tested this too and it works as expected.

douira:
Indirectly fixes #1977

PR #1997, opened 2023-08-13
Immutable gl buffer arena + fix for maxElements

MCRcortex:
Replaces the mutable buffers in GlBufferArena with immutable buffers
Inline computation of max element index counts

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
@@ -5,6 +5,8 @@ import org.lwjgl.opengl.GL30C;
 import org.lwjgl.opengl.GL44C;
 
 public enum GlBufferStorageFlags implements EnumBit {
+    NONE(0),
+
     PERSISTENT(GL44C.GL_MAP_PERSISTENT_BIT),
     MAP_READ(GL30C.GL_MAP_READ_BIT),
     MAP_WRITE(GL30C.GL_MAP_WRITE_BIT),
--- a/src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
@@ -5,8 +5,6 @@ import org.lwjgl.opengl.GL30C;
 import org.lwjgl.opengl.GL44C;
 
 public enum GlBufferStorageFlags implements EnumBit {
-    NONE(0),
-
     PERSISTENT(GL44C.GL_MAP_PERSISTENT_BIT),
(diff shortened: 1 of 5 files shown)


src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
```diff
@@ -5,6 +5,8 @@
 import org.lwjgl.opengl.GL44C;
 
 public enum GlBufferStorageFlags implements EnumBit {
+    NONE(0),
```
jellysquid3: This is a bit field... just create an empty set, rather than defining NONE here.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
```diff
@@ -138,15 +138,18 @@ private static void addDrawCommands(MultiDrawBatch batch, long pMeshData, int ma
         final var pElementCount = batch.pElementCount;
 
         int size = batch.size;
-
+        int maxIndex = batch.maxElements;
```
jellysquid3: Why did you inline this?

jellysquid3 (changes requested):

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
MCRcortex: This was inlined as it was showing a high amount of usage in the profiler, was probably lied to but this should still be faster

src/main/java/me/jellysquid/mods/sodium/client/gl/buffer/GlBufferStorageFlags.java
MCRcortex: Fixed, you cannot just create an empty bitfield as you have to pass in an element to the underlying EnumSet so that it can obtain the class

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
jellysquid3: I don't believe this is faster. The original code is very simple and only accounts for a very small part of the runtime (<0.5%) according to YourKit Profiler...

It also doesn't behave the same way. Previously, only visible slices would be considered when calculating the maximum element index, but this considers every slice.

MCRcortex: good point i will revert this

jellysquid3:
This PR still cannot be accepted... `glBufferStorage` (as used when creating an immutable buffer) requires OpenGL 4.4 or ARB_buffer_storage. We will have to wait until the next major version before we start breaking OpenGL 3.x support.

PR #2005, opened 2023-08-14
Add donation and other links to manifest

Madis0:
Compatible with [Mod Menu](https://github.com/TerraformersMC/ModMenu/wiki/API#links) and potentially anything else that might want to read it.

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -25,9 +25,11 @@
   },
   "custom": {
     "fabric-renderer-api-v1:contains_renderer": true,
-    "links": {
+    "modmenu": {
+      "links": {
         "modmenu.donate": "https://jellysquid.me/donate"
       }
+    }
   },
   "accessWidener": "sodium.accesswidener",
   "mixins": [
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -27,10 +27,10 @@
     "fabric-renderer-api-v1:contains_renderer": true,
     "modmenu": {
       "links": {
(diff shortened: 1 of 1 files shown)


douira:
Is this sufficient for FO to stop hiding Sodium's donation button by default?

Madis0:
No, this is supplementary for any modpack or user who might hide the button, or any user who is looking to find the link from Mod Menu in the first place. I'll make a separate PR for FO concern.

src/main/resources/fabric.mod.json
```diff
@@ -24,7 +24,16 @@
     ]
   },
   "custom": {
-    "fabric-renderer-api-v1:contains_renderer": true
+    "fabric-renderer-api-v1:contains_renderer": true,
+    "modmenu": {
+      "links": {
+        "modmenu.donate": "https://jellysquid.me/donate",
```
jellysquid3: Should be `https://caffeinemc.net/donate` even if it just redirects currently. We may change it at some point.

src/main/resources/fabric.mod.json
```diff
@@ -24,7 +24,16 @@
...
+        "modmenu.discord": "https://caffeinemc.net/discord",
+        "modmenu.github_releases": "https://github.com/CaffeineMC/sodium-fabric/releases",
+        "modmenu.kofi": "https://ko-fi.com/jellysquid_",
```
jellysquid3: Shouldn't directly link to Ko-Fi, see prior comment.

jellysquid3 (changes requested):

jellysquid3 (approved):

Issue #1863, opened 2023-07-03
If my pc is just weak for sodium?


AlexCigarettaGeza:
Sodium freezes when 1 start but one times it's loads prefectly fine!

There is my latest log here:
[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/11941374/latest.log)

pajicadvance:
Your Intel GPU drivers are outdated and won't work with Sodium (see #899). If you have a dedicated GPU, then the game is not using it. Also, hacked clients are not supported.

AlexCigarettaGeza:
> Your Intel GPU drivers are outdated and won't work with Sodium (see #899). If you have a dedicated GPU, then the game is not using it. Also, hacked clients are not supported.

My hacked client is disabled i crash when i use it. But damn in my country GPU drivers more than 100k money i only make 10k in a month.  But thanks

AlexCigarettaGeza:
> > Your Intel GPU drivers are outdated and won't work with Sodium (see #899). If you have a dedicated GPU, then the game is not using it. Also, hacked clients are not supported.
> 
> My hacked client is disabled i crash when i use it. But damn in my country GPU drivers more than 100k money i only make 10k in a month. But thanks

This is how to lose a 400 days hardcore minecraft world.

jellysquid3:
Then make a backup of your world if you are worried about losing it. The problem is not Sodium, it's your currently installed graphics drivers. Upgrading your graphics drivers **does not** require any hardware modifications and can be done for free. It's just a software update, and it comes from Intel (the official manufacturer) itself.

AlexCigarettaGeza:
> It's says my videocard is already upgraded to the best

Issue #1817, opened 2023-06-16
Game stuck on starting screen with Sodium only


GodProZEE:
### Version information

mc1.20-0.4.10+build.27

### Reproduction Steps

1. Start the game
2. The game doesn't get past the loading screen which says "Mojang".

### Crash Report file

https://mclo.gs/ryfEHCj

### Additional information

I use Sodium, Lithium, and Starlight, and after removing everything but Sodium, the error is still there so it's Sodium causing this issue.

pajicadvance:
Duplicate of #899 . Update your Intel GPU drivers.

Issue #1710, opened 2023-04-30
stop loading in the beginning 


Mufi432:
### Version information

1.19.4

### Reproduction Steps

1. Restart my computer
2. Reinstall sodium
3. Disable other mods

Not worked

### Crash Report file

Here is the errors and warns

[13:58:01] [Render thread/WARN]: Assets URL 'quilt.zfs://transformed-mod-minecraft.i0:0/assets/.mcassetsroot' uses unexpected schema

[13:58:01] [Render thread/WARN]: Assets URL 'quilt.zfs://transformed-mod-minecraft.i0:0/data/.mcassetsroot' uses unexpected schema

[13:58:02] [Render thread/ERROR]: Failed to verify authentication

[13:58:08] [Render thread/WARN]: Static binding violation: PRIVATE @Overwrite method method_23182 in #sodium:sodium.mixins.json: features.item.MixinItemRenderer from mod sodium cannot reduce visibiliy of PUBLIC target method, visibility will be upgraded.

pajicadvance:
Upload the full log please.

Mufi432:
[minecraft-exported-logs-2023-04-30T14-51-30.log](https://github.com/CaffeineMC/sodium-fabric/files/11361340/minecraft-exported-logs-2023-04-30T14-51-30.log)

pajicadvance:
Seems to be a duplicate of #899. Try updating your graphics driver from the link provided in the mentioned issue report.

Mufi432:
thanks for help

jellysquid3:
Not an issue with Sodium. You probably just need to update your graphics drivers as described above.

Issue #899, opened 2021-09-12
Game freezes during startup with older Intel HD Graphics drivers


jellysquid3:
### **See [this comment](https://github.com/CaffeineMC/sodium-fabric/issues/899#issuecomment-1025203807) for a solution to the problem.**

---

There is no crash log or debugging information associated with this problem, aside from the following hardware information: 
[forced-crash.txt](https://github.com/CaffeineMC/sodium-fabric/files/7149883/forced-crash.txt)

Following a bisect, the following commit was revealed to be causing the problem: https://github.com/CaffeineMC/sodium-fabric/commit/bb33a03d705e7bf7a845445befb9a43b39bcca03

My guess is that our fencing logic in MixinMinecraftClient is causing the hang, as it would be the only thing here running early enough that could cause such a problem. It seems that this is dependent on hardware, as we've only seen older Intel cards hang in this manner.

LucilleTea:
Had another report of this on the same hardware with a slightly newer driver version (10.18.10.4358).
Updating to the latest available driver, 15.33.53.5161, reportedly fixed the problem.
[forced crash.txt](https://github.com/CaffeineMC/sodium-fabric/files/7168148/forced.crash.txt)

jellysquid3:
This almost certainly seems like a driver bug with 10.18.10.4358, which affects those with an Baytrail or 3rd generation Intel graphics. [Another user had the affected driver](https://discord.com/channels/602796788608401408/896802064506699846/896811654170439691), and upgrading to 15.33.53.5161 (the latest available) fixed it.

jellysquid3:
Possibly related: https://github.com/CaffeineMC/sodium-fabric/commit/ac0bec6e64697ade618aac2640ece9735bb00790

RockStone-God:
I have intel HD 4000 graphics, could that be an issue for it hanging on loading screen?

jellysquid3:
In every case we've seen, this is caused by out-of-date graphics drivers on Intel _Bay Trail_ and _Ivy Bridge_ CPUs, and is fixed by upgrading the driver to version 15.33.53.5161 (found [here](https://www.intel.com/content/www/us/en/download/18606/intel-graphics-driver-for-windows-15-33.html)).

<details>
<summary>
List of known affected Intel processors
</summary>

    Intel® Celeron® 1000M 
    Intel® Celeron® 1005M 
    Intel® Celeron® 1007U 
    Intel® Celeron® 1017U 
[...]

PR #2022, opened 2023-08-18
Fix incorrect Intel broken driver check

pajicadvance:
This changes the pre-launch check for the broken Intel drivers to include every Intel driver version that starts with `10.18.10`.
Blacklisting only one minor version is not enough, as there are several of them which are confirmed broken:

`10.18.10.4358` - https://github.com/CaffeineMC/sodium-fabric/issues/899
`10.18.10.4242` - https://github.com/CaffeineMC/sodium-fabric/issues/1710
`10.18.10.4885` - https://github.com/CaffeineMC/sodium-fabric/issues/1817
`10.18.10.4425` - https://github.com/CaffeineMC/sodium-fabric/issues/1863
`10.18.10.4653` - Support thread on the Discord server

As such, it would most likely be safe to assume that all minor versions of this driver are broken in the same manner. We direct users to download version `15.33.53.5161` to fix the problem, anyways.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/PreLaunchChecks.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/PreLaunchChecks.java
@@ -52,7 +52,7 @@ public class PreLaunchChecks {
             try {
                 var version = WindowsDriverStoreVersion.parse(adapter.version());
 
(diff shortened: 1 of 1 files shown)


jellysquid3 (approved):

jellysquid3:
Thanks. I think you're right, the old version check had a typo. Unfortunately it seems the range of broken drivers is larger than we initially thought too. This solution should work though.

PR #2037, opened 2023-08-23
Changes to make hotswap work

Altirix:
There are 4 mixins that can cause hotswap to fail
```
features.render.immediate.buffer_builder.sorting.BufferBuilderMixin -- fails due to enhanced switch statement
features.world.biome.FoliageColorsAccessor -- errors with javaagent
features.world.biome.GrassColorsAccessor -- errors with javaagent
features.render.world.clouds.BackgroundRendererInvoker  -- errors with javaagent
```
Using either an if-else or a switch (as in https://github.com/CaffeineMC/lithium-fabric/pull/41) on the value allows hotswap with the `BufferBuilderMixin`

it also seems something makes the jvm unhappy about static accessors and invokers, i dont know why but hopefully someone who does can work that out so these access wideners can be replaced without the loss of hotswap. 

i did find some mentions of people having hotswap breaking with static accessors in the past. 

(fabric discord) https://discord.com/channels/507304429255393322/507982478276034570/742063021085687898. 

however, adding the javaagent helped in their case but upon trying this with the changes to `BufferBuilderMixin` it would work but the 2 static accessors and 1 static invoker would not be reloaded.


jellysquid3:
Thanks for your work on this, and sorry for the delay. Your changes seem to solve the problem on my machine, and they make things much less insufferable. Hopefully these bugs can be solved without workarounds in the future.

jellysquid3:
Merged manually with https://github.com/CaffeineMC/sodium-fabric/commit/665a3c18b07df86fb7f6d1490f66124848475c4b.

PR #2038, opened 2023-08-23
Switch render order of block aligned and unaligned block faces 

Altirix:
This results in the internal cube for slime and honey blocks being rendered more correctly.
262338598-f1d33679-49a5-4644-a8ac-13339b5c1e25

it seems to render fine but does this negatively affect performance if those faces are done before axis-aligned ones?

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
@@ -68,18 +68,19 @@ public class BlockRenderer {
             renderOffset = Vec3d.ZERO;
         }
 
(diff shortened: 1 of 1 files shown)


douira:
I don't think it would affect performance beyond what rendering translucent faces in the right order to begin with would cost.

douira (commented):
I wonder how this actually makes a difference because I think the order in which the quads are eventually rendered depends on the order of the enum and how it's then used in https://github.com/CaffeineMC/sodium-fabric/blob/dev/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java#L136

jellysquid3:
This PR does not make sense to me. The render order of geometry is determined by the renderer and is independent of the meshing process. Unless the slime/honey blocks for some reason just don't have their geometry tagged as sided for some reason (which would be a different bug) this PR should not fix anything.

PR #2063, opened 2023-09-04
Add Fabric BlockView API v2 support

PepperCode1:
- Modify `BiomeSlice` to store and use `RegistryEntry<Biome>` instead of `Biome`
- Rename "block entity attachment" to "block entity render data" to match terminology of BlockView API v2
- Trim BlockEntity and BlockEntity render data maps

`WorldSlice` still implements the deprecated `RenderAttachedBlockView` interface and Rendering Data Attachment v1 is still JiJed so that Rendering Data Attachment v1 is still supported, since `RenderAttachedBlockView` is only automatically implemented on instances of `WorldView`, which `WorldSlice` does not implement. `WorldSlice` can stop implementing `RenderAttachedBlockView` and the module dependency can be dropped once most mod developers have migrated to Fabric BlockView API v2.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,8 +82,12 @@ dependencies {
 
     // Fabric API
     modIncludeImplementation(fabricApi.module("fabric-api-base", project.fabric_version))
+    modIncludeImplementation(fabricApi.module("fabric-block-view-api-v2", project.fabric_version))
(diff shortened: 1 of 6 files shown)


IMS212 (approved):
Looks good to me.

PR #2066, opened 2023-09-08
Fix block updates being ignored before initial build completes

embeddedt:
If the following sequence of events takes place:

* A chunk section is loaded and scheduled for an initial build (including creation of the world snapshot).
* A block update, etc. attempts to schedule the section for re-rendering.
* The initial build of the section completes.

it appears that the section is never actually built a second time with the new information. This causes problems with mods like AE2, which render disconnected cables initially and then connect them with a block update later. When Sodium is installed, and with a PC of the correct speed, it's possible to reliably reproduce behavior in which the cables remain visually disconnected (despite their hitbox changing) until F3+A is run manually. I was also able to make the reproduction more reliable by artificially adding a 2 second delay to chunk meshing tasks.

After several hours of tinkering and analysis, I believe I've found the culprit: the section manager explicitly avoids scheduling render updates for chunks which are not marked as built. This behavior seems incorrect; we need to process block updates as soon as the section has been added to our tracker and a snapshot of its data has been taken.

I've thus changed the logic to always schedule the block update even for non-null sections. On my system this appears to have fixed the issue, but I would like someone more familiar with the chunk tracker to validate that this does not violate some other invariant.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RenderSectionManager.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RenderSectionManager.java
@@ -454,7 +454,7 @@ public class RenderSectionManager {
                 this.needsUpdate = true;
 
                 CancellationToken token = section.getBuildCancellationToken();
-                if(token != null) {
+                if (token != null) {
                     token.setCancelled();
                     section.setBuildCancellationToken(null); // to allow queuing it again
                 }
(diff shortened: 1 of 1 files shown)


embeddedt:
Nevermind, this still doesn't seem to work 100% of the time... but I do think the logic error is real.

embeddedt:
I've created a fix that actually works and updated this to the latest `dev` branch. For posterity, here is the test case I made to reproduce the issue:
* Add https://modrinth.com/mod/forcecloseworldloadingscreen/version/2.1.1 to `run/mods`. (Needed to allow in-game GUI to display before chunks load)
* Apply [this diff](https://cdn.discordapp.com/attachments/651120262129123330/1150496991902695484/patch.diff) to the current `dev` branch to add ability to stall any chunk updates in progress until the U key is pressed.
* Launch the dev client and ensure always defer chunk updates is enabled.
* **Create a fresh superflat world** in creative mode. (It MUST be a new world, reloading an existing world causes the issue to not happen for some reason.)
* Wait for the in-game GUI to display. No chunks will be visible yet, as chunk updates are suppressed.
* Open the creative inventory, get stone, look down, and place it in front of you (I placed it two blocks away from me, not sure if that's important).
* Hold the U key until chunks start rendering.
* Observe that the stone does not render.
* Press F3+A.
* Observe that the stone now renders.

With the change in this PR, the stone renders after pressing U, without requiring an extra refresh of the chunks.

PR #2086, opened 2023-09-23
Move particle vertex processing to the GPU

wahfl2:
Currently, all particle vertices are processed and rotated on the CPU, before uploading them to a shader to be rendered. Sodium optimizes this slightly with more direct and explicit math operations for more optimization opportunities. However, a lot of vector math is still being done on the CPU. 

This PR moves that processing (for most particles) to the GPU in a new shader. Specifically, this PR optimizes billboard particles that haven't overridden the buildGeometry method, which means they face the camera. This is the vast majority of particles, with only a few overrides in vanilla Minecraft (shrieker, elder guardian, etc). The normal particle rendering is still in place for the particles that are built some other way.


umbrageodotus:
@douira plans for this? Or is async occlusion first? I can review and rebase the PR if needed.

douira:
If you or the original author of this PR can rebase it and demonstrate a performance improvement I can consider it for review.

PR #2116, opened 2023-10-10
Reintroduce optional full-precision vertex format

### LINKED ISSUE #2004: Modification of data precision should be optional
https://github.com/CaffeineMC/sodium-fabric/blob/0a096a73ca962b27e1500a109d082d2df2481790/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/impl/CompactChunkVertex.java#L32-L41

## Issue
The precision of the modified data is too low, the color has no alpha anymore, and the light only goes from 0-15.
I can see its value. but sometimes other mods require the original precision range to do additional things.
For example, Shimmer will make fragments bloom by checking lights value bigger than 240. There is no way to make it compatible with sodium with the current modification. 

## Solution
I agree that this is a good improvement and will significantly reduce the overhead of data upload to the gpu. But can we make it optional? 

IMP, at least the reduced precision of the uv will have a significant impact on the ultra-high resolution texture resourcepack.
remove alpha really shouldn't be, it's risky.
besides, I'm afraid of z-fighting issue by the precision of the position.

embeddedt:
This PR fixes #2004 and the remaining compact vertex format issues on the tracker by reintroducing a full-precision vertex format, similar to what was available in Sodium 0.2.0. To do this some fairly substantial changes had to be made to the shader binding code, since the old logic had hardcoded attribute/tessellation bindings.

Notably, the new format:

* Uses full 32-bit single-precision floats for vertex positions.
* Uses full 8-bit coordinates for lighting, as the compression trick used by Sodium 0.5 causes artifacts in some cases.

There is likely cleanup that can be done to the code, but this seems to be completely functional and stable in its current state.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/gl/tessellation/TessellationBinding.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/gl/tessellation/TessellationBinding.java
@@ -4,10 +4,13 @@ import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexAttributeBinding;
 import me.jellysquid.mods.sodium.client.gl.buffer.GlBuffer;
 import me.jellysquid.mods.sodium.client.gl.buffer.GlBufferTarget;
 
+import java.util.Objects;
+
 public record TessellationBinding(GlBufferTarget target,
                                   GlBuffer buffer,
                                   GlVertexAttributeBinding[] attributeBindings) {
     public static TessellationBinding forVertexBuffer(GlBuffer buffer, GlVertexAttributeBinding[] attributes) {
+        Objects.requireNonNull(attributes);
         return new TessellationBinding(GlBufferTarget.ARRAY_BUFFER, buffer, attributes);
     }
 
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
@@ -23,7 +23,7 @@ import me.jellysquid.mods.sodium.client.render.chunk.terrain.TerrainRenderPass;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkMeshAttribute;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkMeshFormats;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkVertexType;
-import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.VanillaLikeChunkMeshAttribute;
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.StandardChunkMeshAttribute;
 import me.jellysquid.mods.sodium.client.render.viewport.CameraTransform;
 import me.jellysquid.mods.sodium.client.util.BitwiseMath;
 import org.lwjgl.system.MemoryUtil;
@@ -232,16 +232,16 @@ public class DefaultChunkRenderer extends ShaderChunkRenderer {
                             compactFormat.getAttribute(ChunkMeshAttribute.VERTEX_DATA))
             };
         } else if (this.vertexType == ChunkMeshFormats.VANILLA_LIKE) {
-            GlVertexFormat<VanillaLikeChunkMeshAttribute> vanillaFormat = (GlVertexFormat<VanillaLikeChunkMeshAttribute>)this.vertexFormat;
+            GlVertexFormat<StandardChunkMeshAttribute> vanillaFormat = (GlVertexFormat<StandardChunkMeshAttribute>)this.vertexFormat;
             return new GlVertexAttributeBinding[] {
                     new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_POSITION,
-                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.POSITION)),
+                            vanillaFormat.getAttribute(StandardChunkMeshAttribute.POSITION)),
                     new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_COLOR,
-                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.COLOR)),
+                            vanillaFormat.getAttribute(StandardChunkMeshAttribute.COLOR)),
                     new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_TEXTURE_UV,
-                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.TEXTURE_UV)),
+                            vanillaFormat.getAttribute(StandardChunkMeshAttribute.TEXTURE_UV)),
                     new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_DRAW_PARAMS_LIGHT,
-                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.DRAW_PARAMS_LIGHT)),
+                            vanillaFormat.getAttribute(StandardChunkMeshAttribute.DRAW_PARAMS_LIGHT)),
             };
         } else {
             return null; // assume Oculus/Iris will take over
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
@@ -2,6 +2,7 @@ package me.jellysquid.mods.sodium.client.render.chunk;
 
 import me.jellysquid.mods.sodium.client.SodiumClientMod;
 import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexAttributeBinding;
+import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexFormat;
 import me.jellysquid.mods.sodium.client.gl.device.CommandList;
 import me.jellysquid.mods.sodium.client.gl.device.DrawCommandList;
 import me.jellysquid.mods.sodium.client.gl.device.MultiDrawBatch;
@@ -20,7 +21,9 @@ import me.jellysquid.mods.sodium.client.render.chunk.shader.ChunkShaderBindingPo
 import me.jellysquid.mods.sodium.client.render.chunk.shader.ChunkShaderInterface;
 import me.jellysquid.mods.sodium.client.render.chunk.terrain.TerrainRenderPass;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkMeshAttribute;
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkMeshFormats;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkVertexType;
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.VanillaLikeChunkMeshAttribute;
 import me.jellysquid.mods.sodium.client.render.viewport.CameraTransform;
 import me.jellysquid.mods.sodium.client.util.BitwiseMath;
 import org.lwjgl.system.MemoryUtil;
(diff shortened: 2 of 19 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/ChunkVertexType.java
```diff
@@ -3,7 +3,11 @@
 import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexFormat;
 
 public interface ChunkVertexType {
-    GlVertexFormat<ChunkMeshAttribute> getVertexFormat();
+    GlVertexFormat<?> getVertexFormat();
 
     ChunkVertexEncoder getEncoder();
+
+    default String getDefine() {
+        return "VERTEX_FORMAT_COMPACT"; // as in original Sodium all vertex formats would use this mode
```
douira, line 11: what is this comment saying?

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
```diff
@@ -217,12 +224,32 @@ private GlTessellation prepareTessellation(CommandList commandList, RenderRegion
...
+                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.POSITION)),
+                    new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_COLOR,
+                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.COLOR)),
+                    new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_TEXTURE_UV,
+                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.TEXTURE_UV)),
+                    new GlVertexAttributeBinding(ChunkShaderBindingPoints.ATTRIBUTE_DRAW_PARAMS_LIGHT,
+                            vanillaFormat.getAttribute(VanillaLikeChunkMeshAttribute.DRAW_PARAMS_LIGHT)),
+            };
+        } else
+            return null; // assume Oculus/Iris will take over
```
douira, line 247: this section needs correct formatting: spaces between `if` and `(`, curly braces for the else block

src/main/java/me/jellysquid/mods/sodium/client/gl/tessellation/TessellationBinding.java
```diff
@@ -4,10 +4,13 @@
 import me.jellysquid.mods.sodium.client.gl.buffer.GlBuffer;
 import me.jellysquid.mods.sodium.client.gl.buffer.GlBufferTarget;
 
+import java.util.Objects;
+
 public record TessellationBinding(GlBufferTarget target,
                                   GlBuffer buffer,
                                   GlVertexAttributeBinding[] attributeBindings) {
     public static TessellationBinding forVertexBuffer(GlBuffer buffer, GlVertexAttributeBinding[] attributes) {
+        Objects.requireNonNull(attributes);
```
douira, line 13: why does this exist?

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ShaderChunkRenderer.java
```diff
@@ -39,6 +39,19 @@ protected GlProgram<ChunkShaderInterface> compileProgram(ChunkShaderOptions opti
...
+        if (this.vertexType == ChunkMeshFormats.COMPACT) {
+            return builder.bindAttribute("in_VertexData", ChunkShaderBindingPoints.ATTRIBUTE_PACKED_DATA);
+        } else if (this.vertexType == ChunkMeshFormats.VANILLA_LIKE) {
+            return builder
+                    .bindAttribute("in_Pos", ChunkShaderBindingPoints.ATTRIBUTE_POSITION)
+                    .bindAttribute("in_Color", ChunkShaderBindingPoints.ATTRIBUTE_COLOR)
+                    .bindAttribute("in_TextureUv", ChunkShaderBindingPoints.ATTRIBUTE_TEXTURE_UV)
+                    .bindAttribute("in_DrawParamsLight", ChunkShaderBindingPoints.ATTRIBUTE_DRAW_PARAMS_LIGHT);
+        } else
+            throw new IllegalArgumentException("Unexpected vertex type");
```
douira, line 52: missing curly braces

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/RenderSectionManager.java
```diff
@@ -82,10 +85,14 @@ public class RenderSectionManager {
     private @Nullable BlockPos lastCameraPosition;
 
     public RenderSectionManager(ClientWorld world, int renderDistance, CommandList commandList) {
-        this.chunkRenderer = new DefaultChunkRenderer(RenderDevice.INSTANCE, ChunkMeshFormats.COMPACT);
+        ChunkVertexType vertexType = SodiumClientMod.options().performance.useCompactVertexFormat ? ChunkMeshFormats.COMPACT : ChunkMeshFormats.VANILLA_LIKE;
```
douira, line 88: there's two places where the option is read and a vertex format selected. Could this maybe be put into some mechanism that supplies a `ChunkMeshFormat` based on the option and then we just read its data here and in `RenderRegion`? Also, there's at least three places I think where there's a case decision on each of the vertex formats. Maybe that could be reduced? I don't know how feasible that is and given that there probably won't be a bunch more vertex formats, it's unlikely to be very important, but it just feels kinda fragile.

src/main/resources/assets/sodium/shaders/blocks/block_layer_opaque.vsh
```diff
@@ -20,9 +20,19 @@ uniform vec3 u_RegionOffset;
 
 uniform sampler2D u_LightTex; // The light map texture
 
+#if defined(VERTEX_FORMAT_COMPACT)
+// expect 0-15 as inputs, Sodium 0.5 logic
 vec3 _sample_lightmap(uvec2 coord) {
     return texelFetch(u_LightTex, ivec2(coord), 0).rgb;
 }
+#elif defined(VERTEX_FORMAT_FULL)
```
douira, line 28: I think this could use `#ifdef` but I'm not sure, doesn't really make a difference.

douira (commented):
Haven't tested it myself but the approach generally looks good to me. I just noticed a few code format issues and commented on the approach of how the option state is used to select a vertex format in multiple places. Thanks for the work!

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/ChunkVertexType.java
embeddedt: We need to allow vertex formats to specify a define so that the shader logic can be adjusted accordingly. However, we must also provide a default implementation, as otherwise existing mods (~~cough Iris, if I recall correctly~~) will crash with an abstract method error. We choose `VERTEX_FORMAT_COMPACT`, as any mods that were designed for an older version of Sodium expect that Sodium's shaders will be reading the vertex data in compact format.

src/main/java/me/jellysquid/mods/sodium/client/gl/tessellation/TessellationBinding.java
embeddedt: To ensure the `null` return value from `DefaultChunkRenderer#getBindingsForType` is caught if for some reason a mixin does not replace it. Returning null there is a dirty hack to ensure binary compatibility with one of the Iris mixins; we could restructure all of this if we're fine breaking addons (I prefer not to, hence the slightly weird choices).

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/ChunkVertexType.java
douira: That makes sense but it's not really clear from the comment as it stands.

src/main/java/me/jellysquid/mods/sodium/client/gl/tessellation/TessellationBinding.java
douira: maybe this possible nullability could be documented

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/impl/VanillaLikeChunkVertex.java
```diff
@@ -0,0 +1,76 @@
...
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.VanillaLikeChunkMeshAttribute;
+import net.caffeinemc.mods.sodium.api.util.ColorABGR;
+import net.caffeinemc.mods.sodium.api.util.ColorU8;
+import org.lwjgl.system.MemoryUtil;
+
+/**
+ * This vertex format is less performant and uses more VRAM than {@link CompactChunkVertex}, but should be completely
+ * compatible with mods & resource packs that need high precision for models.
+ */
+public class VanillaLikeChunkVertex implements ChunkVertexType {
```
jellysquid3: This should just be called "StandardChunkVertex" or something.

src/main/resources/assets/sodium/shaders/include/chunk_vertex.glsl
```diff
@@ -27,6 +41,7 @@ uint _vert_material;
 uint _vert_mesh_id;
 
 void _vert_init() {
+#if defined(VERTEX_FORMAT_COMPACT)
```
jellysquid3: For clarity sake, `_vert_init()` should be split into `_vert_init_compact()` and `_vert_init_standard()` , then conditionally selected with compile-time macros.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/ChunkVertexType.java
```diff
@@ -3,7 +3,11 @@
 import me.jellysquid.mods.sodium.client.gl.attribute.GlVertexFormat;
 
 public interface ChunkVertexType {
-    GlVertexFormat<ChunkMeshAttribute> getVertexFormat();
+    GlVertexFormat<?> getVertexFormat();
 
     ChunkVertexEncoder getEncoder();
+
+    default String getDefine() {
```
jellysquid3, line 10: The chunk vertex type shouldn't arbitrarily define macros in the shader code. We can instead use `#if _VERTEX_FORMAT == COMPACT` in the shader to pick a branch at compile-time.

The fog shader component doesn't currently do this, but it should be fixed too (not in this PR)...

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/shader/ChunkShaderOptions.java
```diff
@@ -2,11 +2,23 @@
...
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkMeshFormats;
+import me.jellysquid.mods.sodium.client.render.chunk.vertex.format.ChunkVertexType;
+
+public record ChunkShaderOptions(ChunkFogMode fog, TerrainRenderPass pass, ChunkVertexType type) {
+    /**
+     * @deprecated Only kept for Iris/Oculus compatibility, do not use
+     */
+    @Deprecated
+    @SuppressWarnings("unused")
+    public ChunkShaderOptions(ChunkFogMode fog, TerrainRenderPass pass) {
```
jellysquid3: Internal APIs are not given compatibility guarantees, so I don't see any reason to maintain this code path. Besides, Iris is version locked to specific versions of Sodium, anyways.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/region/RenderRegion.java
```diff
@@ -192,7 +194,15 @@ public static class DeviceResources {
         private GlTessellation tessellation;
 
         public DeviceResources(CommandList commandList, StagingBuffer stagingBuffer) {
-            int stride = ChunkMeshFormats.COMPACT.getVertexFormat().getStride();
+            int stride;
+            if(SodiumClientMod.options().performance.useCompactVertexFormat) {
```
jellysquid3, line 198: See my other remark about compatibility with Iris.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/ShaderChunkRenderer.java
```diff
@@ -49,10 +63,9 @@ private GlProgram<ChunkShaderInterface> createShader(String path, ChunkShaderOpt
                 new Identifier("sodium", path + ".fsh"), constants);
 
         try {
-            return GlProgram.builder(new Identifier("sodium", "chunk_shader"))
+            return bindAttributesForType(GlProgram.builder(new Identifier("sodium", "chunk_shader"))
```
jellysquid3: I don't think this is the right pattern. `GlProgram.builder` should be pulled out into a separate variable, and `bindAttributesForType` should be invoked like such:

```java
var builder = GlProgram.builder(...); // pull out
bindAttributesForType(builder); // new code

builder.attachShader(...); // etc
```

jellysquid3 (changes requested):
I still need to think about how we will handle having different attribute types between the two vertex formats, but here is my feedback for now.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/region/RenderRegion.java
embeddedt: Not sure how to refactor this without having to add a `ChunkVertexType` parameter to a lot of methods, unless we start storing the current vertex type on `SodiumWorldRenderer`.

jellysquid3:
#2231 is relevant and may also solve the problems this pull request seeks to do.

PR #2117, opened 2023-10-10
Round lightmap coordinates in compact chunk vertex encoder

### LINKED ISSUE #2052: Smooth Lighting + Sodium 0.5.x: In some cases, blocks don't obstruct the passage of light.
### Bug Description

https://github.com/CaffeineMC/sodium-fabric/assets/68275908/87677b86-3ae9-43b6-9431-83a5559a85b6

This does not happen with Sodium 0.4.10

### Reproduction Steps

Watch the video (snow required)

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/12475356/latest.log)

### Crash Report

[crash-2023-08-30_14.04.50-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/12475361/crash-2023-08-30_14.04.50-client.txt)

embeddedt:
Vanilla and mods can generate lightmap coordinates with the bottom 4 bits nonzero. We cannot store these with perfect precision with the CVF as it only has space for the top 4 bits, however rounding is a more accurate/correct way to handle the issue than simply discarding the bits.

Fixes https://github.com/CaffeineMC/sodium-fabric/issues/2052 in my testing.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/impl/CompactChunkVertex.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/vertex/format/impl/CompactChunkVertex.java
@@ -62,6 +62,14 @@ public class CompactChunkVertex implements ChunkVertexType {
         int block = (light >> 4) & 0xF;
         int sky = (light >> 20) & 0xF;
 
+        // Some mods/parts of vanilla use the lower 4 bits, so round if needed.
+        if(((light >> 0) & 0xF) > 0x7 && block < 0xF) {
+            block++;
+        }
+        if(((light >> 16) & 0xF) > 0x7 && sky < 0xF) {
+            sky++;
+        }
+
         return ((block << 0) | (sky << 4));
     }
 


jellysquid3:
This doesn't appear to fix the linked issue in my testing. The resulting images before/after your patch are exactly the same. Maybe I'm not reproducing it exactly right?

embeddedt:
If I recall correctly, it fixes the parity difference shown in Lolothepro's videos, but it might not fix what was shown in your screenshots.

jellysquid3:
I was comparing Lolothepro's videos. The light isn't blocked by the stained glass either in trunk or with the patch, but it works correctly in vanilla.

Lolothepro:
> I was comparing Lolothepro's videos. The light isn't blocked by the stained glass either in trunk or with the patch, but it works correctly in vanilla.

No, this PR fixes the bug

Lolothepro:
ok, it's fixed just for the x and z axis, but not for the y axis
It's a regression, because it works in the y-axis in Sodium 0.5.3

jellysquid3:
I'm closing this pull request since it doesn't fully solve the problem, and only seems to "nudge" it in the right direction some of the time. Simply moving the problem around isn't really an acceptable solution to me.

We need to revert the 16-byte vertex format and go back to using 8-bit light values to properly fix the issue. It was an incorrect assumption that vanilla and other mods did not need it.

PR #2121, opened 2023-10-12
Fix parity issue in Sodium's getFacing optimization

embeddedt:
The optimization introduced in https://github.com/CaffeineMC/sodium-fabric/pull/1587 has a subtle parity issue: if `x, y, z` are all zero, Sodium's `Direction.getFacing` implementation returns `Direction.DOWN`, while vanilla returns `Direction.NORTH`. This could potentially cause modded code to crash if it implicitly assumed that the returned direction was horizontal in this case (I believe this caused https://github.com/AlexModGuy/Ice_and_Fire/issues/4767).

This PR fixes that by checking for this corner case and returning the same value as vanilla. An alternate solution would be to reorder the chain of if statements such that NORTH would end up being selected.


douira:
Doing this check explicitly feels cleaner than manipulating the other statements, also because it avoids needing to calculate the absolute values in that case. Though I suspect either way doesn't have any performance relevance.

PR #2131, opened 2023-10-21
Fix biome colors being retrieved differently from vanilla

embeddedt:
This PR fixes a parity issue with how biome colors are retrieved compared to vanilla. Vanilla uses the lambdas in the `static final` fields of `BiomeColors` in order to call the appropriate methods on the biome, while Sodium just calls them directly. This causes compatibility issues with mods that use an AT/AW and replace the lambda in those fields to add custom logic.

In the case where no mod has replaced the lambda, the behavior of this patch should be exactly the same as Sodium 0.5.3, as the lambda delegates to the same method Sodium would have called.


jellysquid3 (approved):
Seems reasonable enough.

PR #2133, opened 2023-10-21
Sort render regions by distance from camera

### LINKED ISSUE #2132: Render regions are not ordered correctly in the translucent pass
### Bug Description

Translucent blocks at region borders are rendered in the wrong order sometimes, resulting in some of them being invisible.

This issue is *distinct* from the existing issues with translucency sorting. Those issues affect faces being sorted within chunks, while this issue appears to be directly correlated to the order in which the larger 8x4x8 groups of chunks are rendered.

### Reproduction Steps

1. Create a world in Minecraft 1.20.2, Sodium 0.5.3, using seed `-8576276202497619228`.
2. Ensure you are in creative mode and flying, and run the following command: `/tp @s -1823.931 63.25615 -1535.557 109.7 24.5`
3. Move the camera left and right slightly, and observe that a face on the ice disappears when the camera is far enough to the right. It may also be useful to add a piece of stained glass for visualization, although there will be some garbled quads on part of it due to no translucent face sorting being implemented.

Correct rendering:

Incorrect rendering:

Notice that the left face on the ice has disappeared, and more of the chunk is invisible behind the glass at one angle than the other. A similar effect has occured on the upper-right portion of the glass block. The face that disappears is precisely the face that borders the region the player is standing in.

I am 99% sure this is related to the order in which the regions themselves are drawn, as adding the following hack inside `RegionChunkRenderer#render` (just for the translucent pass) fixes the issue. Presumably the approach of computing the list of regions based on the data from the occlusion search results in the wrong order being chosen sometimes.

```java
            Map<RenderRegion, List<RenderSection>> map = new Reference2ReferenceOpenHashMap<>();
            List<RenderRegion> regions = new ArrayList<>();
            for(Map.Entry<RenderRegion, List<RenderSection>> entry : sortedRegions(list, true)) {
                regions.add(entry.getKey());
                map.put(entry.getKey(), entry.getValue());
            }

            regions.sort(Comparator.<RenderRegion>comparingDouble(region -> {
                float dx = region.getOriginX() - camera.posX;
                float dy = region.getOriginY() - camera.posY;
                float dz = region.getOriginZ() - camera.posZ;
                return (dx * dx) + (dy * dy) + (dz * dz);
            }).reversed());
            // and then draw in the order of `regions`
```

### Log File

The log file is not relevant for this issue.

### Crash Report

There is no crash.

embeddedt:
Fixes https://github.com/CaffeineMC/sodium-fabric/issues/2132 by ensuring the list of render regions is always sorted by distance from the camera.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
@@ -24,6 +24,8 @@ public class ChunkRenderList {
 
     private int lastVisibleFrame;
 
+    private double distanceFromCamera;
+
     public ChunkRenderList(RenderRegion region) {
         this.region = region;
     }
@@ -101,4 +103,14 @@ public class ChunkRenderList {
         return this.region;
     }
 
+    /**
+     * Get the squared distance of this region from the camera.
+     */
+    public double getDistanceFromCamera() {
+        return this.distanceFromCamera;
+    }
+
+    public void setDistanceFromCamera(double d) {
+        this.distanceFromCamera = d;
+    }
 }
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
@@ -24,7 +24,7 @@ public class ChunkRenderList {
 
     private int lastVisibleFrame;
 
-    private double distanceFromCamera;
+    private double sqrDistanceFromCamera;
 
     public ChunkRenderList(RenderRegion region) {
         this.region = region;
@@ -106,11 +106,11 @@ public class ChunkRenderList {
     /**
      * Get the squared distance of this region from the camera.
      */
-    public double getDistanceFromCamera() {
-        return this.distanceFromCamera;
+    public double getSqrDistanceFromCamera() {
+        return this.sqrDistanceFromCamera;
     }
 
-    public void setDistanceFromCamera(double d) {
-        this.distanceFromCamera = d;
+    public void setSqrDistanceFromCamera(double d) {
+        this.sqrDistanceFromCamera = d;
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/SortedRenderLists.java
```diff
@@ -44,7 +48,17 @@ public void add(RenderSection section) {
...
+            //noinspection ForLoopReplaceableByForEach
+            for (int i = 0; i < this.lists.size(); i++) {
+                ChunkRenderList renderList = this.lists.get(i);
+                RenderRegion region = renderList.getRegion();
+                double dx = cameraPos.x - region.getCenterX();
+                double dy = cameraPos.y - region.getCenterY();
+                double dz = cameraPos.z - region.getCenterZ();
+                renderList.setDistanceFromCamera((dx * dx) + (dy * dy) + (dz * dz));
+            }
+            this.lists.sort(LIST_DISTANCE_COMPARATOR);
```
douira, line 61: it might be faster to use `unstableSort` here because that can use quicksort which is allocation free. I can't think of a reason for a stable sort to be necessary. `ObjectArrayList` does a stable sort by default.

embeddedt: Is this implementation of quicksort O(n) on sorted inputs? I think `sort` uses timsort which should have that behavior.

douira: yeah the default `Arrays.sort` that it uses for stable sort might just be better. The quicksort  is kinda hard to read so I'm not sure it'd actually fare better.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
```diff
@@ -101,4 +103,14 @@ public RenderRegion getRegion() {
         return this.region;
     }
 
+    /**
+     * Get the squared distance of this region from the camera.
+     */
+    public double getDistanceFromCamera() {
```
douira: to avoid confusion calling this explicitly squared in the method name might be good. (same for the companion setter method and the field)

douira (commented):
I'm in favour of this PR

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/lists/ChunkRenderList.java
embeddedt: Agreed. Changes made.

jellysquid3:
The render regions should always be sorted as a matter of how the chunk traversal algorithm works. If that isn't happening, it means the traversal algorithm is broken. I would much rather fix the underlying issue rather than try to workaround it.

PR #2135, opened 2023-10-22
fix: don't check entities if it requires iterating too many sections

### LINKED ISSUE #2114: Display entities with large width/height make entity culling incredibely slow
### Bug Description

And I do mean *slow*: 1.5s per frame!

### Reproduction Steps

1. Enable entity culling in Sodium options.
2. `/summon minecraft:item_display ~ ~ ~ {"item":{"id":"minecraft:red_dye","Count":1b},"transformation":{"translation":[0.0,0.0,0.0],"right_rotation":[0.0,0.0,0.0,1.0],"scale":[1.0,0.0,1.0],"left_rotation":[0.0,0.0,0.0,1.0]},"width":9999,"height":9999}`
3. Optional: disable entity culling to return to playable FPS.
4. `/kill @e[type=minecraft:item_display,sort=nearest,limit=1,distance=..10]`

### Log File

N/A (No other performance mods loaded.)

### Crash Report

N/A

douira:
Closes #2114

Works by calculating the volume of each entity and checking if it's too large. Too large entities are just always rendered because checking all sections they touch for visibility is too expensive. The threshold of 15 sections can be adjusted if necessary.

Based on the suggestion embeddedt made here: https://github.com/CaffeineMC/sodium-fabric/issues/2114#issuecomment-1755555960

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -371,6 +371,8 @@ public class SodiumWorldRenderer {
         matrices.pop();
     }
 
+    // the volume of a section multiplied by the number of sections to be checked at most
+    private static final double MAX_ENTITY_CHECK_VOLUME = 16 * 16 * 16 * 15;
 
     /**
      * Returns whether or not the entity intersects with any visible chunks in the graph.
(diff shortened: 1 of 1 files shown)


embeddedt:
The fix looks good, and I can confirm this fixes #2114 on my system as well (5-7 seconds per frame before, regular FPS after).

I wonder if instead of unconditionally rendering the entity, we could fall back to a frustum check for boxes over a certain size. This is not a blocker for the PR being merged, though - it can be added later.

douira:
That's a good idea, I've added a comment that it would be a good idea for this niche case to check for the frustum. I didn't add it immediately because there's no immediate access to the frustum at this location and the `Viewport` might need modification to test arbitrary AABBs on the frustum it contains.

jellysquid3 (approved):

Issue #2147, opened 2023-11-01
Float.NaN in clouds_height of the DimensionEffects


SuperCatMaster:
### Bug Description

The End and The Nether use Float.NaN in constructor of the DimensionEffects to remove clouds. But when I tried use it in my DimensionEffects, clouds still renders.
Although in my other DimensionEffects, in which i use constant values in clouds_height, it works fine.

### Reproduction Steps

1. Register cutsom DimensionEffects via DimensionRenderingRegistry.
2. Set in registered DimensionEffects clouds_height to Float.NaN
3. Launch the game.

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/13224200/latest.log)

### Crash Report

[crash-2023-11-01_14.24.57-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/13224201/crash-2023-11-01_14.24.57-client.txt)

embeddedt:
This is a good catch. Vanilla explicitly checks whether cloud height is NaN and skips rendering if so. The same check needs to be added to Sodium.

PR #2150, opened 2023-11-01
Fix clouds still being rendered if height is NaN

embeddedt:
### Linked Issues

#2147

### Proposed Changes

Vanilla checks whether the cloud height is NaN and skips rendering them if so. We should do the same.

Note: the local variables were reordered, in order to bail out early if height is NaN, rather than computing the color.


jellysquid3:
Looks fine to me. Can you add a comment explaining why that check exists, since this is fixing a bug?

jellysquid3 (approved):

PR #2161, opened 2023-11-12
Ignore world state when rendering debug worlds

### LINKED ISSUE #2069: Sodium is weird with some blocks (desynchronized)
### Bug Description

Sodium is weird with some blocks

image
https://cdn.discordapp.com/attachments/1148716397610946623/1150450982103285820/2023-09-10_17-21-08.mp4

### Reproduction Steps

Enter the world of debugging
Try to open/close trapdoors

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/12568571/latest.log)

### Crash Report

[crash-2023-09-10_17.32.28-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/12568570/crash-2023-09-10_17.32.28-client.txt)

embeddedt:
Fixes #2069 by mimicking vanilla's very hacky approach of using the debug chunk generator directly inside `getBlockState` calls. To avoid polluting the rest of the code with `isDebugWorld` checks, we fabricate a fake block container that contains the same blocks a debug world would. This should be indistinguishable from vanilla at runtime, while not having any measurable impact on performance for regular worlds.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java
@@ -26,6 +26,7 @@ import java.util.Map;
 public class ClonedChunkSection {
     private static final ChunkNibbleArray DEFAULT_SKY_LIGHT_ARRAY = new ChunkNibbleArray(15);
     private static final ChunkNibbleArray DEFAULT_BLOCK_LIGHT_ARRAY = new ChunkNibbleArray(0);
+    private static final PalettedContainer<BlockState> DEFAULT_STATE_CONTAINER = new PalettedContainer<>(Block.STATE_IDS, Blocks.AIR.getDefaultState(), PalettedContainer.PaletteProvider.BLOCK_STATE);
 
     private final ChunkSectionPos pos;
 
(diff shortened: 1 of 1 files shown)


src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java
```diff
@@ -71,6 +75,34 @@ public ClonedChunkSection(World world, WorldChunk chunk, @Nullable ChunkSection
...
 
+    /**
+     * Construct a fake PalettedContainer whose contents match those of the debug world. This is needed to
+     * match vanilla's odd approach of short-circuiting getBlockState calls inside its render region class.
+     */
+    @NotNull
+    private static PalettedContainer<BlockState> constructDebugWorldContainer(ChunkSectionPos pos) {
+        var container = new PalettedContainer<>(Block.STATE_IDS, Blocks.AIR.getDefaultState(), PalettedContainer.PaletteProvider.BLOCK_STATE);
+        // We use swapUnsafe to avoid going through the lock when we own the container
+        for (int y = 0; y < 16; y++) {
```
jellysquid3: Can we avoid iterating over every Y-coordinate here? The code only cares about Y-level 60 and 70.

src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java
```diff
@@ -71,6 +75,34 @@ public ClonedChunkSection(World world, WorldChunk chunk, @Nullable ChunkSection
...
+            int worldY = ChunkSectionPos.getOffsetPos(pos.getY(), y);
+            if (worldY == 60) {
+                BlockState state = Blocks.BARRIER.getDefaultState();
+                for (int z = 0; z < 16; z++) {
+                    for (int x = 0; x < 16; x++) {
+                        container.swapUnsafe(x, y, z, state);
```
jellysquid3: Why use `PalettedContainer.swapUnsafe`? We don't care about the existing value. We can just use `PalettedContainer.set(int, Object)` instead.

jellysquid3 (changes requested):
I hate this workaround, but understand that it's a necessary evil. While it is not likely to be called often in practice, I would like to clean up the code slightly.

src/main/java/me/jellysquid/mods/sodium/client/world/cloned/ClonedChunkSection.java
embeddedt: `set` acquires and releases a lock on each call.

jellysquid3: Ah... that method is not public. Oh well.

jellysquid3 (approved):

PR #2165, opened 2023-11-16
Add a flag to disable Sodium checking for RivaTuner

Niterux:
I would like to add this flag to Sodium to disable its check to RTSS as I use RTSS along side Minecraft all of the time and I have no issues pertaining to log spam or performance problems.
A picture of Minecraft a long side the client log and RTSS

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/platform/windows/WindowsModuleChecks.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/util/workarounds/platform/windows/WindowsModuleChecks.java
@@ -27,7 +27,7 @@ public class WindowsModuleChecks {
         }
 
         // RivaTuner hooks the wglCreateContext function, and leaves itself behind as a loaded module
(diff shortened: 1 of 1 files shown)


MCRcortex:
from what i can tell in the screenshot however is that rivatuner isnt displaying an overlay, drawing the overlay itself is done via legacy opengl calls which is what causes havoc (unless i am missing something and rivatuner is drawing something in minecraft)

jellysquid3:
Since we don't really know what conditions will cause the problem, I guess that I am okay with adding a flag. At some point we should consider moving these flags into a separate config file, or something.

jellysquid3 (approved):

Niterux:
> from what i can tell in the screenshot however is that rivatuner isnt displaying an overlay, drawing the overlay itself is done via legacy opengl calls which is what causes havoc (unless i am missing something and rivatuner is drawing something in minecraft)

Here, I have enabled the overlay
rtsswithoverlay

PR #2169, opened 2023-11-19
Fix MatrixHelper producing slightly different results from vanilla

embeddedt:
Vanilla uses JOML's `Vector4f#mul` or `Vector3f#mul` methods to apply matrix transformations to vectors. The implementation of these methods uses calls like
```java
dest.x = Math.fma(mat.m00(), x, Math.fma(mat.m10(), y, Math.fma(mat.m20(), z, mat.m30() * w)));
```

and the `Math.fma` wrappers have the effect of doing the final additions right-to-left, rather than left-to-right. Since Sodium's `MatrixHelper` uses left-to-right addition, the value it produces might differ very slightly from JOML's as a result of floating-point precision errors. This can cause Z-fighting.

The solution I used is to simply add brackets to force the additions to be performed in the same order as JOML.

Discord thread where the issue was reported: https://discord.com/channels/602796788608401408/1174666147828879400

The user and I both tested & confirmed this fixes their issue.

### DIFF
--- a/src/api/java/net/caffeinemc/mods/sodium/api/math/MatrixHelper.java
+++ b/src/api/java/net/caffeinemc/mods/sodium/api/math/MatrixHelper.java
@@ -45,7 +45,7 @@ public class MatrixHelper {
      * @return The transformed X-coordinate for the normal vector
      */
     public static float transformNormalX(Matrix3f mat, float x, float y, float z) {
-        return (mat.m00() * x) + (mat.m10() * y) + (mat.m20() * z);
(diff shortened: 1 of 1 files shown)


jellysquid3:
This bug description makes my bones hurt.

jellysquid3 (approved):

jellysquid3:
Jokes aside, the solution looks good, and it seems like our suspicions in Discord were correct as well. Thanks for the write-up.

PR #2173, opened 2023-11-20
Add bounds checking in WorldSlice

embeddedt:
Currently, calling `WorldSlice#getBlockState` or related methods with a block position that wasn't cloned will either return `null` or throw `ArrayIndexOutOfBoundsException`. This tends to cause crashes in mods, and is not very useful for debugging.

To solve this, we check that the given position is within bounds, and return a sane default value if not:

* `Blocks.AIR.getDefaultState()` for block states
* 0 for light levels
* `null` for block entities and block entity render attachments

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/WorldSlice.java
@@ -62,6 +62,9 @@ public final class WorldSlice implements BlockRenderView, BiomeColorView, Render
     // The number of bits needed for each local X/Y/Z coordinate.
     private static final int LOCAL_XYZ_BITS = 4;
 
+    // The default block state used for out-of-bounds access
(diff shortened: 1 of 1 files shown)


jellysquid3 (changes requested):
Checking the section index against the bounds of the array will probably not do what you want. Each coordinate has to be checked for the out-of-range condition individually. For example, coordinates [4, 0, 0] will produce the same index as coordinates [1, 0, 1]. While that is still in bounds of the flat array, it will produce hard to debug issues that manifest as seemingly random corruption.

jellysquid3:
Some other thoughts:
- We could just initialize the block data arrays to the default state, rather than leaving null entries in there. That would eliminate the need for checking if there is a null entry.
- The implementation could be changed to store a single `34x34x34` array of block data, rather than a table of sections -> block data arrays. That way, we can consolidate the "out of bounds" and "no data exists" cases into one, and simplify the code a bit. This might actually be faster, anyways, since it reduces the amount of indirection and index calculations.

jellysquid3 (approved):

PR #2174, opened 2023-11-21
Hardened Gradle builds with Gradle dependency verification and defined module restrictions on third-party maven repositories.

0-x-2-2:
This prevents anyone from replacing dependencies from under our feet for any purpose, either with malicious intent or by accident.

See the Gradle documentation for more information:
https://docs.gradle.org/current/userguide/dependency_verification.html


build.gradle
```diff
@@ -74,12 +74,50 @@ jar {
...
 }
 
-dependencies {
-    //to change the versions see the gradle.properties file
-    minecraft "com.mojang:minecraft:${project.minecraft_version}"
-    mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
-    modImplementation "net.fabricmc:fabric-loader:${project.loader_version}"
+repositories {
+    // Log if any unknown repositories were added non-explicitly.
+    forEach { repo ->
```
0-x-2-2, line 79: This can be removed if you don't like it.

build.gradle
```diff
@@ -74,12 +74,50 @@ jar {
     from sourceSets.api.output.resourcesDir
 }
 
-dependencies {
-    //to change the versions see the gradle.properties file
-    minecraft "com.mojang:minecraft:${project.minecraft_version}"
-    mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
-    modImplementation "net.fabricmc:fabric-loader:${project.loader_version}"
+repositories {
```
0-x-2-2, line 77: This entire `repositories` block is not strictly required either but prevents downloading artifacts from the wrong place.

0-x-2-2:
I have left some comments on the `repositories` block as it is the only thing that is a tiny bit problematic.

jellysquid3:
Thanks for this pull request. We talked about it again earlier today, and I think it's a great improvement to the build pipeline. This will help prevent some kinds of build pipeline attacks, and certainly catch many subtle issues on the CI/CD. 

I want to investigate merging this for the next major release, as changing it now in the Sodium 0.5.x branch would probably be disruptive to other projects and developers.

jellysquid3:
I also think we will likely need some documentation added to the repository (possibly CONTRIBUTING.md) explaining how this works, and how to update the pinned dependencies.  As far as I can tell, we'd be one of the only Fabric mods doing this, so people aren't going to be familiar with it.

douira:
I'm going to close this as it has gone thoroughly stale but I think the approach is still interesting.

PR #2210, opened 2023-12-15
Change BlockRenderType check to match vanilla

embeddedt:
Does what it says on the tin, vanilla only requires that the type is not INVISIBLE.


embeddedt:
Turns out vanilla has a second check in `BlockRenderManager` (that I missed) which does require the type to be MODEL, so the current behavior is fine.

PR #2222, opened 2023-12-26
Added biome color cache support for custom color resolver

### LINKED ISSUE #2166: `WorldSlice#getColor` cannot be called with custom ColorResolvers
### Request Description

In Sodium 0.5+, `WorldSlice#getColor` converts to a `BiomeColorSource` internally before retrieving the color, and `BiomeColorSource` is an enum that only supports GRASS, WATER, and FOLIAGE. Therefore, there is no way to pass a custom `ColorResolver` and have it work correctly with Sodium 0.5+. An example of a crash this causes (albeit with a Forge mod) can be found [here](https://gist.githubusercontent.com/Gbergz/4dc829382e4a488611cca52787c6031e/raw/c1b621f76b46ad70bc403ff242dd0e0c807f8981/gistfile1.txt).

I think the best way to solve this issue is to remove the enum, and refactor `BiomeColorCache` to look up the appropriate `ColorBuffer` based on the provided `ColorResolver` instance. If this solution sounds reasonable, I am willing to make a PR.

### LINKED ISSUE #2221: Incompatibility with Vanadium and Continuity together: crash with UnsupportedOperationException
### Bug Description

Game version: 1.20.1
Fabric version: 0.15.3

### Mod list (minimum):
- Cloth config 11.1.118
- Continuity 3.0.0-beta.4+1.20.1
- Indium 1.0.27+mc1.20.1
- Sodium 0.5.3
- Vanadium 0.1.675
- fabric-api-0.91.0+1.20.1

### Description
Sodium doesn't support custom ColorResolver (`net.minecraft.world.biome.ColorResolver`) and may throw an UnsupportedOperationException error if attempted.

### Reproduction Steps

1. Install the above-mentioned mods
2. Load the `Default Connected Textures` resource pack from Continuity and any resource pack containing a custom colormap property file (`/assets/minecraft/optifine/colormap/blocks/{any}.properties`). For instance, you can use [Patrix 32x](https://www.curseforge.com/minecraft/texture-packs/patrix-32x)
3. Enter a world
4. The game will then crash

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/13770939/latest.log)

### Crash Report

[crash-2023-12-26_18.41.27-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/13770941/crash-2023-12-26_18.41.27-client.txt)

CPTProgrammer:
Added biome color cache support for custom color resolver. 

`WorldSlice#getColor` can now be invoked with a custom `ColorResolver`.

- Fixed #2221
- Fixed #2166

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
@@ -15,10 +15,8 @@ public class BiomeColorCache {
     private static final int NEIGHBOR_BLOCK_RADIUS = 2;
     private final BiomeSlice biomeData;
 
-    private final Slice[] slices;
-    private final HashMap<ColorResolver, CustomSlice[]> customSlices;
-    private final boolean[] populatedSlices;
-    private final HashMap<ColorResolver, boolean[]> populatedCustomSlices;
+    private final HashMap<ColorResolver, Slice[]> slices;
+    private final HashMap<ColorResolver, boolean[]> populatedSlices;
 
     private final int blendRadius;
 
@@ -36,14 +34,8 @@ public class BiomeColorCache {
         this.sizeXZ = 16 + ((NEIGHBOR_BLOCK_RADIUS + this.blendRadius) * 2);
         this.sizeY = 16 + (NEIGHBOR_BLOCK_RADIUS * 2);
 
-        this.slices = new Slice[this.sizeY];
-        this.customSlices = new HashMap<>();
-        this.populatedSlices = new boolean[this.sizeY];
-        this.populatedCustomSlices = new HashMap<>();
-
-        for (int y = 0; y < this.sizeY; y++) {
-            this.slices[y] = new Slice(this.sizeXZ);
-        }
+        this.slices = new HashMap<>();
+        this.populatedSlices = new HashMap<>();
 
         this.tempColorBuffer = new ColorBuffer(this.sizeXZ, this.sizeXZ);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -94,6 +137,29 @@ private void updateColorBuffers(int relY) {
...
+        int worldY = this.minY + relY;
+
+        for (int worldZ = this.minZ; worldZ <= this.maxZ; worldZ++) {
+            for (int worldX = this.minX; worldX <= this.maxX; worldX++) {
+                Biome biome = this.biomeData.getBiome(worldX, worldY, worldZ).value();
+
+                int relativeX = worldX - this.minX;
+                int relativeZ = worldZ - this.minZ;
+
+                slice.buffer.set(relativeX, relativeZ, BiomeColors.GRASS_COLOR.getColor(biome, worldX, worldZ));
```
jellysquid3: This does not actually use the resolver passed to the function, so custom resolvers will always use the default grass color...

jellysquid3 (changes requested):

jellysquid3:
I don't think the solution is to duplicate code paths for "custom" and "vanilla" color resolvers, since it's fundamentally the same interface being used. There should just be one code path, and a `ColorResolver->int[]` map.

CPTProgrammer:
Sorry for my carelessness, I'll change the code as soon as possible

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
CPTProgrammer: it has been fixed :)

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -6,15 +6,17 @@
...
+import java.util.HashMap;
 
 public class BiomeColorCache {
     private static final int NEIGHBOR_BLOCK_RADIUS = 2;
     private final BiomeSlice biomeData;
 
-    private final Slice[] slices;
-    private final boolean[] populatedSlices;
+    private final HashMap<ColorResolver, Slice[]> slices;
+    private final HashMap<ColorResolver, boolean[]> populatedSlices;
```
jellysquid3: We should avoid having two hash maps, and instead just store the "is populated" flag inside the `Slice` since it avoids the overhead of additional lookups, and makes the code cleaner.

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -23,21 +25,19 @@ public class BiomeColorCache {
...
-
-        this.slices = new Slice[sizeY];
-        this.populatedSlices = new boolean[sizeY];
+        this.sizeXZ = 16 + ((NEIGHBOR_BLOCK_RADIUS + this.blendRadius) * 2);
+        this.sizeY = 16 + (NEIGHBOR_BLOCK_RADIUS * 2);
 
-        for (int y = 0; y < sizeY; y++) {
-            this.slices[y] = new Slice(sizeXZ);
-        }
+        this.slices = new HashMap<>();
```
jellysquid3: Should be using `Reference2ReferenceOpenHashMap` since performance is critical, and caches are only valid according to reference equality. The type of the field should also be updated to match.

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -49,26 +49,47 @@ public void update(ChunkRenderContext context) {
...
+
+    public int getColor(ColorResolver resolver, int blockX, int blockY, int blockZ) {
         var relX = MathHelper.clamp(blockX, this.minX, this.maxX) - this.minX;
         var relY = MathHelper.clamp(blockY, this.minY, this.maxY) - this.minY;
         var relZ = MathHelper.clamp(blockZ, this.minZ, this.maxZ) - this.minZ;
 
-        if (!this.populatedSlices[relY]) {
-            this.updateColorBuffers(relY);
+        if (!this.slices.containsKey(resolver)) {
+            this.slices.put(resolver, new Slice[this.sizeY]);
```
jellysquid3: This branch should be extracted to a new function since it's rather large and a rarely taken path (we don't want the CPU fetching instructions into the cache which are not helpful in most invocations).

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -49,26 +49,47 @@ public void update(ChunkRenderContext context) {
...
+        var slice = this.slices.get(resolver)[relY];
+        var buffer = slice.getBuffer();
 
         return buffer.get(relX, relZ);
     }
 
-    private void updateColorBuffers(int relY) {
-        var slice = this.slices[relY];
+    private void updateColorBuffers(int relY, ColorResolver resolver) {
+        var slice = this.slices.get(resolver)[relY];
```
jellysquid3: We already created the slice before calling this method, so we should pass it to this function instead of querying it again.

jellysquid3 (changes requested):
The abstract of this looks good, just some performance concerns.

src/main/java/me/jellysquid/mods/sodium/client/world/biome/BiomeColorCache.java
```diff
@@ -49,26 +49,47 @@ public void update(ChunkRenderContext context) {
         this.maxY = (context.getOrigin().getMaxY() + NEIGHBOR_BLOCK_RADIUS);
         this.maxZ = (context.getOrigin().getMaxZ() + NEIGHBOR_BLOCK_RADIUS) + this.blendRadius;
 
-        Arrays.fill(this.populatedSlices, false);
+        for (boolean[] p : this.populatedSlices.values()) {
+            Arrays.fill(p, false);
+        }
```
CPTProgrammer: Optimizing performance by using an incremental flag (without loops) to check for populated status instead of relying on two nested for-loops.

CPTProgrammer (commented):
Performance optimization

CPTProgrammer:
All requested changes have been implemented

jellysquid3 (approved):

jellysquid3:
Thanks for the contribution. It is worth mentioning that Fabric API is considering [implementing an interface](https://github.com/FabricMC/fabric/pull/3503) for mods to use custom ColorResolvers (since vanilla _currently_ doesn't allow this, and Sodium is exposing additional functionality) but this should not affect your patch at all.

embeddedt (commented):
Looks good from my end as well.

MehVahdJukaar:
Will this be added to previous version aswell? 
The fact that its not possible to add custom ColorResolvers is a deal breakers for many mods that use those things and its made way worse by how locked and up unforgiving the current architecture is with that enum that cant be extended or the fact that the game hardcrashes when a color resolved that is not vanilla shows up.
This is relevant to my mod [Polytone](https://legacy.curseforge.com/minecraft/mc-mods/polytone) which currently crashes with Sodium

jellysquid3:
Currently there's no back-porting policy (see #2230). But we'll likely publish future versions of Sodium 0.5.x for both MC 1.20.1 (LTS) and MC 1.20.4 (Latest) in the meanwhile.

PR #2227, opened 2023-12-29
change: added ID argument to OptionImpl, OptionGroup and OptionPage

SrRapero720:
# ID argument
- Uses Identifier/ResourceLocation to avoid any colliding with any extra option added by other mods/addons
- Splitted in minecraft and sodium namespaces
- Implemented with not breaking changes at the moment: my suggestion was mark ID as required after any sodium release with more relevant breaking changes
- Implemented List to store Suppliers for extra option pages and attach them after sodium pages.

## WHY?
Mixin into options is more easier, instead of inject and specify high ordinals. you can inject on all and just do `OptionImpl#getId().equals(new ResourceLocation("minecraft:fullscreen"))`, this helps to sodium addons which adds options on empty spaces and forks like embeddium which option pages differs each minecraft version

## EXTRA PAGES 
`public static void addOptionsPage()` and `private void addExtraPages()` is a simple helper just to add extra pages from sodium addons (or mod integrations) at the end of sodium options, ~~Is a crime do mixins just to add a page~~

extra pages was added outside constructor just to avoid any potential mixin issue

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
@@ -21,7 +21,7 @@ public class SodiumClientMod implements ClientModInitializer {
     @Override
     public void onInitializeClient() {
         ModContainer mod = FabricLoader.getInstance()
-                .getModContainer(ID)
+                .getModContainer("sodium")
                 .orElseThrow(NullPointerException::new);
 
         ID = mod.getMetadata()
--- a/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
@@ -14,19 +14,27 @@ public class SodiumClientMod implements ClientModInitializer {
     private static SodiumGameOptions CONFIG;
     private static Logger LOGGER;
 
-    private static String MOD_VERSION;
+    public static String ID;
+    public static String NAME;
+    public static String MOD_VERSION;
 
     @Override
     public void onInitializeClient() {
         ModContainer mod = FabricLoader.getInstance()
-                .getModContainer("sodium")
+                .getModContainer(ID)
                 .orElseThrow(NullPointerException::new);
+
+        ID = mod.getMetadata()
+                .getId();
+
+        NAME = mod.getMetadata()
+                .getName();
 
         MOD_VERSION = mod.getMetadata()
                 .getVersion()
                 .getFriendlyString();
 
-        LOGGER = LoggerFactory.getLogger("Sodium");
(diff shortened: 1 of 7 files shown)


src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
```diff
@@ -14,19 +14,27 @@ public class SodiumClientMod implements ClientModInitializer {
...
 
     @Override
     public void onInitializeClient() {
         ModContainer mod = FabricLoader.getInstance()
-                .getModContainer("sodium")
+                .getModContainer(ID)
                 .orElseThrow(NullPointerException::new);
 
+        ID = mod.getMetadata()
+                .getId();
```
Octol1ttle: These assignments create a circular dependency

src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptionPages.java
```diff
@@ -104,34 +115,40 @@ public static OptionPage general() {
...
+        return new OptionPage(new Identifier(SodiumClientMod.ID, "general"), Text.translatable("stat.generalButton"), ImmutableList.copyOf(groups));
     }
 
     public static OptionPage quality() {
         List<OptionGroup> groups = new ArrayList<>();
 
         groups.add(OptionGroup.createBuilder()
+                .setId(new Identifier("minecraft", "graphics"))
                 .add(OptionImpl.createBuilder(GraphicsMode.class, vanillaOpts)
+                        .setId(new Identifier("minecraft", "graphics"))
```
Octol1ttle:
```suggestion
                        .setId(new Identifier("minecraft", "graphics_mode"))
```

src/main/java/me/jellysquid/mods/sodium/client/SodiumClientMod.java
SrRapero720: whoops!

jellysquid3:
This code isn't part of any public API which Sodium is exposing... If other mods are using mixins to inject their own options, we aren't under the obligation to then support that. Doing so would make it impossible to change anything about our internal code, because we would be giving the expectation that every internal detail is "exposed API" and that mods should be able to rely on it.

In other words, we are _not Minecraft itself_, and we don't provide a stable API for other mods to interact with. So I am not considering any pull requests to better support mods that are using mixins to modify our code. We have enough issues with it already, and giving any precedent that we're somehow okay with it is not a great idea.

If these mods _need_ a clean way to extend the Video Settings screen, then a solution needs to be reached that doesn't require other mods to depend on internal details and the use of mixins. But so far we've not received any proposals for doing that, and it's not really clear to me what people actually need (or whether Sodium itself should even be providing it.)

PR #2234, opened 2023-12-30
Add new viewport test method

MCRcortex:
Adds a viewport frustum check method that can accept per axis sizes

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/viewport/Viewport.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/viewport/Viewport.java
@@ -41,6 +41,21 @@ public final class Viewport {
         );
     }
 
+    public boolean isBoxVisible(int intX, int intY, int intZ, float size_x, float size_y, float size_z) {
+        float floatX = (float)(intX - this.transform.intX) - this.transform.fracX;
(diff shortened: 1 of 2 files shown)


src/main/java/me/jellysquid/mods/sodium/client/render/viewport/Viewport.java
```diff
@@ -25,20 +25,19 @@ public Viewport(Frustum frustum, Vector3d position) {
...
+    public boolean isBoxVisible(int intX, int intY, int intZ, float size_x, float size_y, float size_z) {
+        float floatX = (float)(intX - this.transform.intX) - this.transform.fracX;
+        float floatY = (float)(intY - this.transform.intY) - this.transform.fracY;
+        float floatZ = (float)(intZ - this.transform.intZ) - this.transform.fracZ;
 
         return this.frustum.testAab(
-                floatX - radius,
-                floatY - radius,
-                floatZ - radius,
+                floatX,
```
jellysquid3, line 34: Why was the negative extent removed here? The code calling this still passes the center of the bounding box, not the bottom-left corner, so the results will be wrong.

jellysquid3 (changes requested):

PR #2250, opened 2024-01-08
Add warning when the user double clicks the jar file like a program

douira:
When the user clicks on the jar file it should now open a dialog that explains that they should either install it as a fabric jar or use the Modrinth app instead. The exact wording of the message can be adjusted as necessary. `LaunchWarn.java` was directly copied and adjusted from Iris, but the integration with gradle is different because Iris uses Brachyura. Because apparently there's sometimes issues when the double-click launched class isn't java 8, this makes that class compiled with java 8.

Known Issue: Doesn't work with macOS' JavaLauncher.app for some reason, I don't know why. It works with an iris jar, but not sodium. I've found reports that JavaLauncher behaves mysteriously in general, so this might be a result of that. If anyone finds a difference between the jar this produces and an Iris jar which _does_ work with JavaLauncher on my system, let me know. It does work with `java -jar sodium.jar` on my system.

### DIFF
--- a/src/launch_warn/java/net/caffeinemc/mods/sodium/launch_warn/LaunchWarn.java
+++ b/src/launch_warn/java/net/caffeinemc/mods/sodium/launch_warn/LaunchWarn.java
@@ -1,0 +1,48 @@
+package net.caffeinemc.mods.sodium.launch_warn;
+
+import javax.swing.JOptionPane;
+import javax.swing.UIManager;
+import javax.swing.UnsupportedLookAndFeelException;
+import java.awt.Desktop;
+import java.awt.GraphicsEnvironment;
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Taken from
+ * https://github.com/IrisShaders/Iris/blob/6c880cd377d97ffd5de648ba4dfac7ea88897b4f/src/main/java/net/coderbot/iris/LaunchWarn.java
+ * and modified to fit Sodium. See Iris' license for more information.
+ */
+public class LaunchWarn {
+    public static void main(String[] args) {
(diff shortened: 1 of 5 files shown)


jellysquid3:
This looks fine, but let's not push Modrinth's launcher so heavily right now, and just use a translation string like:

> You have tried to launch Sodium (a Minecraft mod) directly, but it is not an executable program or mod installer. You must install Fabric Loader for Minecraft, and place this file in your mods directory instead.
> 
> If this is your first time installing mods for Fabric Loader, click ["Help"](https://) for a guide on how to do this.

We can update the Wiki page later on to recommend the Modrinth launcher (or whatever else) without updating the mod.

src/launch_warn/java/net/caffeinemc/mods/sodium/launch_warn/LaunchWarn.java
```diff
@@ -0,0 +1,48 @@
+package net.caffeinemc.mods.sodium.launch_warn;
```
jellysquid3: It might make more sense to use `desktop` instead of `launch_warn` for the source set, just in case we need any other desktop-mode stuff.  And, to be pedantic, it avoids the underscore.

douira: I've changed the source set and package to be `desktop`.

douira:
Since `JOptionPane` has very limited options for the button labels I've worded this to be compatible with that. (if there's a way to change what the buttons say let me know)

douira:
I've changed it to actually say "cancel" and "help" now.

jellysquid3 (approved):
Thanks!

douira:
Turns out I messed up the wording in the comment in `build.gradle`: `// this source set is should be compiled with Java 8`

PR #2254, opened 2024-01-10
Mark sprites rendered through vanilla block model renderer as active

embeddedt:
Sodium already includes an intrinsic code path for the simpler `BlockModelRenderer#render` method (the one which only loops over the quads and pipes them to a VertexConsumer), and that marks animated sprites as active. However, mods can call the *other* `BlockModelRenderer#render` method (the one which applies smooth lighting if needed) during immediate-mode rendering, and currently sprites rendered there are not detected. This PR fixes that.

Thanks to @19PHOBOSS98 for reporting this and noticing the second code path in `BlockModelRenderer`.


jellysquid3 (approved):
LGTM. Eventually we should replace this immediate-mode path with our own, but our current pipeline expects a terrain context and other stuff.

PR #2267, opened 2024-01-19
Use the same vanilla terminology for biome blend option in config screen

MeeniMc:
Use the same vanilla terminology for biome blend option in config screen (e.g., 5x5 is a radius of 2)

The blend None (a.k.a. 1x1, or distance 0) can be set through the Shift-P vanilla screen (or inherited from a prior config), that does change the visual result compared to 3x3 (distance 1), so it has been enabled in sodium config as well (but it doesn't enable sharp non-blended mode like in vanilla). 

Thanks to the new option screen the value shows with the formatter correctly even when the widget is held.


src/main/resources/assets/sodium/lang/en_us.json
```diff
@@ -24,8 +24,8 @@
   "sodium.options.leaves_quality.tooltip": "Controls whether leaves will be rendered as transparent (fancy) or opaque (fast).",
   "sodium.options.particle_quality.tooltip": "Controls the maximum number of particles which can be present on screen at any one time.",
   "sodium.options.smooth_lighting.tooltip": "Enables the smooth lighting and shading of blocks in the world. This can very slightly increase the amount of time it takes to load or update a chunk, but it doesn't affect frame rates.",
-  "sodium.options.biome_blend.value": "%s block(s)",
-  "sodium.options.biome_blend.tooltip": "The distance (in blocks) which biome colors are smoothly blended across. Using higher values will greatly increase the amount of time it takes to load or update chunks, for diminishing improvements in quality.",
+  "sodium.options.biome_blend.value": "%sx%s block(s)",
+  "sodium.options.biome_blend.tooltip": "The distance (in blocks) which biomes color blend smoothly. Default is 5x5. Using higher values will greatly increase the amount of time it takes to load or update chunks, for diminishing improvements in quality.",
```
Madis0: The first sentence change of the tooltip doesn't seem better here

MeeniMc:
Updated according to discord discussion. "[...] but it should just be implemented using a custom formatter (manipulating the actual config value like that is problematic)".

2 notes: 
1. When using the shift-p vanilla screen, it would be possible to set blending radius to 0. The resulting Sodium configuration works. Solution: I let the Sodium slider take the 0 input, and clarified in the tooltip that this case is not blended, but still smoothed. 
2. It was not clear to me what was 'problematic' with the prior approach. The value in the config file, and in the rest of the codebase would still be expressed in radius (unchanged from prior versions of Sodium, that is, 2 for the 5x5 case). I did the requested changes anyway, thus when holding the slider widget, it will now show the radius value, but we still clarify how e.g., 2 is equivalent to 5x5 Vanilla nomenclature when the widget is not held.

MeeniMc:
I'm happy to fix the conflict if there is still interest in this. Please close otherwise.

RedCarlos26:
Was visiting old PRs. This would be great for game consistency.

douira:
Works well for me.

douira:
Is it correct that this displays 0 as "None" when it doesn't actually disable blending but just turns it down to a minimum?

MeeniMc:
It turns biome blending off completely, but color smoothing remains.

PR #2271, opened 2024-01-24
Avoid shader compiler warnings

douira:
Usually it prints the following warnings to the console:
```
[18:32:03] [Render thread/WARN] (GlProgram) Program link log for sodium:chunk_shader: WARNING: Output of vertex shader 'v_MaterialAlphaCutoff' not read by fragment shader

[18:32:03] [Render thread/WARN] (GlProgram) Program link log for sodium:chunk_shader: WARNING: Output of vertex shader 'v_MaterialAlphaCutoff' not read by fragment shader
```
This might not happen on all systems due to differences in drivers, but on macOS it always does this when loading or reloading a world. Not declaring this out variable avoids the shader compiler warning by only adding it if the fragment shader is actually going to read it.


jellysquid3:
I noticed this recently when looking at the source, seems like an easy thing to fix.

PR #2272, opened 2024-01-25
Self Rendering/No Occlusion api

IThundxr:
Port of Steam 'n' Rails no occlusion api
[Usage](https://github.com/Layers-of-Railways/Railway/blob/49748c58a90ea9eb91ce821a4c938635cb6098ee/common/src/main/java/com/railwayteam/railways/content/palettes/boiler/BoilerBlock.java#L57-L60) [Interface](https://github.com/Layers-of-Railways/Railway/blob/1.19/dev/common/src/main/java/com/railwayteam/railways/mixin_interfaces/IForceRenderingSodium.java) [Mixin](https://github.com/Layers-of-Railways/Railway/blob/1.19/dev/common/src/main/java/com/railwayteam/railways/mixin/compat/sodium/MixinBlockOcclusionCache.java)

Currently we mixin into sodium to stop our boiler blocks from culling with any normal block since they are a obj model that extends outside the 16x16 block size, forge/porting-lib's obj loader doesnt support specifiying how it should cull so we had to resort to this, the names and stuff can definitely be improved on so if anyone has better ideas lmk!


src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/BlockOcclusionCache.java
```diff
@@ -35,6 +36,10 @@ public boolean shouldDrawSide(BlockState selfState, BlockView view, BlockPos pos
 
         BlockState adjState = view.getBlockState(adjPos);
 
+        // If a mod/block wants to handle culling itself
+        if (RenderBlockCallback.EVENT.invoker().selfManageOcclusion(selfState, adjState, facing))
```
douira: The style requires curly braces here.

src/main/java/me/jellysquid/mods/sodium/client/events/RenderBlockCallback.java
```diff
@@ -0,0 +1,34 @@
...
+ * });
+ * }
+ * </pre>
+ */
+@FunctionalInterface
+public interface RenderBlockCallback {
+    Event<RenderBlockCallback> EVENT = EventFactory.createArrayBacked(RenderBlockCallback.class,
+            (listeners) -> (state, adjacentBlockState, direction) -> {
+                for (RenderBlockCallback listener : listeners) {
+                    return listener.selfManageOcclusion(state, adjacentBlockState, direction);
```
douira: Does it make sense to return directly in this loop if there are multiple listeners. Also, is it performant to run this double lambda and an iterator loop on every block? (it is cached afaik, but still)

IThundxr: 1. Pretty sure it'll have to be inside the loop since this is whats actually calling the method on the like callback letting x mod do whatever checks/culling they want
2. havent tested but ill try it out right now

douira: I meant that only the first listener is called if you return in the loop like that, which I assume isn't intentional.

IThundxr: RenderBlockCallback takes up 255ms on my machine which is 0.14% of all

running `RenderBlockCallback.EVENT.register((state, adjacentBlockState, direction) -> (adjacentBlockState.isOf(Blocks.DIAMOND_BLOCK)));` in sodiums oninit takes up 138ms

for the invoker in shouldDrawSide it takes up 586ms which is a little higher then `BlockState adjState = view.getBlockState(adjPos)` line, in total its 24% of shouldDrawSide so it should be fine

(tested with intellij's profiler on linux with nvidia 3060 & i9-9900k)

IThundxr: > I meant that only the first listener is called if you return in the loop like that, which I assume isn't intentional.

unless im missing something it should be calling all the listeners as seen in https://fabricmc.net/wiki/tutorial:events but i may be understanding this wrongly

IThundxr: after testing yeah your correct only the first listener is called, lemme go fix that

jellysquid3:
This should really be fixed in the OBJ model loader itself, because it's not emitting the correct cull face, and our behavior isn't "wrong" -- just too strict compared to vanilla. Allowing mods to override the behavior of hidden-surface elimination, when they are the ones passing the models in the first place (and for which they have full control over the cull face in the first place), does not make a lot of sense.

PR #2273, opened 2024-01-26
Account for XDG_CURRENT_DESKTOP being unset

Phoenix-Starlight:
Simply put some distributions (puppy linux) don't have `XDG_CURRENT_DESKTOP` set.


jellysquid3:
Sizeable oof. Thanks for the fix.

PR #2291, opened 2024-01-30
Fix Fast/Fancy for graphics, clouds, weather, leaves

Madis0:
- Adds new separate "fast" and "fancy" strings for leaves and weather, for languages like Russian where reusing the Graphics ones would sound incorrect.
- Fixes graphics options using clouds' fast/fancy, and clouds options using graphics' fast/fancy


src/main/java/me/jellysquid/mods/sodium/client/gui/SodiumGameOptions.java
```diff
@@ -89,6 +89,48 @@ public boolean isFancy(GraphicsMode graphicsMode) {
...
+        @Override
+        public Text getLocalizedName() {
+            return this.name;
+        }
+
+        public boolean isFancy(GraphicsMode graphicsMode) {
+            return (this == FANCY) || (this == DEFAULT && (graphicsMode == GraphicsMode.FANCY || graphicsMode == GraphicsMode.FABULOUS));
+        }
+    }
+
```
Madis0: Not fully sure if these enums are still wanted or if I should just do a `Text[]` like it is in clouds.

PR #2295, opened 2024-01-30
Fix Value 'clicked' is always 'true'

Julienraptor01:
i saw it and then when IDEA finished loading, it highlighted it :)


douira:
This has merge conflicts but the change is trivial.

Julienraptor01:
did a simple rebase, it's now good

PR #2296, opened 2024-01-31
Add resolution control option to the config screen

### LINKED ISSUE #46: Missing vanilla graphics settings
### Expected Behavior
Expected all vanilla graphics settings to be available in Sodium.

### Actual Behavior
The following graphics settings are missing:

- [ ] Fullscreen Resolution
- [ ] Distortion Effects
- [ ] FOV Effects
- [x] Entity Shadows
- [x] Entity Distance

### Reproduction Steps

1. Open Minecraft with Sodium, Settings > Video Settings...
2. Open vanilla Minecraft, Settings > Video Settings...
3. Compare the settings

### Attachments
N/A

### System Information

- Java Version: 14.0.1 64bit
- CPU: i7-1065G7
- GPU: Mesa Iris Plus Graphics

MeeniMc:
Adds resolution control (missing option as compared to Vanilla Shift-P options).

fixes #46

The resolution controls would not fit in the allocated space, so I changed the rendering of slidercontrols to enable rendering the slider bar and the value text in the same area.
1. the slider label is right adjusted
2. the slider is rendered under the text
3. the slider is always rendered (opinionated change, don't mind rolling back that part if you don't like it) 

The code is mostly new but is liberally inspired from #957 (in particular the setBinding part)

This is how it looks like:


MeeniMc:
Note that I did not implement FOV, Distortion, Glint Speed, Glint Strenght, because these are **duplicated** in the vanilla video options from the **accessibility** option menu and so are still available without 'tricks'.

Under these assumption that we don't want to re-replicate them, we now have all options covered.

MeeniMc:
ready for review.

MeeniMc:
updated for mojmap

As for the discussion regarding delaying this after a full config screen refactor, here are some arguments in favor of acting now
1. users are calling for it on a regular basis (or get confused that the setting is accessible in shift-p which is not intuitive).
2. the learning curve for users for the new design is very low (looks slightly different, completely equivalent in use, in particular the bounding box for mouse interaction is not changed).
3. this change should add no complication to said refactor intended to replace it with something better, later.

MeeniMc:
#2642 merged

PR #2297, opened 2024-01-31
Fix screen prompts not being keyboard-navigable

haykam821:
Previously, the buttons of screen prompts could not be navigated through the keyboard because they were being recreated upon each render. This pull request changes the behavior of screen prompts to only recreate these buttons when necessary, matching vanilla behavior.


jellysquid3 (approved):

jellysquid3:
The code for the screen prompt stuff is really ugly right now, but this is a fine stop gap solution.

haykam821:
Vanilla opens a screen that renders a prompt and the parent screen behind it. I think that is a good solution, but I didn't want to make such sweeping changes in this pull request.

PR #2016, opened 2023-08-16
Efficient CPU Translucency Sorting with BSP Trees and Heuristics

douira:
This PR implements efficient CPU translucency sorting. See https://hackmd.io/@douira100/sodium-sl-gfni for a slightly outdated but still useful overview of the concepts.

### Included in this PR
- Implementation of GFNI and direct (angle or distance) sort triggering
- Implementation of various sorting algorithms of increasing complexity (static normal relative, static topological, BSP tree, dynamic topological)
- User-facing debug info and option to turn it off or restrict to non-dynamic sorting only
- Index buffer rendering of translucent geometry
- Compatibility with Indium is implemented [in this branch](https://github.com/douira/Indium/tree/sodium-ts-gfni-compat)

Closes https://github.com/CaffeineMC/sodium-fabric/issues/38

### Non-goals
Translucency sorting on the GPU is certainly possible to be integrated with triggering with GFNI but not the focus of this PR. A first implementation of sorting on the CPU that works with GFNI is important to gauge performance of the triggering system and to enable the translucency sorting problem to be fixed without getting into the considerable effort of GPU sorting. Furthermore a CPU sorting implementation is needed in order to take advantage of the significant work-savings achievable by designating sections statically sorted.

My measurements indicate that triggering and sorting perform well, and usually beat distance-based sorting by either not needing to do any work, by doing the work more efficiently, or less often.

PR #2311, opened 2024-02-02
Migrate to official mappings

### LINKED ISSUE #2277: Migrate to Mojang ("official") mappings
Right now, the support for Yarn mappings in Architectury is not great, and it is likely to become worse as NeoForge looks to target the official mappings (dropping the utility classes for remapping in the process.)

While we don't have any particular issues with Yarn (and are very thankful for the additional documentation and parameter names it provides), it is ultimately going to provide less friction for us going forward to just use the official mappings, and will help prevent some divergence with other third-party forks.

We should try to complete this work alongside #2278 since it will be highly disruptive to existing pull requests and third-party projects.

Ideally, we would also want to use [Parchment](https://github.com/ParchmentMC/Parchment) for mappings, which combine the javadoc and parameter names of Yarn with the official mappings. Doing this does not introduce any compile/runtime compatibility issues since it only affects how the decompiled Minecraft sources are rendered.

embeddedt:
## Goal

This PR migrates the Sodium codebase to use official mappings. The remapping was largely automated using Mercury & MercuryMixin, with some manual fixups afterwards to fix compile errors. As such, imports may have been shuffled around compared to trunk.

This work resolves https://github.com/CaffeineMC/sodium-fabric/issues/2277.

Renaming Sodium classes to use Mojmap-inspired names is not planned for this PR, as it's a separate job from compiling against Mojmap. Such renames would greatly complicate verifying correctness of the remapping (see below).

## Testing

I have decompiled a built JAR from this commit and 0.5.8. All mod classes (that is, any non-`@Mixin` class) produce identical source files except `NativeImageHelper` (which is only different due to an accessor method name changing). Thus, extra manual verification should not be required for any of these.

Unfortunately, it appears that unlike Architectury's toolchain, Fabric Loom emits mapped names into mixin annotations. This means that all of the mixin classes have different bytecode, and my decompilation trick does not work to verify them. What I can say is that the mod appears completely functional in dev, so I am doubtful that there are any incorrectly remapped mixins.


embeddedt:
Mixins have been validated using the following procedure:

* Run the game with `-Dmixin.debug.export=true` and copy the contents of `.minecraft/.mixin.out/class` to separate directories for both 0.5.8 & this PR.
 You will also need to cause Mixin's auditing to run on startup and load all mixins. One option is to insert `MixinEnvironment.getCurrentEnvironment().audit();` somewhere in the mod initialization process.
* Decompile the classfiles by running Vineflower on each of the directories.
* Use the following command to strip `sessionId =` from the decompiled mixin annotations to avoid false diffs: `find sodium_upstream_decomp sodium_embedded_decomp -name \*.java | xargs sed -i '/sessionId = "/d'` (replace directory names as needed)
* Compare the two directories in your favorite diff tool (e.g. Meld), and validate that the final classes are identical aside from some accessor renames (which should be harmless).

Credit to @IThundxr for suggesting this approach.

jellysquid3 (approved):
Verified that everything is remapped correctly, including mixins.

jellysquid3:
We need to figure out how to migrate the branches for #2016 and #1851 to the official mappings before we can merge this.

embeddedt:
#2016 has been [migrated](https://github.com/douira/sodium-fabric/tree/gfni-mojmap-rename), and #1851 is trivial enough to either merge the commit into or simply redo on top of the Mojmap tree.

PR #2320, opened 2024-02-05
add tooltip flag requirements

SrRapero720:
How it looks without flags or impact
image

how it looks with flags or impact
image


jellysquid3:
Why is it necessary to indicate that a renderer update is needed? That's an internal only thing which basically only tells the renderer to not re-use the last frame's contents.

Most of these states for that matter don't seem useful outside of the asset reload & game restart.

SrRapero720:
Game restart and assets reload are the most relevant. But to not make more ~~complex~~ unnecessary extended logic to ignore render reload and render restart I keep these flags displayed in tooltip.

SrRapero720:
rebased + mojmapised.
Also i add a small (hardcoded) check for low ordinal flags (``REQUIRES_RENDERER_RELOAD`` and ``REQUIRES_RENDERER_UPDATE``)

PR #2326, opened 2024-02-10
Add support for custom fluid rendering

PepperCode1:
This pull request adds support for custom fluid rendering, specifically through `FluidRenderHandler#renderFluid`, which is added by Fabric API. These changes have been tested with Bumblezone, after modifying the jar to disable the mixin to Sodium's `FluidRenderer`.

### DIFF
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuilder.java
@@ -1,11 +1,15 @@
 package me.jellysquid.mods.sodium.client.render.chunk.compile.buffers;
 
 import me.jellysquid.mods.sodium.client.model.quad.properties.ModelQuadFacing;
+import me.jellysquid.mods.sodium.client.render.chunk.terrain.material.Material;
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.builder.ChunkMeshBufferBuilder;
+import net.minecraft.client.render.VertexConsumer;
 import net.minecraft.client.texture.Sprite;
 
 public interface ChunkModelBuilder {
     ChunkMeshBufferBuilder getVertexBuffer(ModelQuadFacing facing);
 
     void addSprite(Sprite sprite);
+
+    VertexConsumer asVertexConsumer(Material material);
 }
--- a/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuilder.java
+++ b/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuilder.java
@@ -11,5 +11,17 @@ public interface ChunkModelBuilder {
 
     void addSprite(Sprite sprite);
 
-    VertexConsumer asVertexConsumer(Material material);
+    /**
+     * <b>This method should not be used unless absolutely necessary!</b> It exists only for compatibility purposes.
+     * Prefer using the other methods in this class instead as they are more efficient.
+     *
+     * <p>Returns a {@link VertexConsumer} which adds geometry to this model builder using the given {@link Material}.
+     * The returned vertex consumer expects quads and requires the position, color, texture, light, and normal
+     * attributes to be provided for each vertex. The returned vertex consumer may be a reused object, so it must not be
+     * stored or cached in any way by the caller.
+     *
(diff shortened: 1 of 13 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -70,8 +70,11 @@
   ],
   "depends": {
     "fabricloader": ">=0.12.0",
-    "fabric-rendering-data-attachment-v1": ">=0.1",
-    "fabric-rendering-fluids-v1": ">=0.1"
+    "fabric-block-view-api-v2": "*",
```
jellysquid3, line 73: Why was the version constraint removed here?

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/FluidRenderer.java
```diff
@@ -59,6 +62,15 @@ public class FluidRenderer {
     private final ChunkVertexEncoder.Vertex[] vertices = ChunkVertexEncoder.Vertex.uninitializedQuad();
     private final ColorProviderRegistry colorProviderRegistry;
 
+    // These fields are set before invoking FluidRenderHandler#renderFluid and cleared afterwards.
```
jellysquid3: These fields should be moved into a separate mutable object (perhaps `FluidRendererContext`) and the `ThreadLocal` should store that instead.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/FluidRenderer.java
```diff
@@ -100,10 +112,72 @@ private boolean isSideExposed(BlockRenderView world, int x, int y, int z, Direct
...
+        currentWorld = world;
+        currentFluidState = fluidState;
+        currentBlockPos = blockPos;
+        currentOffset = offset;
+        currentMeshBuilder = meshBuilder;
+        currentMaterial = material;
+        currentHandler = handler;
+        CURRENT_RENDERER.set(this);
+
+        handler.renderFluid(blockPos, world, meshBuilder.asVertexConsumer(material), blockState, fluidState);
```
jellysquid3: Should be wrapped in try/finally to clear the `ThreadLocal` state.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/buffers/ChunkModelBuilder.java
```diff
@@ -1,11 +1,15 @@
...
 import me.jellysquid.mods.sodium.client.render.chunk.vertex.builder.ChunkMeshBufferBuilder;
+import net.minecraft.client.render.VertexConsumer;
 import net.minecraft.client.texture.Sprite;
 
 public interface ChunkModelBuilder {
     ChunkMeshBufferBuilder getVertexBuffer(ModelQuadFacing facing);
 
     void addSprite(Sprite sprite);
+
+    VertexConsumer asVertexConsumer(Material material);
```
jellysquid3: Should be renamed to `asFallbackVertexConsumer` so that we can discourage using this for anything other than when it's strictly necessary. I would also appreciate a brief Javadoc here explaining that a single instance is used, and that shared state is modified rather than constructing a new instance.

jellysquid3 (changes requested):
Some minor changes requested, but the premise looks good otherwise.

src/main/resources/fabric.mod.json
PepperCode1: There was a mistake and originally I felt it was inaccurate, so I will double check what minimum version of each module Sodium actually needs and add those instead of the wildcard.

PepperCode1: After double checking, the parts of `fabric-rendering-data-attachment-v1` that Sodium uses/expects are present in every version of it. The parts of `fabric-rendering-fluids-v1` that Sodium uses/expect only exist in version 2.0.0 and onward. All other dependencies are also correct.

src/main/java/me/jellysquid/mods/sodium/client/render/chunk/compile/pipeline/FluidRenderer.java
PepperCode1: As discussed, the default rendering has been moved to `DefaultFluidRenderer` and the context necessary to invoke it is now bundled in an inner class of `FluidRenderer`.

jellysquid3 (approved):

PR #2334, opened 2024-02-14
Add jitpack config making recent builds not fail

IThundxr:
- Adds a jitpack.yml to use open jdk 17 to build sodium
- Adds the maven-publish plugin so jitpack will run the right commands
- Purge's gradle cache before building a artifact that will be on jitpack

Requested by @MCRcortex so they can depend on the latest commit of sodium for their mods


jellysquid3:
We've tried to support JitPack multiple times in the past, but it is a highly unreliable service that can and does cause issues for people who depend on their auto-generated artifacts. For people who need *very tight* integration with Sodium, they should use a Git submodule + Gradle composite builds to track the latest commit.

PR #2350, opened 2024-02-20
[TS] Fix waterlogged glass panes

douira:
Fixes the sorting of waterlogged glass panes by changing the intersection detection that's run when it can't partition things. It now produces a list of "primary intersectors", quads that intersect with a large fraction of the quads in the current partition, and then puts those in a fixed-order partition tree node that always renders them after the other quads. This means that quads that intersect with lots of things will get rendered after everything else within that set of unpartitionable quads. It doesn't get into infinite recursion by not doing this if there's no non-primary intersector.


douira:
Before and after:
Screenshot 2024-02-20 at 01 09 09
Screenshot 2024-02-20 at 01 10 46

MeeniMc:
looks like you are re-merging the commits from your previous merged PR. You need to create a new branch on dev and cherry-pick the new commits to create a new PR.

douira:
I was thinking we could squash merge again but I'll just make a new branch.

PR #2352, opened 2024-02-21
[TS] Fix waterlogged glass panes and other TS-related follow-up work

douira:
Cherry-picked edition of https://github.com/CaffeineMC/sodium-fabric/pull/2350

Fixes the sorting of waterlogged glass panes by changing the intersection detection that's run when it can't partition things. It now produces a list of "primary intersectors", quads that intersect with a large fraction of the quads in the current partition, and then puts those in a fixed-order partition tree node that always renders them after the other quads. This means that quads that intersect with lots of things will get rendered after everything else within that set of unpartitionable quads. It doesn't get into infinite recursion by not doing this if there's no non-primary intersector.

Before and after:
Screenshot 2024-02-20 at 01 09 09
Screenshot 2024-02-20 at 01 10 46


douira:
I've amended this by refactoring I did to buffer and sort result handling. This more cleanly separates the buffers and the translucent sorting data that generates those buffers. It might also solve previously undiscovered concurrency issues. It now always frees buffers when they've been uploaded without keeping them around for the next time. This previously caused high memory usage. Instead, the OS will now have to deal with re-supplying us with buffers efficiently.

A custom buffer caching and eviction system might be better at this because we can keep buffers around for a certain time before freeing them. (It's not guaranteed that it's actually better at this than the OS is) I have code for this at https://github.com/douira/sodium-fabric/tree/gfni-cache-eviction but it sometimes leaks buffers or crashes since there's unresolved bugs and unimplemented features that are necessary to maintain the caching structure.

MeeniMc:
just tried it (no other mods) and it works as intended

douira:
I didn't anticipate needing to add another commit to this but we found a bug last week so it got added.

jellysquid3:
Code looks fine to me, and no regressions were observed in testing.

PR #2313, opened 2024-02-03
Fix cycling controls not playing a sound when activated using the keyboard

haykam821:
Helper methods of controls now contain functionality for playing a sound when changing the control's value to prevent inconsistency between code paths, which was apparent with cycling controls.

Fixes #2309

PR #2370, opened 2024-03-07
2313-mojmapped: Fix cycling controls not playing a sound when activated using the keyboard

MeeniMc:
This is the same as pr #2313 but updated for the mojmapped Sodium.

Commit authorship and attribution from @haykam821 has been preserved.


douira:
This seems like a pretty easy fix. Are there concerns with this? The code looks good to me.

MeeniMc:
rebased

douira:
Currently neither using the mouse nor the keyboard to change sliders makes a click noise, with and without this PR.

MeeniMc:
I won't be able to look at it before this we

douira:
merged into 0.7, it's not actually broken I just misunderstood.

PR #2388, opened 2024-03-24
Updated text of brightness slider tooltip

### LINKED ISSUE #2347: Brightness slider description is wrong and misleading
### Bug Description

Sodium shows the following tooltip when hovering over the Brightness slider:
"Controls the brightness (gamma) of the game."

This makes no sense because Minecraft's Brightness slider has nothing to do with gamma. It's instead a slider that changes lighting intensity according to user preference.

This text should be replaced with something like this example:
"Controls the lighting intensity in darker areas of the game. Well lit areas are not affected."

### Reproduction Steps

Go to Video Settings and hover your mouse over the Brightness slider.

### Log File

-

### Crash Report

-

pikachu0542:
Resolves #2347 by changing the tooltip of the brightness slider to use less misleading language. Changed the text to use the phrase "lighting intensity" to replace "brightness (gamma)", to make it more clear and less ambiguous about what is actually being changed.


MeeniMc:
the name of the option in the vanilla menu is "Brightness", correct or not, this is confusing to introduce a new name for the same thing.

douira:
Since this doesn't change the label but just the tooltip to give more explanation I think it's fine.

PR #2397, opened 2024-03-31
Attribute locations are not influenced by out locations, fix an offset of 1

thejudge156:
To my knowledge, when it comes to shader attribute binding points, 1, 2, 3, and 4, should probably be 0, 1, 2, 3. fragColor falls under a separate func for binding. So a 0 and 0 collision is not possible here. Very much a nitpick, but I wanted to make sure I am correct in this. 

Very much a nitpick, but could avoid a hassle with the recent development of new drivers, if there are any other areas that should be changed as well let me know.


jellysquid3 (approved):

jellysquid3:
Looks fine to me. Thanks for the patch. To my knowledge, there used to be another vertex shader input at binding 0, but it was removed some time ago and we never updated the other binding points.

Issue #2330, opened 2024-02-12
Minecraft classes (net.minecraft.util) should not be accessed from pre-launch


jellysquid3:
This causes undesirable class loading and appears to break some things. Another mod ([code](https://github.com/ishland/EarlyLoadingScreen/blob/eb0d89626d3fdf4edbe4191498e290cc83df4ff7/src/main/java/com/ishland/earlyloadingscreen/patch/SodiumOSDetectionPatch.java)) is implementing a rather massive workaround to prevent issues.

Fixing this will likely require us to rewrite a bit of code, but that's not so bad... probably.

PR #2403, opened 2024-04-05
Remove dependency on net.minecraft.Util from Prelaunch phase.

Cassunshine:
This should fix issue #2330.

All this change does is remove the dependency on net.minecraft.Util's OS functions, replacing them with basically the same thing in the new OsUtils class, using SystemUtils (from apache commons) instead. We could just as easily also use Java's built-in OS detection instead of apache, which is what Minecraft seems to do, if there's opposition to this way of doing it.

The only place this doesn't change the implementation is EarlyDriverScanner's call to showMessageBox, which still uses minecraft's utils. But considering immediately after that call the JVM is shut down, I don't think class load issues are really a big deal at that point...


jellysquid3:
If memory serves, we have our own implementation for opening URLs in the browser. It gets used for the code path where a user double-clicks on the JAR file. We can just use that code instead, which is more reliable, anyways.

jellysquid3:
Your patch looks fine to me otherwise. Thanks.

PR #2404, opened 2024-04-05
Add language file.

tomo25neko:
I have added a translation for Japanese.


tomo25neko:
I added Japanese display because I wanted it, and Sodium is popular in Japan, so I would be happy if you could accept this request.

Julienraptor01:
sodium doesn't accept translations rn, but you can come and help translate here
https://crowdin.com/project/sodium-fabric/ja

PR #2467, opened 2024-05-05
Reorient water quads as needed based on corner height to fix #609

### LINKED ISSUE #609: Sodium changes fluid rendering compared to vanilla
A water flow rendered by the current Sodium next branch:

The same water flow, rendered in vanilla:

I'm not sure if this is intended behavior or not.

douira:
the seam is placed such that it touches the highest or lowest corner

It was necessary to offset the inner water quads by 1 to align their seam with their upwards facing counterparts.

Fixes #609


MeeniMc:
LGTM, I tried some waterlogged stairs and other weird shapes and it always looked like ground truth.

jellysquid3:
Seems reasonable enough to me... Thanks for the patch.

PR #2468, opened 2024-05-05
Bugfix/0.6 water occlusion behind transparent blocks

MeeniMc:
When a transparent/translucent block surrounds a fluid, the fluid is considered to be occluded, however this is incorrect as the fluid can be seen through the transparent block. 

Issue present:

Issue resolved with this PR:

bug introduced in 0.6 branch during bc38d5638d391d3dfd1a928f6c7981116a7510f7 (not in dev)


IMS212:
Thanks; not sure how I didn't notice this.

PR #2477, opened 2024-05-06
bugifx: 0.6: The top face of fluids must not be occluded even if the block above h…

MeeniMc:
The top face of fluids must not be occluded even if the block above has a full-face bottom. The top fluid face is not coplanar with the bottom full-face, and thus can be seen from sideways, and the top side should be rendered so that underground aquifers and lava pockets render properly in spectator mode. 

Note how under the **solid block** (here, packed ice, #2468 fixed only the case for translucent blocks) the top face is occluded, but remains visible from the side air-blocks, or from translucent blocks on the side.

Bug: Underground water aquifers are invisible, same for lava pockets

This PR: returns to ground truth: 

Bug present only in 0.6 (not in dev) from https://github.com/CaffeineMC/sodium-fabric/commit/bc38d5638d391d3dfd1a928f6c7981116a7510f7

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/DefaultFluidRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/DefaultFluidRenderer.java
@@ -76,7 +76,7 @@ public class DefaultFluidRenderer {
         if (blockState.getFluidState().getType().isSame(fluid)) {
             return true;
         }
-        return blockState.canOcclude() && blockState.isFaceSturdy(world, adjPos, dir.getOpposite(), SupportType.FULL);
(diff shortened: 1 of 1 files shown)


IMS212:
Leaving this open for now; I suspect there are more major issues with this area.

MeeniMc:
I have found no other visual defects. Do you have a hunch about what else may be broken here?

From what I understand, the patch that introduced the errors main change is that the isFaceSturdy is now applying to the adjBlock rather than the fluid block itself. The top face of fluid is not sturdy so fluids would never be occluded (?). Now it checks the adjBlock, which is sturdy and so it occludes in more cases, which is correct, except for the UP face, for the spectactor case, it must never be occluded.

There may be some performance benefit in enabling top fluid face occlusion in more cases but it is somewhat hard to determine when its safe to do so.

IMS212:
Couldn't find anything, merging. Thanks!

Issue #1902, opened 2023-07-22
Block model cuboids and their textures are misaligned compared to vanilla


WeegeeTheDoggo:
### Bug Description

Expected: Block models and textures to line up as they normally do in vanilla

Actual: Block models and textures are slightly out of line

I believe this to be a bug. It does not happen with any other mods, or just by using Fabric itself, or in vanilla. I'm not aware of a setting that could remedy this, after looking through all the in-game settings for Sodium and testing each one. This bug affects an unknown number of block models, but I can verify it in end rods, torches, and fences. I expect that it affects many more.

Normal:
![2023-07-22_07 35 25](https://github.com/CaffeineMC/sodium-fabric/assets/47928239/74f9af8a-fe65-4e8c-ab22-ac58eb17f104)
Using Sodium:
![2023-07-22_07 33 26](https://github.com/CaffeineMC/sodium-fabric/assets/47928239/733ff24a-8cc1-4e95-9c32-c05cb7a00d9a)

### Reproduction Steps

- Make sure fabric-loader-0.14.21-1.20.1 is installed
- Download fabric-api-0.86.0+1.20.1 and sodium-fabric-mc1.20-0.4.10+build.27 and place them into the mods folder
- Launch Minecraft Launcher
- Select fabric-loader-0.14.21-1.20.1 profile
- Wait for game to load, then enter a world and look closely at block model (such as end rod) (use spectator mode if needed)

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/12136305/latest.log)

### Crash Report

[crash-2023-07-22_07.39.29-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/12136303/crash-2023-07-22_07.39.29-client.txt)

pajicadvance:
I can reproduce this on my system.

muzikbike:
Vanilla has its own alignment issues with models as well. (#1316)

I wonder if it'd be possible for Sodium to fix this issue and the vanilla cases as well (#1607). There exist mods such as Model Gap Fix and Model Fix Multi which resolve vanilla's issues with this, so integrating such a fix into Sodium as well so that more people can benefit would be a nice touch.

jellysquid3:
We can't really fix any of these vertex precision issues until the fluid renderer is rewritten to not abuse them. At that point, it should become possible to correct the off-by-one issue which affects texture coordinates, since it won't cause texture gaps in fluid rendering.

[...]

Issue #1316, opened 2022-05-27
Fix vanilla precision issues in block models


muzikbike:
### Version information

sodium-fabric-mc1.18.2-0.4.1+build.15

### Expected Behavior

All pixel sizes should be consistent and pixels and their borders should align perfectly.

### Actual Behavior

In a few select cases, there appear to be precision loss issues in model handling which result in ever so slightly incorrect sizes for pixels, model elements and so on. Examples are given below.

### Reproduction Steps

Doors ([MC-169897](https://bugs.mojang.com/browse/MC-169897)):
- Place a door
- Position yourself at the hinge portion (the region that stays constant when the door is closed and open)
- Open and close the door and observe closely

Cauldrons, hoppers and composters ([MC-246224](https://bugs.mojang.com/browse/MC-246224)):
- Place any or all of these three blocks
- Look very closely at all of the top corners

Farmland, dirt paths ([MC-252256](https://bugs.mojang.com/browse/MC-252256)):
- Position farmland or a dirt path next to a full block
- Try to see the second row of pixels from the top on the full block beside it

### Java version

17.0.2 64bit

### CPU

4× Intel® Core™ i3-3240 CPU @ 3.40GHz

### GPU

Intel® HD Graphics 2500

### Additional information

Door (hard to tell from screenshots, so I recommend reproducing this in game):
![2022-05-27_15 51 29](https://user-images.githubusercontent.com/52297970/170725679-22d0dc2c-2173-402a-a34a-06b968b8a532.png)
![2022-05-27_15 51 30](https://user-images.githubusercontent.com/52297970/170725684-6bd55746-e7d5-4fb1-b9e8-364378e76d73.png)

Cauldron, hopper, composter:
![2022-05-27_15 52 19](https://user-images.githubusercontent.com/52297970/170725724-d8804d9d-a33f-48e6-89e8-ab1afa1d18da.png)
![2022-05-27_15 52 23](https://user-images.githubusercontent.com/52297970/170725727-0ea984dc-cec3-4f1f-bf1a-7ab8b6a25535.png)

Dirt path, farmland:
![2022-05-27_15 52 49](https://user-images.githubusercontent.com/52297970/170725783-6a179c67-982c-4813-89b3-e24624af5411.png)

These effects can likely be found in far more circumstances.

muzikbike:
[...]

PR #2508, opened 2024-05-18
Optimizations to vanilla block models

### LINKED ISSUE #2460: Optimize some of vanilla's block models
### Request Description

Several blocks in vanilla use models that are very unoptimized, using far more texture planes than are strictly needed to achieve said appearance. We can get a visually identical result using less planes - the recent fix for issues like #1316 should eliminate any subpixel discrepancies that would otherwise exist in vanilla.

However, I'm not sure if including these optimized models in Sodium would jeopardise their implementation into vanilla for licensing reasons.

Affected models:
- Stairs (https://bugs.mojang.com/browse/MC-262461)
- Cauldrons (https://bugs.mojang.com/browse/MC-262470)
- Item frames (https://bugs.mojang.com/browse/MC-262527)
- Chorus flowers (https://bugs.mojang.com/browse/MC-262641)
- Hanging mangrove propagules (https://bugs.mojang.com/browse/MC-262689)
- Composters (https://bugs.mojang.com/browse/MC-267127)

The following fixes would also change texture mapping, which may be undesirable (even if vanilla's is wrong):
- Dragon egg (https://bugs.mojang.com/browse/MC-262652)
- Fence gates (https://bugs.mojang.com/browse/MC-262953)
- Fences (https://bugs.mojang.com/browse/MC-267281)
- Glass panes (https://bugs.mojang.com/browse/MC-267315)

Hoppers (https://bugs.mojang.com/browse/MC-262452) would require unused pixels to be removed from the top texture (https://bugs.mojang.com/browse/MC-262454). In addition, optimizing flower pots may require redoing how the texture works entirely, which might be bad news for mods that add their own potted objects based on the vanilla model (https://bugs.mojang.com/browse/MC-262427).

muzikbike:
Fixes #2460.

Models optimized so far:
- Stairs (straight and inner)
- Hoppers (down and horizontal)
- Composter
- Chorus flower
- Turtle eggs of stack size 4
- Candle cakes
- Anvil
- Beacon
- Item frame, with or without map
- Hanging lantern
- Hanging mangrove propagule

Models I'm still to do:
- Dragon egg
- Flower pots and potted objects
- Cauldron

See #2460 for explanations of these optimizations as well as performance comparisons.


mrjasonn:
Wouldn't doing ones that fix custom texture mapping break some custom resource packs which are adapted to the fact that they have broken texture mapping? If that is the case, shouldn't we just optimise it and remember to keep them looking the same as vanilla? (the looking same as vanilla should also be for everything in this PR)

mrjasonn:
And if we are using the attached resource packs in the Mojiras to optimise the models, shouldn't we have a credits thanking them for their work?

mrjasonn:
> ...and for the rest (item frames, propagules) nobody will notice the difference and if they do they won't care....

Please just make them look vanilla (I don't mind them looking different if clip in in spectator mode), as people might want them to actually look vanilla and not just ignore the fact it doesn't look vanilla. You have said that fixes for #1316 and #1902 may make this not visible, so please test on top of those commits. If it is harder to see, still best to keep vanilla in my opinion.

jellysquid3:
> And if we are using the attached resource packs in the Mojiras to optimise the models...

As far as I know, this is Muzikbike's original work. If it's not, then we need to know, since we require that all contributions be the submitter's original work...

jellysquid3:
The changes seem fine to me. The minor details you mentioned are very unlikely to cause issues. But, for us to import these into the repository, each model file should probably have documentation (in the form of a `__comment` line at least) explaining what changes were made from Vanilla.

jellysquid3:
@muzikbike Do you have any update on this? I'd like to merge it if you think it's ready. The only remaining blocks that don't appear to be addressed by this pull request are not so common in the world anyways (Dragon Eggs, Flower Pots, Cauldrons...), so the performance impact of improving them is very, very small.

djmrFunnyMan:
Apparently dragon eggs got optimized ([MC-262652](https://bugs.mojang.com/browse/MC-262652) was fixed) in the latest snapshot

muzikbike:
I'd prefer to finish optimizing the remaining blocks. Cauldrons are a major offender and show up in a good handful of farms, so those are a priority to fix.

mrjasonn:
@jellysquid3, if this gets done in time, would this be something that potentially gets added in Sodium 0.6?

jellysquid3:
Uhh... If the pull request was ready in the day or so, then yes, we could package it for Sodium 0.6.0. But the timeline on that is likely to be too close.

It's also not necessary to have all block models addressed in the initial pull request, if those need more time. We can always follow up with another pull request later down the road.

muzikbike:
I should be able to do cauldrons in a later PR if people want this in 0.6. What else needs done before I can consider this ready?

jellysquid3:
There is nothing blocking this pull request on our side. But while it is marked as a draft we cannot do anything with it.

jellysquid3 (changes requested):
There are many issues with inconsistent formatting and indentation in this pull request. Can you please run the files through a JSON formatter to normalize everything? We use 4-wide spaces in the project, so that should be used here too.

muzikbike:
Just threw everything into Blockbench and exported it. Does this suffice?

jellysquid3:
I've merged the changes manually with https://github.com/CaffeineMC/sodium/commit/d3aed21458560384abbac9e47dcaa096112aa910 to accommodate code style requirements, and because your branch was not tracking the multiloader changes.

Thank you for your efforts.

PR #2509, opened 2024-05-19
Force windows to use DGPU instead of IGPU when possible

MCRcortex:


mrjasonn:
If it doesn't, it needs to respect user manually set settings as people might force it to use the iGPU or another card. But it makes sense that we need Windows to use the dGPU as their "Windows chosen" instead.

jellysquid3:
There is no documentation with this pull request. I understand that you have spoken to me in the past about these techniques, but a detailed write up of what is happening still needs to be provided, given the code and pull request are completely undocumented.

mrjasonn:
> If it doesn't, it needs to respect user manually set settings as people might force it to use the iGPU or another card. But it makes sense that we need Windows to use the dGPU as their "Windows chosen" instead.

Thank you for following my above suggestions.

Issue #2422, opened 2024-04-14
Make custom version of the Vanilla Tweaks better fast leaves for programmer art that actually uses the original texture instead of using the oldest one available from Vanilla Tweaks


mrjasonn:
### Request Description

Please create a custom version of Vanilla Tweaks’ Better Fast Leaves for Programmer Art, using the original texture from Programmer Art to make it instead of the oldest texture for the leaves from Vanilla Tweaks. This will provide a more original and authentic look for users using Programmer Art and which most would probably appreciate.

This is something that has been improperly brought up a long time ago in the pull request that first added Better Fast Leaves for Programmer Art (#1862) that no one has done or brought up properly until now I'm bringing this up properly that this would be good to do.

djmrFunnyMan:
this is the sodium issue tracker (wrong place)

You can contact the vanilla tweaks team through their website https://vanillatweaks.net/feedback/

IMS212:
No, this is correct; Sodium includes textures sourced from Vanilla Tweaks.

jellysquid3:
I've compared the textures in Sodium right now and they're identical to the programmer art resource pack, so this looks to have been resolved.

PR #2512, opened 2024-05-20
Reworked leaves textures as well as adding waterlogging support.

TheNukeLore:
Remade the better leaf system from scratch, and alongside that, I included an experimental waterlogging system that helps people visually see waterlogged leaves with fast graphics.

What's currently included:

* Support for all leaves
* Waterlogged blockstates for leaves
* Programmer art leaves

Known and/or potential issues:

* Waterlogged leaves texture will likely override other resource packs texture unless they specifically add waterlogging support to said pack.
* Waterlogged programmer art leaves revert back to non-programmer art leaves due to the issue mentioned above.

Something to note: For packs to add support for waterlogged leaves, they will need to add a folder called "``waterlogged``" in "``textures/block/``". In that folder, they can add their own form of waterlogged leaves or simply copy their existing texture there if they don't care to support waterlogged leaves on fast graphics.

When it comes to models, most models should simply work, as they are only replacing the texture. There is a niche case where it could still break, but it is extremely unlikely.

If requested, I may remove waterlogged support due to the issues that it can cause.


TheNukeLore:
These changes:

Current sodium(uses vanilla tweaks leaves): 

There is partial support for programmer art leaves. however they have a few issues when it comes to water-logging, may look back into that issue as it also affects external packs. image showing the issue at hand: 

Should be noted that non of my changes copy or use assets from Vanilla Tweaks, these are all manually made using the base leaf textures.

jellysquid3:
I'm not okay with breaking resource pack compatibility for adding waterlogged textures into the game. For the most part, we don't even encourage using _Fast Graphics_ mode since it offers little to no performance improvement and introduces all kinds of issues with texture handling.

jellysquid3:
That said, can you please elaborate on how you "improved" the existing textures? I don't see much of a difference from the screenshots given. There are bugs like #2422 but I don't actually understand what is described or if your pull request fixes this.

TheNukeLore:
My textures are not inherently better just done in a consistent method, They were more so made in a methodology that avoided copying any part of vanilla tweaks to avoid any potential concerns of copying(cleanroom per say), however the programmer art leaves should be the most updated ones possible, they uses the latest assets from programmer art in 1.20.6. I can update this pr with the removal of water logging blockstates/textures if you still would like the other aspects.

jellysquid3:
If you can create a new pull request with just the changes that update the Leaves textures, that would be appreciated. Let's not change the purpose of this pull request in case it's useful in the future.

TheNukeLore:
That includes both regular and programmer art textures correct?

jellysquid3:
Yes, both sets of textures.

jellysquid3:
There seems to be no tangible difference between the resources in this pull request and the ones in trunk right now. And we are not interested in adding waterlogged textures for Fast graphics, since this would technically be new functionality, and would change aesthetics significantly.

PR #2513, opened 2024-05-20
Replace sodiums fast leaves

### LINKED ISSUE #2422: Make custom version of the Vanilla Tweaks better fast leaves for programmer art that actually uses the original texture instead of using the oldest one available from Vanilla Tweaks

TheNukeLore:
This pull request started out as: #2512. However this pull request only updates the existing textures in sodium.
It potentially fixes: #2422 as well however I am not entirely sure what the original issue is suggesting.


mrjasonn:
I have made #2422 potentially more clear. Also this does not fix #2422, it's just adding extra bulk to the files. Also a good thing to mention in regards to this adding more bulk to the files, Vanilla Tweaks has granted access to Sodium for the leaves textures for Sodium to use so it might something we can be less concerned about of copying the textures. If you are still unclear about #2422, please ask.

jellysquid3:
Closing per my comments in #2512.

PR #2523, opened 2024-05-30
Port to Minecraft 1.21-pro1

meglinge:


IMS212:
Sadly, the fix provided is not correct; I'll be updating to pre1 soon with a different fix for the entity renderer.

PR #2536, opened 2024-06-12
Disable rendering clouds if texture is completely blank

Moulberry:
Some resourcepacks/maps may disable clouds by blanking out the texture. This PR introduces a simple check to skip cloud rendering when this is the case


IMS212:
Thanks!

PR #2563, opened 2024-06-20
Do not call sodium free and open source

blt-r:
In the donation message it says "free and open source" which sodium no longer is.

### DIFF
--- a/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumOptionsGUI.java
+++ b/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumOptionsGUI.java
@@ -442,7 +442,7 @@ public class SodiumOptionsGUI extends Screen implements ScreenPromptable {
     static {
         DONATION_PROMPT_MESSAGE = List.of(
                 FormattedText.composite(Component.literal("Hello!")),
-                FormattedText.composite(Component.literal("It seems that you've been enjoying "), Component.literal("Sodium").withColor(0x27eb92), Component.literal(", the free and open-source optimization mod for Minecraft.")),
+                FormattedText.composite(Component.literal("It seems that you've been enjoying "), Component.literal("Sodium").withColor(0x27eb92), Component.literal(", the optimization mod for Minecraft.")),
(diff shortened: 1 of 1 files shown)


Lolothepro:
It's still free and open source

jellysquid3:
The suggested edit makes no grammatical sense.

Gaming32:
It's free, but officially Polyform Shield isn't an open-source license (I think).

douira:
The OSI didn't approve it as such, but that [doesn't mean you can't nevertheless call it that](https://opensource.stackexchange.com/questions/895/does-a-license-needs-to-be-approved-by-the-osi-to-be-an-open-source-license). One way or another, this PR doesn't make any sense.

blt-r:
It's not about being approved by OSI or someone else. If the license doesn't allow to fork the software, it is not free and open source license.

jellysquid3:
The conversation about whether it's free and open source is not relevant. We've already edited out mentions of it from other places. But the suggested edit here does not make any sense and that's why it was closed.

blt-r:
So this mention of Sodium being foss will also be removed? You just don't like the way I changed the statement in this PR, am I understanding correctly?

jellysquid3:
Yes. We updated this across our online webpages and the project documentation [a long time ago](https://github.com/CaffeineMC/sodium-fabric/commit/d70c1042c7e4fb01388d53ac4871ce124dbee763#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5). The flavor text in the donation prompt will be edited, it was just missed.

davepusey:
How is it not free and open source? There is zero cost to be able to use it, and the source is literally right here in this repo.

jellysquid3:
The FSF has (for most intents and purposes) defined the "free and open source" phrase, and the Polyform Shield license is not blessed by them. This is largely because the license places limitations on _how_ you can use the software (i.e. you aren't allowed to compete with us, using our own code.) 

While this does not matter for _essentially anyone_ but the people who re-upload our mod with a few small changes, a lot of people view it as some kind of infringement on their personal liberties. And frankly, it's not worth arguing with open-source evangelists about whether the software is _truly_ "free and open source", so we just avoid the terminology entirely.

PR #2588, opened 2024-07-06
Do not refer to sodium as 'free and open-source'

Un1q32:
Sodium is not free or open-source according to the FSF or OSI definitions, referring to sodium as FOSS may cause confusion for people who expect that to mean they are given certain rights not guaranteed by sodium's current license.  It also could be seen as deceptive to have this in a message requesting donations, as some people might not have donated if they had known sodium used a more restrictive license then is implied.


jellysquid3:
See #2563.

Un1q32:
#2563 was closed for not making gramatical sense, this does.  If you have a preferred way of saying it then please tell me, as long as it doesn't incorrectly refer to sodium as free and open source.

mrmangohands:
It's a small piece of flavor text that was missed on the first pass, the maintainers will decide on a new wording and push it, this isn't really an area outside contribution is needed

PR #2599, opened 2024-07-14
Disable Cloud Rendering While in Lava

Decencies:
Disables cloud rendering when the camera is in lava fluid.


Decencies:
My apologies, I don't know what the fuck git was doing, but a proposed fix is there for the [clouds.fsh](https://github.com/CaffeineMC/sodium-fabric/blob/4514388536fbc4ac6e570ad31c06793c5b66acb1/src/main/resources/assets/minecraft/shaders/core/clouds.fsh#L27) file

douira:
Is this a good solution to the problem? Are there actionable better solutions?

PR #2600, opened 2024-07-15
Update Actions, Gradle and Deps

CanerKaraca23:


IMS212:
This is not fully updated; there is a more efficient way to setup Gradle in Github Actions. I will make a commit soon to add this.

CanerKaraca23:
> This is not fully updated; there is a more efficient way to setup Gradle in Github Actions. I will make a commit soon to add this.

So, what about other changes? At least you can edit PR and wait, why immediately close.

IMS212:
I have applied all the changes + a new Gradle action as of a few minutes ago.

CanerKaraca23:
> I have applied all the changes + a new Gradle action as of a few minutes ago.

Some action versions left, like upload.

PR #2604, opened 2024-07-16
Fix delay with FREX Flawless Frames

Moulberry:
Fix delay with FREX Flawless Frames due to dependency between building & visibility.

In order for a chunk to be built, it must be visible.
In order for a chunk to become visible via the graph search, it must have a neighbor that is built.

This dependency results in the graph search propagating visibility one section at a time. Normally this would be unnoticeable, but when using FREX Flawless Frames we want the entire world to be visible ASAP.

The solution in this PR is to simply to do building & visibility in a loop until there are no chunks left to build (but only when flawless frames is on)

This has also been ported to 1.20.6 here, might be useful as a reference if doing another 1.20.6 release: https://github.com/Moulberry/sodium-fabric/tree/fix_flawless_frames_delay/1.20.6

### DIFF
--- a/src/main/java/net/caffeinemc/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/src/main/java/net/caffeinemc/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -206,9 +206,9 @@ public class SodiumWorldRenderer {
             this.lastCameraPos = new Vector3d(pos);
         }
 
-        boolean runChunkUpdates = true;
+        int maxChunkUpdates = updateChunksImmediately ? this.renderDistance : 1;
 
-        while (runChunkUpdates) {
+        for (int i = 0; i < maxChunkUpdates; i++) {
             if (this.renderSectionManager.needsUpdate()) {
                 profiler.popPush("chunk_render_lists");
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/caffeinemc/mods/sodium/client/render/SodiumWorldRenderer.java
```diff
@@ -207,20 +206,26 @@ public void setupTerrain(Camera camera,
             this.lastCameraPos = new Vector3d(pos);
         }
 
-        if (this.renderSectionManager.needsUpdate()) {
-            profiler.popPush("chunk_render_lists");
+        boolean runChunkUpdates = true;
 
-            this.renderSectionManager.update(camera, viewport, frame, spectator);
-        }
+        while (runChunkUpdates) {
```
IMS212: Is there anything stopping this from timing out? It should not be able to loop forever.

Moulberry: There isn't right now.
I could potentially use `this.renderDistance` as an upper bound on the number of iterations
It shouldn't need more than that in order to visit everything, right?

IMS212 (approved):
Looks reasonable to me.

Issue #2591, opened 2024-07-07
Crash when a graphics adapter is present which does not support OpenGL


D2Lx0wse:
### Bug Description

The game crashes with this error
`A mod crashed on startup!
net.fabricmc.loader.impl.FormattedException: java.lang.RuntimeException: Could not execute entrypoint stage 'preLaunch' due to errors, provided by 'sodium'!
	at net.fabricmc.loader.impl.FormattedException.ofLocalized(FormattedException.java:63)
	at net.fabricmc.loader.impl.launch.knot.Knot.init(Knot.java:162)
	at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:68)
	at net.fabricmc.loader.impl.launch.knot.KnotClient.main(KnotClient.java:23)
	at org.prismlauncher.launcher.impl.StandardLauncher.launch(StandardLauncher.java:100)
	at org.prismlauncher.EntryPoint.listen(EntryPoint.java:129)
	at org.prismlauncher.EntryPoint.main(EntryPoint.java:70)
Caused by: java.lang.RuntimeException: Could not execute entrypoint stage 'preLaunch' due to errors, provided by 'sodium'!
	at net.fabricmc.loader.impl.FabricLoaderImpl.lambda$invokeEntrypoints$2(FabricLoaderImpl.java:388)
	at net.fabricmc.loader.impl.util.ExceptionUtil.gatherExceptions(ExceptionUtil.java:33)
	at net.fabricmc.loader.impl.FabricLoaderImpl.invokeEntrypoints(FabricLoaderImpl.java:386)
	at net.fabricmc.loader.impl.launch.knot.Knot.init(Knot.java:160)
	... 5 more
Caused by: java.lang.NullPointerException: Cannot invoke "String.matches(String)" because "driverName" is null
	at me.jellysquid.mods.sodium.client.compatibility.checks.PreLaunchChecks.findIntelDriverMatchingBug899(PreLaunchChecks.java:123)
	at me.jellysquid.mods.sodium.client.compatibility.checks.PreLaunchChecks.onGameInit(PreLaunchChecks.java:60)
	at me.jellysquid.mods.sodium.client.SodiumPreLaunch.onPreLaunch(SodiumPreLaunch.java:13)
	at net.fabricmc.loader.impl.FabricLoaderImpl.invokeEntrypoints(FabricLoaderImpl.java:384)
	... 6 more
`

### Reproduction Steps

[...]

PR #2613, opened 2024-07-23
Fix issue #2591 by ignoring null driver names

ading2210:
This is a fairly trivial fix for #2591.

The function this is in seems to only be for detecting Intel drivers (not applicable for Adreno cards), so this shouldn't break any existing behavior.


jellysquid3:
I'm fine with merging this, but I'd like for those fields to be documented with `@Nullable` annotations correctly so that adding this doesn't create compiler warnings.

IMS212 (approved):
Looks good to me.

Issue #2601, opened 2024-07-15
Custom animated sprites are frozen


kikugie:
### Bug Description

In my mod Elytra Trims I have an animated texture, which is then rendered on top of the elytra model (see links for code references). In vanilla it renders correctly, where as with Sodium it freezes at a single frame. Also reloading resource packs with F3+T advances the frozen frame.

(Copied from the [Discord thread](https://discord.com/channels/602796788608401408/1261424852342931579) as requested)

1. [Texture file and metadata](<https://github.com/kikugie/elytra-trims/tree/kotlin/src/main/resources/assets/elytratrims/textures/animation>)
2. [Texture loading code](<https://github.com/kikugie/elytra-trims/blob/c77d0f6e79fabcd3d8d2d1c9bdc4ab9d1dccbdce/src/main/kotlin/dev/kikugie/elytratrims/client/resource/ETAtlasHolder.kt#L116>)
3. [Texture rendering code](<https://github.com/kikugie/elytra-trims/blob/c77d0f6e79fabcd3d8d2d1c9bdc4ab9d1dccbdce/src/main/kotlin/dev/kikugie/elytratrims/client/render/FeatureRenderer.kt#L51>)

### Reproduction Steps

Mod list:
- Minecraft 1.21
- Elytra Trims 3.4.1
- Sodium 0.5.11
- Fabric API 0.100.7
- Fabric Language Kotlin 1.11.0

Join a singleplayer world. Add the animated texture by combining elytra with an apple and a rotten flesh in a crafting grid. The texture will appear frozen.  
Close Minecraft, disable Sodium and rejoin the world - the texture will be animated.

### Log File

[latest.log](https://github.com/user-attachments/files/16236104/latest.log)

### Crash Report

[crash-2024-07-15_16.34.05-client.txt](https://github.com/user-attachments/files/16236110/crash-2024-07-15_16.34.05-client.txt)

douira:
Does the proposed solution in https://github.com/CaffeineMC/sodium/pull/3084 fix this?

PR #2616, opened 2024-07-23
Patch for sprites not being marked as active when used in a SpriteCoordinateExpander 

MCRcortex:
Fix for https://github.com/CaffeineMC/sodium-fabric/issues/2601 class of issues, however it assumes that a new SpriteCoordinateExpander is created every frame. Marking it as active in the hot path every write might have performance implications, thus only a partial fix.


jellysquid3:
I'm not really sure this is robust enough. At least, if you create a `VertexBuffer` from a `BufferBuilder`, and render that, it won't track the sprites. Can we perhaps put something into `BufferBuilder.BuiltBuffer` to track the sprites and have them be updated each time the buffer is rendered?

PR #2617, opened 2024-07-23
Compare perspective matrix when marking graph as dirty

MCRcortex:


douira:
Is it still necessary to compare the camera angle (and position too?) if the projection matrix is compared?

MCRcortex:
Projection matrix doesnt contain the camera transform

douira:
Ok, then lgtm.

jellysquid3 (approved):
This change looks fine to me. But in the future, we should look at creating a record object to hold this state so that we don't have this complex if-statement.

IMS212 (approved):
Looks good to me.

Issue #2586, opened 2024-07-05
Sodium causing clientside chunks leak


frosthuuuund:
### Bug Description

The Sodium installed causing clientsided chunks leak either by itself or makes it much noticeable.
Below i attached screenshot which illustrates the problem.
![image](https://github.com/CaffeineMC/sodium-fabric/assets/174178570/7dc40f20-7b69-4737-84ed-114fd8e71604)

The amount of chunks on vanilla is a bit above from the starting point, but fairly enough to not set off the alarm.
![image](https://github.com/CaffeineMC/sodium-fabric/assets/174178570/8d542c59-ccd9-4af7-975f-d612e8dedcf0)

### Reproduction Steps

Install 1.21 with Fabric Mod Loader.
Use `jdk-21.0.3.9-hotspot` as your runtime.
Install Sodium 0.5.11.
Launch the game and create the new world with Creative mode.
Set the game rule keepInventory to true.
Equip elytra and firework.
Set a spawnpoint on `0 320 0`.
Open F3 menu and follow one axis both in positive and negative direction on Gizmo debug crosshair.
Fly around 1000 blocks and kill yourself with the console command.
Do the same for the all axes.
Keep an eye on amount of geometry pool

### Log File

https://paste.nolog.cz/?9a65e0dee4dddbf2#FdqxtsRo6zU2XyumZ9ZdM1jHRjVyBWyLbduZezS9er4x

### Crash Report

https://paste.nolog.cz/?4daec5abb631e610#CbPBK2LbpBALLJ2WyPgmrXYqtFm7VHuTSM7DFgLrkPnr

jellysquid3:
I think this problem is related to a regression in Minecraft 1.21 itself, as the chunks don't seem to get unloaded there either. Unfortunately, it would seem Sodium exacerbates the issue, because it will only unload geometry when the chunks themselves are unloaded, and not when they fall out of the render distance.

We might need to rework things to address this.

frosthuuuund:
Alright after the additional testing i got next information:
1. This definitely happens on vanilla. If you teleports and kills yourself enough, you'll get OOM error, with 4GB allocated, you will run out of memory pretty fast.
2. The dev snapshot showing fair amount of clientsided chunks, but still showing leaks in the geometry pool.
3. This happens on 1.20.1 with Sodium 0.5.11 (Java 17) and also 1.20.4 (but Java 21, probably won't make a difference).

douira:
[...]

Issue #2584, opened 2024-07-01
Chunks disappear on death resulting in longer loading terrain screen


MrKinau:
### Bug Description

When dying on a Paper server, the chunks immediately disappear, so you can't see them in the death screen. Additionally, since the chunks are no longer rendered, the loading terrain screen that appears after clicking the Respawn button takes much longer.

Video: https://youtu.be/eYxW55Ov5tg
In the video you can see: First part is played with Sodium 0.5.11 1.21 on a Paper 1.21 server, second part is played without Sodium on the same server.

### Reproduction Steps

1. Join a Paper server (tested with 1.21, but it should be the same with older versions)
2. Kill yourself (e.g. `/damage @s 100 minecraft:generic_kill`)
3. You are not able to see the world in death screen and respawning takes much longer than without sodium

### Log File

https://mclo.gs/7fgwszG

### Crash Report

https://mclo.gs/go98XWm

jellysquid3:
This bug appears to be a regression in Sodium for Minecraft 1.21. The chunk tracking code might need to be updated to account for other changes in the game.

jellysquid3:
This is potentially related to #2586.

jellysquid3:
This is a bug with Paper. They are sending chunk unload packets immediately after the player dies, which causes Sodium to unload them. This is why the chunks disappear behind the death screen, and why the loading screen is shown after respawning (as it is waiting for chunks to be sent back to the client.)

Fundamentally, there is no good way to fix this on our side. The only reason it works correctly in Minecraft is because the renderer isn't even aware of when chunks get unloaded. So it just _never does so_ until new chunks overwrite the existing chunks (or they are too far away from the camera). This behavior is frankly terrible, and is responsible for many graphical bugs and memory leaks, hence why Sodium does not (and will not) mimic it.

The best thing we can do here is at least _try_ to workaround Paper sending the chunk unload packets too early, so they don't disappear on the death screen. But this will not prevent the player from seeing a (short) loading screen when they spawn in the world again.

PR #2619, opened 2024-07-24
Fix memory leaks when server doesnt send chunk forget packets

MCRcortex:
Fix for #2586 (does not fix or change #2584 as that is a "bug"/behavior difference in vanilla and paper servers and requires a more complex patch)


IMS212 (approved):
There needs to be something better... but this will probably do for now.

douira:
Can the referenced issue https://github.com/CaffeineMC/sodium-fabric/issues/2586 be closed since this is merged?

djmrFunnyMan:
Jelly said no
Screenshot_20240813_132705_Discord

PR #2639, opened 2024-08-06
Merge Multiloader into dev

IMS212:
take two


neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/features/model/WeightedBakedModelMixin.java
```diff
@@ -0,0 +1,79 @@
...
+
+        return weighted;
+    }
+
+    /**
+     * @author JellySquid
+     * @reason Avoid excessive object allocations
+     */
+    @Overwrite
+    public List<BakedQuad> getQuads(@javax.annotation.Nullable BlockState state, @javax.annotation.Nullable Direction face, RandomSource random, ModelData modelData, @org.jetbrains.annotations.Nullable RenderType renderType) {
```
jellysquid3, line 53: We should look at sending a patch to NeoForge so they aren't using multiple implementations of `@Nullable`.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/model/quad/BakedQuadMixin.java
```diff
@@ -0,0 +1,134 @@
...
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(BakedQuad.class)
+public class BakedQuadMixin implements BakedQuadView {
```
jellysquid3, line 19: How is this different from the Fabric mixin?

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/ResourcePackLoaderMixin.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+import java.util.Set;
+
+@Mixin(ResourcePackLoader.class)
+public class ResourcePackLoaderMixin {
+    @Unique
+    private static final IModFile SODIUM_FILE = LoadingModList.get().getModFileById("sodium").getFile();
+
+    @Redirect(remap = false, method = "packFinder", at = @At(value = "INVOKE", target = "Ljava/util/Map;entrySet()Ljava/util/Set;"))
+    private static Set<Map.Entry<IModFile, Pack.ResourcesSupplier>> changeSet(Map<IModFile, Pack.ResourcesSupplier> instance) {
```
jellysquid3, line 29: This looks to be a patch for mods that also overwrite the leaves texture. Can you document the reasoning here for future reference?

common/src/main/java/net/caffeinemc/mods/sodium/client/services/SodiumModelData.java
```diff
@@ -0,0 +1,8 @@
+package net.caffeinemc.mods.sodium.client.services;
+
+/**
+ * Template class for the platform's model data.
+ */
+public interface SodiumModelData {
```
jellysquid3, line 6: This should be documented to explain that it's used to shadow the Forge types in a Fabric environment.

neoforge/src/main/resources/META-INF/neoforge.mods.toml
```diff
@@ -0,0 +1,43 @@
+modLoader = "javafml"
+loaderVersion = "*"
+license = "Polyform-Shield-1.0"
```
jellysquid3: Should be `Polyform-Shield-1.0.0`.

neoforge/src/main/resources/META-INF/neoforge.mods.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[[mods]]
+modId = "sodium"
+
+version = "${version}"
+displayName = "Sodium"
+
+logoFile = "sodium-icon.png" #optional
+
+authors = "JellySquid"
```
jellysquid3: We should also include a secondary authors field (maybe credits?) so that our contributors are listed.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/ClientHooksMixin.java
```diff
@@ -0,0 +1,17 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(ClientHooks.class)
+public class ClientHooksMixin {
+    // TODO: This is not a proper fix at all. Figure out why Forge is causing the UV shrink ratio to go insane.
+    @Redirect(method = "fixItemModelSeams", at = @At(value = "INVOKE", target =
+            "Lnet/minecraft/client/renderer/texture/TextureAtlasSprite;uvShrinkRatio()F"))
+    private static float alterUvShrinkRatio(TextureAtlasSprite sprite) {
```
jellysquid3, line 16: Does this still need to be investigated before release?

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelAccess.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public void renderAdditionalRenderers(List<?> renderers, Function<RenderType, VertexConsumer> typeToConsumer, LevelSlice slice) {
+        AddSectionGeometryEvent.SectionRenderingContext context = new AddSectionGeometryEvent.SectionRenderingContext(typeToConsumer, slice, emptyStack.get());
+        for (Object o : renderers) {
+            ((AddSectionGeometryEvent.AdditionalSectionRenderer) o).render(context);
+        }
+    }
+
+    @Override
+    public @Nullable Object getLightManager(LevelChunk chunk, SectionPos pos) {
```
jellysquid3: Can the type erasure be avoided here?

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelAccess.java
```diff
@@ -0,0 +1,62 @@
...
+import net.neoforged.neoforge.client.ClientHooks;
+import net.neoforged.neoforge.client.event.AddSectionGeometryEvent;
+import org.jetbrains.annotations.Nullable;
+import org.joml.Matrix4f;
+
+import java.util.List;
+import java.util.function.Function;
+
+public class NeoForgeLevelAccess implements PlatformLevelAccess {
+    private static final ThreadLocal<PoseStack> emptyStack = ThreadLocal.withInitial(PoseStack::new);
```
jellysquid3: Why does this need to be cached in a thread local? The allocation rate is only once per chunk mesh event. Keeping it around increases the risk that a mod could break the pose stack persistently between chunks and would make debugging harder. And if the stack gets very large, that memory will be kept around.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/block/NeoForgeBlockAccess.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.world.level.BlockAndTintGetter;
+import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.material.FluidState;
+import net.neoforged.neoforge.client.model.data.ModelData;
+import org.jetbrains.annotations.Nullable;
+
+public class NeoForgeBlockAccess implements PlatformBlockAccess {
+    private static final TriState[] TRI_STATES = new TriState[] {
```
jellysquid3: Is `TriState.of` not sufficient? If there is some platform difference here, let's map each platform's tri-state to a special `ShadeMode` or such.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/NeoForgeInfoAccess.java
```diff
@@ -0,0 +1,46 @@
...
+        return FMLPaths.GAMEDIR.get();
+    }
+
+    @Override
+    public Path getConfigDirectory() {
+        return FMLPaths.CONFIGDIR.get();
+    }
+
+    @Override
+    public boolean isFlawlessFramesActive() {
```
jellysquid3: Why is this in platform-specific code? The Fabric mod initializer should scan the mods and set up the global Flawless Frames consumers.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformInfoAccess.java
```diff
@@ -0,0 +1,46 @@
+package net.caffeinemc.mods.sodium.client.services;
+
+import java.nio.file.Path;
+
+public interface PlatformInfoAccess {
```
jellysquid3: "PlatformInfo" is a big vague here, it's more about the runtime environment info.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
```diff
@@ -0,0 +1,79 @@
...
+import net.minecraft.core.SectionPos;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.chunk.LevelChunk;
+import org.jetbrains.annotations.Nullable;
+import org.joml.Matrix4f;
+
+import java.util.List;
+import java.util.function.Function;
+
+public interface PlatformLevelAccess {
```
jellysquid3, line 9: This should be split into render-specific functionality (`PlatformLevelRenderHooks`, maybe) and world-query functionality (what the file is currently labeled.)

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    void runChunkLayerEvents(RenderType renderLayer, LevelRenderer levelRenderer, Matrix4f modelMatrix, Matrix4f projectionMatrix, int ticks, Camera mainCamera, Frustum cullingFrustum);
+
+    /**
+     * Returns any NeoForge chunk renderers to run. <b>This is not thread safe.</b>
+     * @param level The current level
+     * @param origin The origin of the current chunk
+     * @return Any NeoForge chunk renderers to run
+     */
+    List<?> getExtraRenderers(Level level, BlockPos origin);
```
jellysquid3: Renderers for what?

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
```diff
@@ -0,0 +1,79 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Creates a new platform dependent fluid renderer.
+     * @param colorRegistry The current color registry.
+     * @param lightPipelineProvider The current {@code LightPipelineProvider}.
+     * @return A new fluid renderer.
+     */
+    FluidRenderer createPlatformFluidRenderer(ColorProviderRegistry colorRegistry, LightPipelineProvider lightPipelineProvider);
```
jellysquid3: Maybe pull this out into a separate `FluidRendererProvider` interface.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Should be run in the top of the Vanilla fluid renderer. If true, cancel the Vanilla render.
+     * @return if the vanilla rendering should be cancelled.
+     */
+    boolean tryRenderFluid();
```
jellysquid3: Why does this exist? Is there no way to get it out of the platform interface? It's incredibly fragile.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    List<?> getExtraRenderers(Level level, BlockPos origin);
+
+    /**
+     * Runs any NeoForge chunk renderers.
+     * @param renderers The list of chunk renderers to run.
+     * @param typeToConsumer A consumer that converts render types to vertex consumers
+     * @param slice The current level slice
+     */
+    void renderAdditionalRenderers(List<?> renderers, Function<RenderType, VertexConsumer> typeToConsumer, LevelSlice slice);
```
jellysquid3: This really needs a better name. Maybe `collectAdditionalChunkGeometry` or something.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformTextureAccess.java
```diff
@@ -0,0 +1,19 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Finds the current sprite in the block atlas ({@code TextureAtlas.LOCATION_BLOCKS})
+     * @param texU The U coordinate of the texture.
+     * @param texV The V coordinate of the texture.
+     * @return The sprite in the location in the block atlas.
+     */
+    TextureAtlasSprite findInBlockAtlas(float texU, float texV);
```
jellysquid3: I think we should instead provide a `SpriteFinder getSpriteFinderForBlockAtlas()` or something, and have each platform provide an implementation which supports queries. Why is the sprite finder platform-specific, anyways?

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformBlockAccess.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    boolean shouldShowFluidOverlay(BlockState block, BlockAndTintGetter level, BlockPos pos, FluidState fluidState);
+
+    /**
+     * Gets the specialized render data for this block entity.
+     * @param blockEntity The block entity to get the render data of.
+     * @return The specialized render data for this block entity. If the platform does not support it or there is no data, null.
+     */
+    @Nullable
+    Object getBlockEntityData(BlockEntity blockEntity);
```
jellysquid3: This should probably be moved into the platform code for accessing level data.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformBlockAccess.java
```diff
@@ -0,0 +1,91 @@
...
+    boolean platformHasBlockData();
+
+    /**
+     * Gets the block shade.
+     * @param quad The quad being rendered.
+     * @param level The level.
+     * @param shade If directional lighting should be added.
+     * @return the block shade
+     */
+    float getAccurateBlockShade(ModelQuadView quad, BlockAndTintGetter level, boolean shade);
```
jellysquid3: Why is it accurate? Is this render code or level code?

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
```diff
@@ -337,23 +335,15 @@ private static void emitCellGeometry2D(VertexBufferWriter writer, int faces, int
             final long buffer = stack.nmalloc(4 * ColorVertex.STRIDE);
 
             long ptr = buffer;
-            int count = 0;
 
-            // -Y
-            if (CloudFaceSet.contains(faces, CloudFace.NEG_Y)) {
-                int mixedColor = ColorMixer.mul(color, CloudFace.POS_Y.getColor());
+            int mixedColor = ColorMixer.mul(color, CloudFace.POS_Y.getColor());
```
jellysquid3: Why are these faces always rendered now? If the goal is to avoid an empty buffer, this is not a good way to do it, since it completely eliminates our optimizations.

Worst case, we can update the method signature to return a boolean indicating whether geometry was added to the buffer.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/material/MaterialViewImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public boolean disableDiffuse() {
+        return (bits & DIFFUSE_FLAG) != 0;
+    }
+
+    @Override
+    public TriState ambientOcclusion() {
+        return TRI_STATES[(bits & AO_MASK) >>> AO_BIT_OFFSET];
```
jellysquid3, line 108: Why not use `TriState.of(bool)` here?

jellysquid3 (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
IMS212: Would implementing this in the FluidRenderer itself be smart?

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformBlockAccess.java
IMS212: This uses the quad’s normal vector to calculate the shade value. What would an appropriate name be?

common/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/material/MaterialViewImpl.java
IMS212: I’m pretty sure this can return DEFAULT.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/model/quad/BakedQuadMixin.java
IMS212: It uses a different injection point. This cannot be simply pseudoed in, as the Mixin target method must always exist.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
IMS212: Removed for now.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformBlockAccess.java
IMS212: Renamed to `getNormalVectorShade`.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/ResourcePackLoaderMixin.java
IMS212: Added. (It's above the preview)

common/src/main/java/net/caffeinemc/mods/sodium/client/services/SodiumModelData.java
IMS212: Added

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelAccess.java
IMS212: Replaced with `SodiumAuxilaryLightManager`, similar to `SodiumModelData`.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/block/NeoForgeBlockAccess.java
IMS212: Replaced with `AmbientOcclusionMode`.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelAccess.java
IMS212: Removed in favor of a `new PoseStack()`.

common/src/main/java/net/caffeinemc/mods/sodium/client/services/PlatformLevelAccess.java
IMS212: Replaced with FluidRendering.render, a Fabric API designed for this.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/features/model/WeightedBakedModelMixin.java
IMS212: They currently do not plan to fix this, as Vanilla itself is inconsistent.

common/src/api/java/net/caffeinemc/mods/sodium/api/util/ColorMixer.java
```diff
@@ -59,4 +59,19 @@ public static int mul(int a, int b) {
         // Pack the components
         return (c0 <<  0) | (c1 <<  8) | (c2 << 16) | (c3 << 24);
     }
+
+    /**
+     * Multiplies the 32-bit colors with one component
+     */
+    public static int mulSingle(int a, byte b) {
```
jellysquid3: This seems like it would be broken. Java uses signed bytes. So the valid range is between `-127` and `128`, and using it in multiplication results in it being sign-extended...

common/src/api/java/net/caffeinemc/mods/sodium/api/util/NormI8.java
```diff
@@ -114,4 +115,8 @@ public static boolean isOpposite(int normA, int normB) {
 
         return normAX == -normBX && normAY == -normBY && normAZ == -normBZ;
     }
+
+    public static Vector3f unpack(int i, Vector3f output) {
```
jellysquid3: i -> packed

common/src/api/java/net/caffeinemc/mods/sodium/api/math/MatrixHelper.java
```diff
@@ -129,6 +138,15 @@ public static void rotateZYX(PoseStack.Pose matrices, float angleZ, float angleY
                 .rotateZYX(angleZ, angleY, angleX);
     }
 
+    /**
```
jellysquid3: This seems like a dangling Javadoc comment? Where is the associated function?

common/src/api/java/net/caffeinemc/mods/sodium/api/math/MatrixHelper.java
```diff
@@ -17,17 +17,26 @@
...
     /**
      * @param mat The transformation matrix to apply to the normal
+     * @param skipNormalization Whether normalizing the vector is unnecessary
      * @param x The X-coordinate of the normal vector
      * @param y The Y-coordinate of the normal vector
      * @param z The Z-coordinate of the normal vector
      * @return The transformed normal vector (in packed format)
      */
-    public static int transformNormal(Matrix3f mat, float x, float y, float z) {
+    public static int transformNormal(Matrix3f mat, boolean skipNormalization, float x, float y, float z) {
```
jellysquid3, line 26: Probably, this function should be split into two (one with and without the `skipNormalization` parameter) so the branch only gets evaluated when it's actually necessary.

common/src/main/java/net/caffeinemc/mods/sodium/client/data/config/MixinConfig.java
```diff
@@ -129,47 +128,23 @@ private void readProperties(Properties props) {
...
-            ModMetadata meta = container.getMetadata();
-
-            if (meta.containsCustomValue(JSON_KEY_SODIUM_OPTIONS)) {
-                CustomValue overrides = meta.getCustomValue(JSON_KEY_SODIUM_OPTIONS);
-
-                if (overrides.getType() != CustomValue.CvType.OBJECT) {
-                    LOGGER.warn("Mod '{}' contains invalid Sodium option overrides, ignoring", meta.getId());
-                    continue;
-                }
+    public abstract void applyModOverrides();
```
jellysquid3: I don't think we want to abstract the MixinPlugin itself, but rather there should be a service for obtaining the overrides from mod manifests.

common/src/main/java/net/caffeinemc/mods/sodium/client/model/color/DefaultColorProviders.java
```diff
@@ -30,7 +26,7 @@ private GrassColorProvider() {
 
         @Override
         protected int getColor(LevelSlice slice, int x, int y, int z) {
-            return slice.getColor(BiomeColorSource.GRASS, x, y, z);
+            return BiomeColors.getAverageGrassColor(slice, blockPosHolder.get().set(x, y, z));
```
jellysquid3: Why did this change?

common/src/main/java/net/caffeinemc/mods/sodium/client/model/color/DefaultColorProviders.java
```diff
@@ -17,9 +15,7 @@ public static ColorProvider<BlockState> adapt(BlockColor color) {
         return new VanillaAdapter(color);
     }
 
-    public static ColorProvider<FluidState> adapt(FluidRenderHandler handler) {
-        return new FabricFluidAdapter(handler);
-    }
+    private static final ThreadLocal<BlockPos.MutableBlockPos> blockPosHolder = ThreadLocal.withInitial(BlockPos.MutableBlockPos::new);
```
jellysquid3: The thread local lookup here is going to be really slow when fetching colors during chunk meshing. Why did it need to be added?

common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/LightPipeline.java
```diff
@@ -18,6 +18,7 @@ public interface LightPipeline {
      * @param cullFace The cull face of the quad
      * @param lightFace The light face of the quad
      * @param shade True if the block is shaded by ambient occlusion
+     * @param isFluid True if the block is being drawn by the fluid renderer
      */
-    void calculate(ModelQuadView quad, BlockPos pos, QuadLightData out, Direction cullFace, Direction lightFace, boolean shade);
+    void calculate(ModelQuadView quad, BlockPos pos, QuadLightData out, Direction cullFace, Direction lightFace, boolean shade, boolean isFluid);
```
jellysquid3: Might make more sense to pass the `BlockState` or `FluidState` here instead of a boolean called `isFluid`.

jellysquid3 (changes requested):

common/src/api/java/net/caffeinemc/mods/sodium/api/util/ColorMixer.java
IMS212: Fixed by replacing it with int. This isn't actively used anywhere (yet).

common/src/main/java/net/caffeinemc/mods/sodium/client/model/color/DefaultColorProviders.java
IMS212: This changed to fix mod compatibility with a few mods, including Hydrological. It does not seem to incur any heavy cost, minus the usage of the BlockPos in the below comment.

common/src/main/java/net/caffeinemc/mods/sodium/client/model/color/DefaultColorProviders.java
IMS212: See the comment above; should it be replaced with a new BlockPos() for each quad?

IMS212: Actually, I have a new idea to replace this, but it may be a bit messy; providing a scratch BlockPos from the BlockRenderer.

jellysquid3 (approved):

PR #2642, opened 2024-08-06
Fullscreen Resolution Slider

### LINKED ISSUE #2692: Add option to change game resolution
### Request Description

In vanilla Minecraft, you can change the resolution/frequency. This is important because it defaults to 1080p60, I want to change it to 1080p75 (the frequency of my monitor). However when sodium is installed, this optioin seems to disapear.

douira:
As a compromise, this adds a fullscreen resolution slider but truncates the label to fit into one line. The design can be refined later, but at least users won't have to go to the vanilla settings for this now.
Thanks to @MeeniMc for their previous work in https://github.com/CaffeineMC/sodium-fabric/pull/2296 on adding the option in the first place. I added the two-line functionality and changed the style.
related: https://github.com/CaffeineMC/sodium-fabric/issues/46
Closes https://github.com/CaffeineMC/sodium-fabric/issues/2692


MeeniMc:
It's a good idea but it doesn't work with AUTO gui scale, there is no vertical scrolling and the extra line renders outside the screen now.

douira:
That's a good point, do you know where one makes the control element notify the GUI system of it's correct size? Is there a place where it calculates the total size of the GUI for determining the scale?

douira:
I've updated it to be only one line but truncate the width of the label dynamically based on the slider's value width.

MeeniMc:
LGTM

douira:
resolved conflicts and brought the branch up to speed with dev

douira:
Closes https://github.com/CaffeineMC/sodium-fabric/issues/2692

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/ControlElement.java
```diff
@@ -20,18 +20,50 @@ public ControlElement(Option<T> option, Dim2i dim) {
...
 
-        if ((this.hovered || this.isFocused()) && this.font.width(name) > (this.dim.width() - this.option.getControl().getMaxWidth())) {
-            name = name.substring(0, Math.min(name.length(), 10)) + "...";
+        // add the star suffix before truncation to prevent it from overlapping with the label text
+        if (this.option.isAvailable() && this.option.hasChanged()) {
+            name = name + " *";
         }
 
+        // on focus or hover truncate the label to never overlap with the control's content
+        if (this.hovered || this.isFocused()) {
```
jellysquid3, line 38: This should really be a separate utility function.

jellysquid3 (approved):

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/ControlElement.java
douira: good point

douira: done

jellysquid3:
This is fine as a temporary solution for now. But I wouldn't want to keep around multi-line controls once we rework the video setting screen.

douira:
This PR doesn't add it as a multiline option (anymore), I replaced it with label truncation.

douira:
My bad for not updating the description

PR #2645, opened 2024-08-07
Shadow com.lodborg.intervaltree into sodium

douira:
Shadows the interval tree library into sodium with insubstantial (just cleanup) modifications.

This avoids a dependency and makes modloading easier, espcially since this dependency is unlikely to get updated again or require changes.

I've copied the MIT license declaration of the original work into each file as a header with a note and a link to the library's own repo.


jellysquid3:
These changes look fine, but we might want to modernize the code for Java 8+ and remove the dead code we aren't using later.

jellysquid3 (approved):

douira:
I already had IJ highlight and remove dead code. As for Java 8, it might require a more substantial refactor to change it to properly use generics throughout.

jellysquid3:
This needs to be rebased on /dev, since we just merged the multi-loader branch.

PR #2648, opened 2024-08-09
Fix fast cloud rendering

### LINKED ISSUE #2646: Game crashes when the camera is above the cloud height and fast rendering is used
The code responsible for culling back faces in cloud geometry is not exactly right. Since the fast-mode rendering does not use back-face culling, it incorrectly culls the bottom most faces and causes a crash due to the geometry buffer being empty.

We should fix this by skipping the face-cull logic when using fast-mode rendering. If the geometry buffer is empty after the meshing step, don't attempt to finalize it into a GPU vertex buffer.

douira:
Fix fast cloud rendering as described in #2646. Closes #2646.

Makes it possible to have no vertex buffer, and to simply not render clouds if there is none. Disables view-dependent culling on fast clouds. Resolves the crash and makes them visible from the top and bottom.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
@@ -245,6 +245,10 @@ public class CloudRenderer {
 
         int cellColor = textureData.getCellColor(cellIndex);
 
+        if (ColorABGR.unpackAlpha(cellColor) == 0) {
+            return;
+        }
+
         float x = offsetX * 12;
         float z = offsetZ * 12;
 
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
@@ -84,6 +84,11 @@ public class CloudRenderer {
             this.cachedGeometry = (geometry = rebuildGeometry(geometry, parameters, this.textureData));
         }
 
+        VertexBuffer vertexBuffer = geometry.vertexBuffer();
+        if (vertexBuffer == null) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
```diff
@@ -337,21 +344,16 @@ private static void emitCellGeometry2D(VertexBufferWriter writer, int faces, int
             long ptr = buffer;
             int count = 0;
 
-            // -Y
-            if (CloudFaceSet.contains(faces, CloudFace.NEG_Y)) {
```
jellysquid3, line 341: Won't this result in a face always being emitted in the 2D mode, even if the texel is fully transparent? Since we already have the texel's color in this function, we could just check if it's fully transparent without looking at the face set.

jellysquid3 (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/CloudRenderer.java
douira: Since this wasn't checking the color at all, I think it's always emitting the face if you're below the cloud level, and always after the changes in this patch. I've added a general check against zero alpha. I'm not sure if the 3d path was already doing this somewhere else (I can't find anything like it).

jellysquid3 (commented):
Since fast clouds are rendered with face culling disabled, we don't need to factor in the orientation when the player goes above the cloud layer. This avoids unnecessary rebuilds of the geometry (though I admit this is unlikely to matter.)

The easiest solution would likely be to replace the `signnum` with a function that returns `-1` or `+1` when fancy graphics are used (camera is below and above the center height of the clouds), and `0` when fast graphics are used.

jellysquid3 (approved):

IMS212 (approved):

PR #2421, opened 2024-04-14
Combine draw commands to improve rendering performance

douira:
This PR makes it so that draw commands that read from adjacent vertex data are combined. This reduces the number of draw commands by around 30% and improves fps on my system by up to 57% depending on the scene and circumstances. I'm on macOS with a 6900 XT. This performance improvement likely comes, as jellysquid stated on discord, from reduced CPU overhead in the driver and better GPU occupancy.

Please test if this results in a similar improvement or other effect, as it's probably dependent on graphics card, memory bandwidth, and platform (os/driver/vendor etc).

Here's a recording of the number of draw commands per pass:
ts on, before:
Draw total for pass Solid: 15531
Draw total for pass Cutout: 13277
Draw total for pass Translucent: 2298

ts on, after:
Draw total for pass Solid: 9571
Draw total for pass Cutout: 8306
Draw total for pass Translucent: 2298

ts off, before:
Draw total for pass Solid: 15531
Draw total for pass Cutout: 13277
Draw total for pass Translucent: 3812

ts off, after:
Draw total for pass Solid: 9571
Draw total for pass Cutout: 8306
Draw total for pass Translucent: 3645

Here's some screenshots without and with this patch. The fps numbers here are outdated, since this branch has been updated in the meantime. See the [newest comments at the bottom of this thread](https://github.com/CaffeineMC/sodium-fabric/pull/2421#issuecomment-2362683899) instead.

Screenshot 2024-04-14 at 04 29 12
Screenshot 2024-04-14 at 04 27 59
Screenshot 2024-04-14 at 04 30 51
Screenshot 2024-04-14 at 04 32 25
Screenshot 2024-04-14 at 04 48 21
Screenshot 2024-04-14 at 04 49 16

PR #2655, opened 2024-08-11
[TS] Delete VertexRange, use isTranslucent, fix partition tree sorting on geometry with negative dot products

douira:
- Delete `VertexRange` as since the introduction of the translucency sorting system its first `offset` field wasn't getting used anymore. It can simply be calculated on the fly as a prefix sum. Currently missing vertex ranges are replicated as `-1` lengths which are distinct from a length of zero. I don't know if this is actually necessary or if we can just use a zero length as the default.
- Use the `isTranslucent` method on terrain render passes and materials instead of checking equality
- Fix a crash related to translucency sorting that happened during the initial sort in the meshing task. This bug was triggered when a certain model (I don't know which one) on MCC Island was being sorted. Importantly, it extended outside of the chunk so that the dot products of its quads were negative with the axes. The partition tree building algorithm incorrectly handled negative dot products and produced a partition tree with duplicate nodes, which in turn led to a buffer overflow since too many quad indexes were being written.
- Fixed `floatToComparableInt` to actually be correct, it doesn't seem to have been correct previously.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/ChunkBuildBuffers.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/ChunkBuildBuffers.java
@@ -69,18 +69,16 @@ public class ChunkBuildBuffers {
             var buffer = builder.getVertexBuffer(facing);
 
             if (buffer.isEmpty()) {
-                vertexCounts[ordinal] = -1;
                 continue;
             }
 
             vertexBuffers.add(buffer.slice());
+            var bufferCount = buffer.count();
             if (!forceUnassigned) {
-                vertexCounts[ordinal] = buffer.count();
-            } else {
-                vertexCounts[ordinal] = -1;
+                vertexCounts[ordinal] = bufferCount;
             }
 
-            vertexSum += buffer.count();
+            vertexSum += bufferCount;
         }
 
         if (vertexSum == 0) {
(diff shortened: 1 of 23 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/ChunkBuildBuffers.java
```diff
@@ -61,41 +60,45 @@ public BuiltSectionMeshParts createMesh(TerrainRenderPass pass, boolean forceUna
...
 
-        int vertexCount = 0;
+        int vertexSum = 0;
 
         for (ModelQuadFacing facing : ModelQuadFacing.VALUES) {
+            var ordinal = facing.ordinal();
             var buffer = builder.getVertexBuffer(facing);
 
             if (buffer.isEmpty()) {
+                vertexCounts[ordinal] = -1;
```
jellysquid3: Is it really necessary to special case this to be `-1` instead of `0` here? The behavior is not very intuitive and results in needing to case both `0` (no data) and `-1` (null) in rendering code.

douira: I think it is unnecessary to use `-1` here, I probably just thought it was necessary since `null` vertex range objects are relevant for my other PR https://github.com/CaffeineMC/sodium-fabric/pull/2421 (which stands in conflict with this one, at least without refactoring it to use some other system of storing the reordering of facings, for example by encoding the positions into a long)

douira: I've pushed changes to remove the `-1` and also to improve the handling of the mesh object (instead it now just uses the vertex counts).

IMS212:
Would it make sense to remove the TS option from settings now, rather than later?

douira:
I don't know if a setting is necessary. If we don't have to have it, we should remove it before release. Maybe keeping it as a setting in the config file only is enough. However, for debugging TS issues it's helpful if it can be disabled as a comparison. Then again, we can also send them a build with the setting and/or it disabled by default.

jellysquid3:
I think we'll get rid of it (or put it behind a developer option) before release. But it's not relevant for this pull request.

jellysquid3 (approved):

IMS212 (approved):

PR #2660, opened 2024-08-12
Hide the TS option to only be visible in dev environments

douira:
- Hide the TS option to only be visible in dev environments
- Rename the option in config files to reset it after having hidden it. (to prevent people who've turned it off from being unable to turn it on again)

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumGameOptionPages.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumGameOptionPages.java
@@ -14,6 +14,7 @@ import net.caffeinemc.mods.sodium.client.gui.options.control.TickBoxControl;
 import net.caffeinemc.mods.sodium.client.gui.options.storage.MinecraftOptionsStorage;
 import net.caffeinemc.mods.sodium.client.gui.options.storage.SodiumOptionsStorage;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.Workarounds;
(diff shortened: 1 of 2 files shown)


jellysquid3 (approved):

NebelNidas:
Why shouldn't users with weak machines be able to turn this off if they don't necessarily care about translucent objects looking correct? Is the impact of translucency sorting really that low?

jellysquid3:
On the GPU, there is no additional overhead (the same stuff just gets rendered in a different order.)

On the CPU, there is _some_ overhead... But the way Douira's approach works allows for translucent objects to only be re-sorted when strictly necessary, meaning that very little (or possibly none) of the geometry needs to be sorted when you move around. And, even then, you may have to move around _a lot_ before you reach that threshold.

Point being, we're talking about less than a percent of the frame time in any reasonable situation, and only periodically at that. We've not had reports from players suggesting it causes a performance regression, or that it is worth disabling in any situation.

jellysquid3:
[Screencast_20240214_001830_trimmed.webm](https://github.com/user-attachments/assets/918d5641-70f8-432d-886e-fef36e844a4b)

Even with a completely unreasonable and contrived example (an entire world made of honey blocks, over 10 GB of geometry) there is minimal performance impact from sorting the geometry, and each frame still manages to look perfectly correct.

NebelNidas:
Oh wow, that's truly impressive indeed. I was asking since I have some family members with _very_ slow laptops, where every option to turn off unnecessary eye-candy helps. But if we're talking about less than 1% frame times, it should be fine I guess. Thanks for the detailed anwser!

PR #2662, opened 2024-08-13
Update breaks in fabric.mod.json

pajicadvance:


jellysquid3:
I'm wondering if we need to `breaks: indium`, or if `provides: indium` is enough.

IMS212:
Indium has a hard lock to my knowledge; a breaks doesn't make sense.

IMS212 (approved):

jellysquid3 (approved):
Looks fine to me, we'll probably refine this list as time goes on.

PR #2666, opened 2024-08-15
Downgrade quad materials by scanning their textures

douira:
Adds texture scanning to the sprite contents mixing to check if any of the pixels in a sprite are fully transparent or even translucent. This information is then used during block rendering to "downgrade" a quad's material to either cutout or solid, if no translucent or transparent pixels are present in the quad's sprite, respectively.

The benefit of material downgrading is that it for one, has a performance benefit since cutout and translucent materials are somewhat more expensive to render. It also helps reduce the load on the translucency sorting system if it can avoid non-translucent quads from being declared translucent and subsequently incorporated in the translucency sorting calculations.

It does not separately scan each quad's area in sprites that are used by multiple quads since this is complex and/or unlikely to perform well. Concrete suggestions of mod-compatible and performant approaches to per-quad texture scanning are welcome. Mods should instead use FRAPI or Forge APIs to assign multiple materials to the quads in their models such that each quad only renders as the most demanding material it requires.


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -175,15 +175,31 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
...
+        ChunkModelBuilder builder = this.buffers.get(material);
+
+        ChunkMeshBufferBuilder vertexBuffer = builder.getVertexBuffer(normalFace);
         vertexBuffer.push(vertices, material);
 
-        modelBuilder.addSprite(SpriteFinderCache.forBlockAtlas().find(quad.getTexU(0), quad.getTexV(0)));
+        builder.addSprite(atlasSprite);
+    }
+
+    private Material getDowngradedMaterial(TextureAtlasSprite sprite, Material material) {
```
jellysquid3: Should be static.

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/mipmaps/SpriteContentsMixin.java
```diff
@@ -51,7 +58,7 @@ public class SpriteContentsMixin {
      * black color does not leak over into sampling.
      */
     @Unique
-    private static void sodium$fillInTransparentPixelColors(NativeImage nativeImage) {
+    private void sodium$fillInTransparentPixelColors(NativeImage nativeImage) {
```
jellysquid3: This new functionality should be placed in a separate mixin class. Most likely the best place to initialize these variables would be in the constructor. Otherwise, it's not possible to disable the changes to mip-map generation without also disabling the material downgrading functionality.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -175,15 +175,31 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
...
+        var contents = (SpriteContentsExtension) (sprite.contents());
+        if (material.pass == DefaultTerrainRenderPasses.TRANSLUCENT && !contents.sodium$hasTranslucentPixels()) {
```
jellysquid3: This cast can fail if the mixin has been disabled. This should check if the cast is successful with:
```java
if (sprite.contents() instanceof SpriteContentsExtension contents) {
  // ...
}
```
If the mixin isn't applied, the interface won't exist on the class, so it will fail gracefully.

jellysquid3 (changes requested):

jellysquid3:
I didn't investigate this fully in the code review, but: When downgrading materials, we must be careful to use a material which has the same alpha-cutoff and mip-mapping behavior. The only thing we should be changing is the render pass the geometry is placed into.

jellysquid3:
I think the abstract you describe here is fine, though. This is very similar to a patch set I wrote, but due to recent events (as we've discussed) it's difficult for me to get access to that code.

Things like Grass blocks contribute to a significant number of fragments in the overall scene, and this patch would make it so the base dirt layer to no longer rendered with alpha-testing enabled, which is likely a decent performance win.

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/mipmaps/SpriteContentsMixin.java
douira: Should it have a separate loop scanning over all the pixels? I did it this way because I thought it would be wasteful scanning the pixels separately.

douira:
I've resolved the comments in the latest commit. This required changing some things to accept the material as bits instead of the material object since I need to change the alpha cutoff parameter independently since otherwise a downgraded translucent material has a zero cutoff which doesn't work. I hope it's kinda ok the way it is now. Thanks to @LlamaLad7 for helping me rewrite the mixin.

douira:
It might also be a good idea to do a performance comparison in both rendering and meshing times to see what impact downgrading has on those.

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/mipmaps/SpriteContentsMixin.java
jellysquid3: Yes, a separate loop would be preferable. I don't think the performance impact is that significant. But we can always profile later and verify that.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -172,14 +175,33 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
...
 
             out.light = quad.lightmap(srcIndex);
         }
 
-        var atlasSprite = SpriteFinderCache.forBlockAtlas().find(quad.getTexU(0), quad.getTexV(0));
-        var pass = getDowngradedPass(atlasSprite, material.pass);
+        var atlasSprite = SpriteFinderCache.forBlockAtlas().find(uSum / 4.0f, vSum / 4.0f);
+
+        // sanity check that the quad's UVs are within the sprite's bounds
+        var spriteUMin = atlasSprite.getU0();
```
jellysquid3, line 215: This check to see if the UVs are within sprite bounds should be pulled into a separate function. Then, the variable `doDowngrade` can be replaced with a function call.

jellysquid3 (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
douira: Good point

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -169,21 +175,67 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
...
 
-            out.u = quad.u(srcIndex);
-            out.v = quad.v(srcIndex);
+            uSum += out.u = quad.u(srcIndex);
+            vSum += out.v = quad.v(srcIndex);
 
             out.light = quad.lightmap(srcIndex);
         }
 
+        var atlasSprite = SpriteFinderCache.forBlockAtlas().find(uSum / 4.0f, vSum / 4.0f);
```
jellysquid3, line 184: There is a potential landmine with mods that specify an alpha component on vertex colors, which is now possible due to [this change](https://github.com/CaffeineMC/sodium-fabric/commit/d61617800141a4d6684bc06a3ec243edba8b4f4b). This is used by mods like [Fluid Void Fading](https://modrinth.com/mod/fluidvoidfading) to have normally opaque blocks (such as Lava) render as translucent. We need to make sure that if any of the vertex colors are non-opaque that material downgrading does not happen.

jellysquid3 (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
douira: ok I've implemented a check for opaque vertexes if the pass is alpha blended

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/scan/SpriteContentsMixin.java
```diff
@@ -0,0 +1,66 @@
...
+    @Final
+    private NativeImage originalImage;
+
+    @Unique
+    public boolean sodium$hasTransparentPixels = false;
+
+    @Unique
+    public boolean sodium$hasTranslucentPixels = false;
+
+    @WrapOperation(method = "<init>", at = @At(value = "FIELD", target = "Lnet/minecraft/client/renderer/texture/SpriteContents;originalImage:Lcom/mojang/blaze3d/platform/NativeImage;", opcode = Opcodes.PUTFIELD))
```
jellysquid3, line 36: I'm not sure using `@WrapOperation` is helping anything here. Would it not be simpler to inject to the tail of the constructor?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/scan/SpriteContentsMixin.java
```diff
@@ -0,0 +1,66 @@
...
+    private void sodium$beforeGenerateMipLevels(SpriteContents instance, NativeImage nativeImage, Operation<Void> original) {
+        sodium$scanSpriteContents(nativeImage);
+
+        original.call(instance, nativeImage);
+    }
+
+    @Unique
+    private void sodium$scanSpriteContents(NativeImage nativeImage) {
```
jellysquid3: Method name prefixes are not needed when using `@Unique`.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -164,23 +174,81 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
             out.z = quad.z(srcIndex) + offset.z;
 
             // FRAPI uses ARGB color format; convert to ABGR.
-            out.color = ColorARGB.toABGR(quad.color(srcIndex));
+            var color = quad.color(srcIndex);
+            alphaSum += ColorARGB.unpackAlpha(color);
```
jellysquid3: Why is a sum here used? If the goal is to keep the code branchless, it would be easier to write `transparent |= ColorARGB.unpackAlpha(color) != 0xFF` (which should also be a little faster, since it compiles to a low byte comparison.)

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer.java
```diff
@@ -164,23 +174,81 @@ private void bufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material
...
+            uSum += out.u = quad.u(srcIndex);
+            vSum += out.v = quad.v(srcIndex);
 
             out.light = quad.lightmap(srcIndex);
         }
 
+        var atlasSprite = SpriteFinderCache.forBlockAtlas().find(uSum / 4.0f, vSum / 4.0f);
+
+        // don't do downgrade if some vertex is not fully opaque
+        if (pass.isTranslucent() && alphaSum < 4 * 255) {
```
jellysquid3: I'd really like to see this logic for material downgrading moved to a separate function, since it makes reading `BlockRenderer.bufferQuad` a lot harder. The check to see if any vertex is transparent could also be moved into it, as I don't think a second loop is really that problematic.

jellysquid3 (changes requested):

jellysquid3:
These changes look good to me now, outside of my comments regarding code style. Thanks for your effort.

douira:
I've refactored it into a separate method as far as possible, and added some documentation to both the mixin and the quad processing.

jellysquid3 (approved):

IMS212 (approved):

jellysquid3:
This seems fine in my testing, so we can merge it for 0.6.0-beta.2 if you can rebase the pull request on /dev.

douira:
Merged dev into the branch, you can squash merge the PR.

PR #2674, opened 2024-08-16
Clean up handling of vertex formats

jellysquid3:
Minecraft 1.21 makes it unnecessary for us to keep track of our own VertexFormatDescription, because they now have the same information we were trying to unpack/cache.

Additionally, they have also implemented something very similar to our element masking approach for BufferBuilder, so it's no longer possible for mods to create vertex formats with the same element type specified multiple times. Since they don't care, and mods can't do it anymore, we can avoid checking whether a vertex format is not "simple".

To finally get rid of the `VertexFormatDescription`, the `VertexFormat` class was extended to store a global identifier, which can then be used to create `VertexSerializer` identifiers. This isn't really any different from before, but the identifier is stored in the `VertexFormat` directly now and avoids a hash table lookup.


IMS212 (approved):

IMS212:
Should `CommonVertexAttribute` be deleted?

PR #2676, opened 2024-08-16
Use hardware copy for render target blits where possible

jellysquid3:
Minecraft performs a blit using a fragment shader, which is unnecessary when blending is not used. Using the fixed function hardware to perform the blit is much faster and doesn't utilize the rasterization engine.


IMS212 (approved):
Seems simple... I don't believe the `blit_screen` shader is changeable by RP's, so we don't need a check.

jellysquid3:
I'm going to merge this since the patch is simple enough and breakages are unlikely.

douira:
It didn't break when I tried it

PR #2677, opened 2024-08-16
Single-pass Compositing

jellysquid3:
This is an early draft and should not be merged, as it currently breaks a few things and the code is very bad.

The idea of this pull request is to eliminate the overhead of rendering the vignette and the alpha blending of the entity outline buffer, which currently contributes to a significant amount of GPU time (sometimes 10%+ of the total frame time.)

We achieve this by creating a special compositing shader which blends these render targets in a single operation.

Additionally, some optimizations are made to how the main render target is configured. We don't need the alpha channel on the color attachment, and the depth texture can be converted to a depth buffer so that drivers are allowed to use specialized depth formats.

These patches, depending on the scene, can improve frame times by 5% to 15% on Intel Xe graphics under Linux. So it would seem a worthwhile investment to try and refine them further, since right now this is a fixed overhead that is *always* applied.


jellysquid3:
We might be able to pull out some parts of this into smaller pull requests, particularly the alpha-blending of the entity outline render target into the main render target.

jellysquid3:
Also, making this into something actually usable is not much of a priority for me right now. It will probably be revisited with Sodium 0.7 at the earliest, unless someone else decides to look into it.

PR #2679, opened 2024-08-16
Track render targets to avoid unnecessary composites

jellysquid3:
The entity outline buffer is always composited into the final image, even if it wasn't rendered to. This accounts for a significant amount of the frame render time on the GPU (up to 5%) and can be avoided when there are no entity outlines.

Unfortunately, it's difficult to find an injection site where we can track whether the entity outline render target is modified. And there are conditions (other than using the VertexBuffer class to draw) where a mod could potentially modify the entity outline buffer. So this is not guaranteed to be robust in all situations, even if those situations are very unlikely.


jellysquid3:
Minecraft 1.21.2 makes some changes around how render targets are bound, which makes it easier to track if the entity outline buffer is getting touched. So it would probably be easier to just wait until that lands, rather than hack it in ourselves.

jellysquid3:
Actually, it seems like Minecraft 1.21.2 _might_ already implement something like this, so it probably won't be necessary to even bother. I'll just close this pull request for now.

PR #2687, opened 2024-08-17
Fix compiler warnings

mrmangohands:
The common project includes Sodium's Fabric API integration, which can be used on either loader. The Fabric API modules reference Fabric Loader's API classes, so we need to include it to prevent warning spam about an unknown enum constant.

This PR also splits out the remaining platform specific code from common, and adds `remap = false` where necessary to fix warnings.

### DIFF
--- a/neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
+++ b/neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
@@ -4,18 +4,18 @@ import net.caffeinemc.mods.sodium.client.render.SodiumWorldRenderer;
 import net.minecraft.client.renderer.LevelRenderer;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.injection.At;
-import org.spongepowered.asm.mixin.injection.Inject;
-import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.Overwrite;
 
 import java.util.function.Consumer;
 
 @Mixin(LevelRenderer.class)
 public class LevelRendererMixin {
-    @Inject(method = "iterateVisibleBlockEntities", at = @At("HEAD"), cancellable = true)
-    public void replaceBlockEntityIteration(Consumer<BlockEntity> blockEntityConsumer, CallbackInfo ci) {
-        ci.cancel();
-
+    /**
+     * @author MrMangoHands
+     * @reason Redirect iteration of visible block entities to our renderer
+     */
+    @Overwrite
+    public void iterateVisibleBlockEntities(Consumer<BlockEntity> blockEntityConsumer) {
         SodiumWorldRenderer.instance().iterateVisibleBlockEntities(blockEntityConsumer);
     }
 }
--- a/neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
+++ b/neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
@@ -1,0 +1,21 @@
+package net.caffeinemc.mods.sodium.neoforge.mixin.core.render.world;
+
+import net.caffeinemc.mods.sodium.client.render.SodiumWorldRenderer;
+import net.minecraft.client.renderer.LevelRenderer;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.function.Consumer;
+
+@Mixin(LevelRenderer.class)
+public class LevelRendererMixin {
+    @Inject(method = "iterateVisibleBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void replaceBlockEntityIteration(Consumer<BlockEntity> blockEntityConsumer, CallbackInfo ci) {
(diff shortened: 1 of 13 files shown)


neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/features/options/render_layers/ItemBlockRenderTypesMixin.java
```diff
@@ -0,0 +1,30 @@
...
+public class ItemBlockRenderTypesMixin {
+    @Unique
+    private static boolean leavesFancy;
+
+    // getRenderLayers is a NeoForge only function required for the leaves to properly work.
+    @Redirect(
+            method = { "getRenderLayers" },
+            at = @At(value = "FIELD", target = "Lnet/minecraft/client/renderer/ItemBlockRenderTypes;renderCutout:Z"))
+    private static boolean redirectLeavesShouldBeFancy() {
+        return leavesFancy;
```
jellysquid3, line 23: Needs `this` to follow code style.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/workarounds/context_creation/WindowMixin.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.function.LongSupplier;
+import java.util.function.Supplier;
+
+@Mixin(Window.class)
+public class WindowMixin {
+    @WrapOperation(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/neoforged/fml/loading/ImmediateWindowHandler;setupMinecraftWindow(Ljava/util/function/IntSupplier;Ljava/util/function/IntSupplier;Ljava/util/function/Supplier;Ljava/util/function/LongSupplier;)J"), remap = false)
+    private long wrapGlfwCreateWindowForge(final IntSupplier width, final IntSupplier height, final Supplier<String> title, final LongSupplier monitor, Operation<Long> op) {
+        final boolean applyNvidiaWorkarounds = Workarounds.isWorkaroundEnabled(Workarounds.Reference.NVIDIA_THREADED_OPTIMIZATIONS);
+
+        if (applyNvidiaWorkarounds && !PlatformRuntimeInformation.getInstance().platformHasEarlyLoadingScreen()) {
```
jellysquid3, line 22: This is in NeoForge-specific code now, so do we even need the abstraction over whether the platform has an early-loading screen?

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
```diff
@@ -0,0 +1,21 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.function.Consumer;
+
+@Mixin(LevelRenderer.class)
+public class LevelRendererMixin {
+    @Inject(method = "iterateVisibleBlockEntities", at = @At("HEAD"), cancellable = true)
+    public void replaceBlockEntityIteration(Consumer<BlockEntity> blockEntityConsumer, CallbackInfo ci) {
```
jellysquid3: Any reason we're preferring the inject-and-cancel instead of an overwrite or otherwise? It doesn't make sense why anyone else would touch this, and it's probably going to cause strange failures if they do.

jellysquid3 (changes requested):

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/features/options/render_layers/ItemBlockRenderTypesMixin.java
mrmangohands: It's already static so this can't be referenced

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/workarounds/context_creation/WindowMixin.java
mrmangohands: No, I'll go ahead and remove that

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
mrmangohands: @IMS212?

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/workarounds/context_creation/WindowMixin.java
mrmangohands: Actually it's querying a value from the fmlconfig, looking into it

mrmangohands: Ok we do still need this because if the user has disabled earlyWindowControl in forge.toml we can apply our workarounds. Dropping the commit removing it.

neoforge/src/main/java/net/caffeinemc/mods/sodium/neoforge/mixin/core/render/world/LevelRendererMixin.java
IMS212: I believe an overwrite would have failed when the Mixin was merged, for weird reasons; it can be done now.

mrmangohands: Updated 👍

douira:
Have the review concerns been resolved? Also this needs to have its conflicts resolved

jellysquid3:
Resolved with https://github.com/CaffeineMC/sodium/commit/dc7dc4c634316ac8b510d2dc8290079ebc462778. Thanks!

PR #2690, opened 2024-08-17
Multiple misc changes to shader interfaces

IMS212:


common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/GlProgram.java
```diff
@@ -51,22 +51,30 @@ public void delete() {
     }
 
     @Override
-    public <U extends GlUniform<?>> U bindUniform(String name, IntFunction<U> factory) {
+    public <U extends GlUniform<?>> U bindUniform(String name, IntFunction<U> factory, boolean required) {
```
jellysquid3: I'd rather see a `bindUniformOptional` which returns `@Nullable`, and a `bindUniform` which returns `@NotNull` and does the check itself.

common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/GlProgram.java
```diff
@@ -51,22 +51,30 @@ public void delete() {
...
+                return null;
+            }
         }
 
         return factory.apply(index);
     }
 
     @Override
-    public GlUniformBlock bindUniformBlock(String name, int bindingPoint) {
+    public GlUniformBlock bindUniformBlock(String name, int bindingPoint, boolean required) {
```
jellysquid3: See comment about `bindUniform`.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/vertex/serializers/VertexSerializerRegistryImpl.java
```diff
@@ -46,6 +46,11 @@ public VertexSerializer get(VertexFormatDescription srcFormat, VertexFormatDescr
         return serializer;
     }
 
+    @Override
+    public void registerSerializer(VertexFormatDescription srcFormat, VertexFormatDescription dstFormat, VertexSerializer serializer) {
```
jellysquid3, line 50: I'm concerned about race conditions if Iris doesn't call into this early enough. But if it's handled in the client initializer, it's _probably_ fine?

IMS212:
Fixed the two things; not sure if any changes should be made to the serializer portion, or if that can wait on b2.

jellysquid3 (approved):

PR #2691, opened 2024-08-19
Shader Line Annotation

### LINKED ISSUE #2680: Add line numbers to merged shader sources and improve the debugging experience
Right now, we don't add `#line` directives to the merged shader sources, which makes reading line numbers a lot harder. To make it easier to debug compilation errors, the following changes should be made:

- Each included shader file (and the prelude with `#version` and specialization constants) should be given a unique index.
- In the merged source code, `#line <line number> <source file index>` declarations should be added, and the original `#include` macro should be commented out (for debugging purposes). 
- The shader compilation log should include the merged source code.

douira:
- add line directive insertion to ShaderParser,
- use assigned source ids to remap the sources in shader compilation logs,
- print merged shader sources to the log file in dev environments

Fixes #2680

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/GlShader.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/GlShader.java
@@ -27,8 +27,8 @@ public class GlShader extends GlObject {
         String log = GL20C.glGetShaderInfoLog(handle);
 
         if (!log.isEmpty()) {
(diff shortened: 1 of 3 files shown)


douira:
Please test if the included regex of the form `\d+:\d+:` works for the shader compilation logs produced by other drivers. If not, I'll adjust the regex to also match those. Also please test that the line numbers produced in the errors actually correspond to the line numbers in the source files, to make sure the drivers aren't interpreting the line directive differently.

douira:
Removed remapping code, should be good to go now.

PR #2697, opened 2024-08-21
Improved texture filling

douira:
Work items:

- [x] Implement better filling algorithm for transparent pixels
- [ ] Handle NPOT textures by expanding them to the next POT
- [ ] Rewrite atlas stitching to take advantage of POT-sized textures
- [ ] Deal with textures smaller than 2^mipLevel (even if they are POT) which otherwise limit the mip level to log2(size) by either upscaling or doing something else

Prerequisite: https://github.com/CaffeineMC/sodium-fabric/pull/2666


djmrFunnyMan:
If I understand correctly, this won't fix low-POT textures dropping mip levels (for example a single 4x4 texture will reduce all mip levels to 2) 
Could this also be adressed? (by resizing to 16x16 or reusing mip levels)

Issue #2304, opened 2024-02-01
Support additional languages


jellysquid3:
Sodium does not currently support any language other than English, as the translation files do not exist for them. However, the mod is widely downloaded in other countries, where the predominant language is something else.

For accessibility, we should begin accepting submissions for translations, and incorporate them into the release process so that every published version contains the latest translation strings. 

We should also eliminate current usages of vanilla translation strings in the video settings screen, as many of them are misappropriated (see graphical toggles for "Fast/Fancy".)

Not being considered:
- Translation of "flavor text" messages (i.e. the donation prompt), at least for now.
- Support for internet joke languages (lolcat and similar.)

jellysquid3:
Using information from Modrinth about what countries the mod is most downloaded from, it appears these languages would be the most important to implement first (in descending order):

- Simplified Chinese (官话)
- Russian (Россия)
- German (Deutsch)
- Portuguese (Português)
- Hindi (हिन्दी)
- Polish (Polski)
- French (Français)
- Korean (한국어)

jellysquid3:
The scope of this work is rather large, so we likely need to think about it in terms of stages:

- Stage 0 - We set up the infrastructure to enable people to submit translations, and prepare the necessary plumbing to have them included as part of the build process.
- Stage 1 - We invite a limited number of collaborators for key languages into the project to help out. Likely there will be some discussion about how we can better format things in the Video Settings screen to accommodate translated texts.
- Stage 2 - We publish a preview version of Sodium with translations for a limited set of languages, to gather user feedback. There may be further improvements at this point.
- Stage 3 - We enable support for other languages, and open the infrastructure to the public.

jellysquid3:
For the time being, this is going to be effectively blocked, since we need to make changes to our layout code that so translation strings don't run off the end of buttons and toggles.

Madis0:
As previously mentioned in #44, me and @FlashyReese are fully open to transferring the [unofficial Crowdin project](https://crowdin.com/project/sodium-fabric) and all of its CC0-licensed translations to you, which was the original intent anyway. That should reduce the work needed on this issue as well.

PR #2735, opened 2024-09-04
Create zh_cn.json

suoyukii:
By en_us file translated zh_cn file.


qznfbnj:
Looks like they will use the existing crowdin translation in the future?
Related: https://github.com/CaffeineMC/sodium-fabric/issues/2304#issuecomment-1921242753

suoyukii:
I think crowdin is bad for text comparison, and it can be troublesome to use crowdin if you add content en_us file or change the key name.

PR #2753, opened 2024-09-15
Combine the vertex position attributes for performance

jellysquid3:
This improves terrain rendering performance significantly for me, with Intel Xe-LP graphics under Linux.

1280x720: 204 fps -> 238 fps (+16%)
2256x1504: 155 fps -> 168 fps (+8%)


douira:
I'm seeing a very slight regression from 210->205 fps on macOS 14.1 AMD 6900 XT at 6720x3674. It persists across restarts of the game, but is maybe too small to be significant.

dev
dev

this patch
vertex-format-1

No difference on 1708x960.

dev
dev 2

this patch
vertex-format-1 2

PR #2766, opened 2024-09-21
Add Fog Occlusion for particles

AnOpenSauceDev:
I'm not sure if this is best suited for Sodium, but this PR fixes Minecraft's behavior of rendering particles regardless of fog occlusion. This PR makes Sodium skip particle rendering if a particle falls behind opaque fog and additionally has Fog Occlusion enabled (essentially just re-using the existing fog occlusion logic).

(Render Distance is set to 2.)

Sodium with Fog|No Fog|No Fog + Particle Fog Occlusion
----------|---|---
 _despite the fog occluding the particles, they are still having `render` called._| _disabling terrain fog shows the particles are still being rendered._| _with Fog Occlusion enabled, the particles no longer render past fog distance, or in some cases, outside of render distance_

Let me know if I need to fix/rework anything!


jellysquid3:
I don't see a problem with the proposed changes as they are written. But has this been measured to be a problem? I've suspected that some servers will spawn many particles outside the render distance, but have not actually observed this to significantly impact performance.

AnOpenSauceDev:
Particle rendering on thousands of particles can take up a few milliseconds every frame even on my PC, and I assume the time penalty from all the unnecessary rendering would have a worse effect on a lower-end computer. I do have some older hardware lying around, so I might benchmark the performance thoroughly there.

AnOpenSauceDev:
On a slower laptop I have (AMD 7530U), a somewhat empty Hypixel Skyblock lobby in the End at 2 Render Distance takes about 1.5ms (lowest was 0.8ms, highest was 2.1ms) to render a frame that stays around 3000 particles. 

Turning on fog occlusion drops that to the range of 0.4 to 0.6ms, while skipping 28% to 34% of all particles. It seems that at the very least particle-heavy servers do get a noticeable performance increase.

[debug class used](https://gist.github.com/AnOpenSauceDev/0ee77acaa65766bb2ce08cb845142778#file-debug-java)

douira:
Interesting that servers send so many particles outside of the fog distance

AnOpenSauceDev:
The worst-case scenario is that after joining a LAN server hosted on my home PC, the particles I'm spawning with a command block still render at over 100 blocks outside of the render distance (2 again) on the laptop.

### No Fog, No Fog Occlusion



command: `particle minecraft:scrape ~ ~15 ~ 0 0 0 30 500 force` (with `normal`, particles won't be sent past 32 blocks regardless of render distance.)

### Fog, No Fog Occlusion



It also seems that **particles are rendering over the stars** when in fog. This behavior also happens in Vanilla.

### Fog + Fog Occlusion



This is fixed with Particle Fog Occlusion.

AnOpenSauceDev:
It seems that particles actually do render over the fog itself (w/o Fog Occlusion), but since they use the sky fog color it's not visible until sunrise/sunset. This then causes some ugly transitions with long-distance particles like campfire smoke. 



Side note: particles don't render behind translucent blocks/fluids, but I think this is intended behavior.

### Benchmarking

In the _somewhat plausible_ situation someone has buildings outside of render distance with lots of campfires going, performance can increase by almost 2X. (Rendering the same setup from the first image, a 15x15 fill of campfires)



Rendering the scene inside of the fog distance looks like it has the same performance, so at the very least the occlusion check has a negligible impact.

The performance gains from this are going to be extremely situational, probably only having an effect on large multiplayer servers, some large builds, and mods/datapacks that make use of long-distance effects (i.e. some biomes from Terralith and Better End.), but it could fix some existing visual issues with particles.

Felix14-v2:
> probably only having an effect on large multiplayer servers, some large builds, and mods/datapacks that make use of long-distance effects

Will underwater fog and fog from blindness effect also affect particle culling? This should significantly improve performance in many mini-games related to these effects.

AnOpenSauceDev:
As of that latest commit, Darkness + Blindness now have an effect on culling. (I forgot to use the fog end distance 😔)

Water fog didn't drop the particle count by much after entering it, but I need to test that properly later.

AnOpenSauceDev:
It looks like water fog isn't actually getting rid of particles because of fog, and I have made a few observations:

1) it looks like after the water fog transition stops, almost all of the world is still visible (at least at 4 chunks), but the fog + overlay just makes it _much_ harder to see. Because of this, the particle count won't really change from being in water past the first few seconds after entering it.

2) despite this, no particles will be visible, as particles currently don't appear through translucent blocks/fluids after rendering. This actually causes thousands of particles to be waste time rendering when not submerged in an ocean. I have a few (probably too slow) ideas on how to stop the rendering anyways, but that doesn't apply to Fog Occlusion anymore.

Because of that being an unrelated problem, this feature is complete and working properly. The only extra thing I could do is also factor in height for distance to cull particles in a sphere instead of a cylinder (This didn't really provide much of a boost in render time in my survival world though).

douira:
Can this be rebased to dev?

douira:
Sorry for taking so long on this. Is this workable on 1.21.11/26.1?

AnOpenSauceDev:
~~My setup is a mess so I currently can't make any changes outside ruining history with the web editor,~~ but a direct port of the code is:

```java
@Mixin(QuadParticleGroup.class)
public abstract class QuadParticleGroupMixin {

    @Redirect(method = "extractRenderState",at = @At(value = "INVOKE", target = "Lnet/minecraft/client/particle/SingleQuadParticle;extract(Lnet/minecraft/client/renderer/state/QuadParticleRenderState;Lnet/minecraft/client/Camera;F)V"))
    public void sodium$checkFogOcclusion(SingleQuadParticle instance, QuadParticleRenderState particleTypeRenderState, Camera camera, float f){
        if(SodiumClientMod.options().performance.useFogOcclusion ? sodium$isParticleFogOccluded(camera.position(),instance.getBoundingBox().getMaxPosition()) : true){
            instance.extract(particleTypeRenderState,camera,f);
        }
    }

    @Unique
    public boolean sodium$isParticleFogOccluded(Vec3 pointA, Vec3 pointB){
        double dx = pointA.x - pointB.x;
        double dz = pointA.z - pointB.z;
        double distance = (dx * dx + dz * dz);

        var shaderFog = ((FogStorage) Minecraft.getInstance().gameRenderer).sodium$getFogParameters();
        double shaderFogDistance = shaderFog.environmentalEnd();

        var renderDistance = Minecraft.getInstance().gameRenderer.getRenderDistance();

        if(!Mth.equal(shaderFog.alpha(),1.0f)){
            return false;
        }

        var fogDist = Math.min(renderDistance,shaderFogDistance);

        return distance < fogDist * fogDist;
    }

}
```

The environment fog's end seems to be safe to use (and also has better fade-in), but can look a _tiny bit_ weird in the sunset, less so than vanilla though. I took these screenshots just using the render end because of this.

This doesn't change the particle (`P:`) count, but as you can see by the performance graphs, the CPU is definitely not trying to render them.

image

> Fog occlusion off (particles visible against skybox like before)

image

> Fog occlusion on (particles hidden correctly)

PR #2771, opened 2024-09-23
Cache multi draw batches on regions

douira:
I get an improvement of around 7% by caching multi draw batches on each region. The margin is somewhat smaller when in motion, but often the batch can still be reused. The increase in memory use seems to be minimal.

If only one section in the region changed, only re-writing the command buffer starting at that section might bring an improvement of 50% on average for regions with block updates, but the metadata management for such partial buffer writes is somewhat complicated and may not be worth it if we're going to be changing this part of the renderer anyway.

dev
dev

dev with this patch
dev with patch

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/device/GLRenderDevice.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/device/GLRenderDevice.java
@@ -274,7 +274,7 @@ public class GLRenderDevice implements RenderDevice {
                     batch.pElementCount,
                     indexType.getFormatId(),
                     batch.pElementPointer,
-                    batch.size(),
+                    batch.size,
                     batch.pBaseVertex);
         }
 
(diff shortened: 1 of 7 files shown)


douira:
I've measured the cache hit ratios for some situations. Average over 100 frames (values eyeballed), limited to 150 fps, average world. The hit ratio will be better the higher the fps, but even with 30 fps the hit ratios were generally good.

not weighted by batch size (number of cached batches / total)
- doing nothing: 1
- random rare chunk updates: 0.98
- flying around rapidly: 0.75
- moving camera around on ground level: 0.65
- violently moving the camera around: 0.45
 
weighted by batch size (number of commands from cached batches / total)
- doing nothing: 1
- random rare chunk updates: 0.98
- flying around rapidly: 0.8
- moving camera around on ground level: 0.75
- violently moving the camera around: 0.52

douira:
I hope I found all the places where cache invalidation needs to happen, but that's something to look out for during review.

jellysquid3 (approved):
Looks fine to me, we can merge this. Can you fixup the merge conflicts? These changes didn't cleanly apply on top of your others.

PR #2773, opened 2024-09-24
no-op when being used on a server

IThundxr:
Mods should not crash on servers even if they are purely client only, they should simply do nothing.


jellysquid3:
Seems reasonable enough to me. Thanks.

PR #2780, opened 2024-09-29
Sort render lists for regions and sections after traversal (#2266)

### LINKED ISSUE #2266: Section sorting is incorrect
### Bug Description

In some rare cases, the section sorting is (still) incorrect and will cause a particular section of stained glass to not render.

image
Bottom right section of the stained-glass wall is missing. See how the missing section boundary does not align with the stained glass block pattern. Screenshot taken with #2016, but can be replicated on commit 17a17ede

The missing section is position/angle dependent. It is not time or reload dependent (f3-a or full reload will determistically show the same missing section at the same place)

The bug is not due to #38, as it does not get fixed by Douira's translucency sorting (#2016)
The bug did not get fixed at the same time as #2234 17a17ede

### Reproduction Steps

1. gradlew runclient
2. create a new world with seed 4957284180073197783 
4. /execute in minecraft:overworld run tp @s -24.50 77.00 -53.50 117.00 2.40
3. /fill -64 63 -44 -64 100 -140 minecraft:white_stained_glass

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/13978385/latest.log)

### Crash Report

[crash-2024-01-18_10.17.13-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/13978399/crash-2024-01-18_10.17.13-client.txt)

douira:
This fixes #2266 by sorting regions after render list generation. The summary of the bug is that it happens because from certain points of view, the frustum is such that the order in which two regions are visited by equally distant sections depends solely on their order in the queue. See [the analysis here](https://discord.com/channels/602796788608401408/651120262129123330/1289730901487652928) for more details and images.

My measurements suggest that sorting regions by distance takes around 5µs and in total generating render lists takes around 1200µs for this scene. So the cost of sorting regions is around 0.5% which seems very acceptable. Measuring something that only takes 5µs with nanoTime is probably not the most precise, but it's averaged and very fast either way.

Obviated by https://github.com/CaffeineMC/sodium/pull/2887, but I hope this PR can be merged far earlier than the other one so that rendering is less broken until it gets merged. I'll resolve conflicts when the time comes.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
@@ -64,15 +64,15 @@ public class RenderRegion {
         return SectionPos.asLong(x, y, z);
     }
 
-    public int getRawX() {
+    public int getX() {
         return this.x;
     }
 
-    public int getRawY() {
+    public int getY() {
         return this.y;
     }
 
-    public int getRawZ() {
+    public int getZ() {
         return this.z;
     }
 
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
@@ -24,13 +24,13 @@ public class RenderRegion {
     public static final int REGION_HEIGHT = 4;
     public static final int REGION_LENGTH = 8;
 
-    private static final int REGION_WIDTH_M = RenderRegion.REGION_WIDTH - 1;
-    private static final int REGION_HEIGHT_M = RenderRegion.REGION_HEIGHT - 1;
-    private static final int REGION_LENGTH_M = RenderRegion.REGION_LENGTH - 1;
+    public static final int REGION_WIDTH_M = RenderRegion.REGION_WIDTH - 1;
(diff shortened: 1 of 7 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
```diff
@@ -64,6 +64,18 @@ public static long key(int x, int y, int z) {
         return SectionPos.asLong(x, y, z);
     }
 
+    public int getRawX() {
```
jellysquid3: Any reason this is `getRawX` instead of just `getX`?

jellysquid3 (approved):
Seems fine to me.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
douira: I thought it would be more appropriate since it's in a different coordinate space (not section space) but I can change it. It's not a very good name.

douira:
Renamed those methods, good to go now.

MeeniMc:
confirm it fixes #2266

douira:
I've added code to sort sections as well since they were suffering from the same problem as regions, also caused by subtle errors in the ordering produced by the bfs near the frustum boundary.
In my test scene with 4875 sections, the render list finalization only takes 24µs, of which 7µs are spent on region sorting and the rest on section sorting.

PR #2811, opened 2024-10-11
Add angle-based section visibility path occlusion

douira:
In my test scene it reduces section count from 4875 down to 4370 sections (11%), and bfs time by 13%.


jellysquid3 (approved):
The changes as proposed look good to me. From testing, it doesn't seem to introduce any errors, but we'll do a smoke test before merging to be certain.

PR #2823, opened 2024-10-19
Add epsilon to section distance check

burgerindividual:
This is done to for parity with the epsilon used on frustum culling. ~~The epsilon is pretty pessimistic, and can be less pessimistic if the algorithm is changed slightly. I chose not to do this because it may make the function slower.~~

~~I chose not to add this value to the `searchDistance` value provided to `OcclusionCuller.findVisible`, because that could potentially change the sections traversed by the BFS, which I'm not sure we want.~~

The epsilon used for this function is `1.0` on each axis, enough to account for large block models, while the the frustum cull uses `1.125`. The extra `0.125` exists in the epsilon of the frustum cull due to floating point imprecision, which shouldn't come up with the minimal amount of floating point operations in the distance check.

With this used, I think the `+ 0.5` epsilon [here](https://github.com/CaffeineMC/sodium-fabric/blob/dd25399c139004e863beb8a2195b9d80b847d95c/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java#L639) may be able to be removed, also. I need confirmation on this to be sure.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
@@ -158,11 +158,10 @@ public class OcclusionCuller {
 
         // coordinates of the point to compare (in view space)
         // this is the closest point within the bounding box to the center (0, 0, 0)
-        float dx = nearestToZero(ox, ox + 16) - camera.fracX;
-        float dy = nearestToZero(oy, oy + 16) - camera.fracY;
-        float dz = nearestToZero(oz, oz + 16) - camera.fracZ;
-
-        maxDistance += DISTANCE_EPSILON;
+        // the bounding box is expanded by 1 block in each direction due to the maximum allowed size of block models.
+        float dx = nearestToZero(ox - 1, ox + 17) - camera.fracX;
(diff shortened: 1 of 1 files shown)


douira:
I think this culls less than it should, because models can't be as big as a sphere enclosing a 3x3x3 box, rather they can at most be a 3x3x3 box.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -179,7 +180,8 @@ private static int nearestToZero(int min, int max) {
     // The bounding box of a chunk section must be large enough to contain all possible geometry within it. Block models
     // can extend outside a block volume by +/- 1.0 blocks on all axis. Additionally, we make use of a small epsilon
     // to deal with floating point imprecision during a frustum check (see GH#2132).
-    private static final float CHUNK_SECTION_SIZE = 8.0f /* chunk bounds */ + 1.0f /* maximum model extent */ + 0.125f /* epsilon */;
+    private static final float CHUNK_SECTION_SIZE = 8.0f /* chunk bounds */ + 1.0f /* maximum model extent */
```
douira: this part of the diff seems unnecessary

douira (commented):
looks good

douira (approved):

jellysquid3 (approved):

Issue #2696, opened 2024-08-20
Entities become invisible in culled chunks when they shouldn't


Felix14-v2:
### Bug Description

In some scenarios, if an entity is partially hidden in a culled chunk section, the entity model can become invisible until you forcefully load the chunk section.

<details>
  <summary> Demo </summary>

https://github.com/user-attachments/assets/a0998bef-0e81-4412-a912-d6a524b7a97a

The interesting part is that disabling the "Entity culling" option does nothing:

https://github.com/user-attachments/assets/cf2c41db-c850-409a-979a-abaeaaa8d22a
</details>

### Reproduction Steps

**Preparing the area** (you can just load the [attached world](https://github.com/user-attachments/files/16681146/Culling.bug.zip))
- Create an empty world with the "void" preset
- Cover a chunk section with a solid block (example: `/fill -1 -1 -1 16 16 16 stone`)
- Unload the chunk and load it again in a way you prefer (for example, just rejoin the world)
**Triggering the bug**
- Place an entity on top of the cube you've just created and disable gravity for it (example: `/summon minecraft:armor_stand 5 17 5 {NoGravity:1b}`)
- Tweak the vertical position of the entity to hide its hitbox origin inside the chunk section (example: `/execute as @e[type=minecraft:armor_stand] at @s run tp ~ ~-1.001 ~`)
- The entity will immediately disappear (expected: it should be visible)

### Debug info
[latest.log](https://github.com/user-attachments/files/16681008/latest.log)
[crash-2024-08-20_22.41.21-client.txt](https://github.com/user-attachments/files/16681011/crash-2024-08-20_22.41.21-client.txt)

### Additional
I heard about this behavior from a server administrator, who complained to me that Sodium breaks decorations on his server. I debugged it a bit and found this bug. So this may be a real-world scenario, not just a limited edge case.

https://github.com/user-attachments/assets/a1199ebc-4670-4285-bdb7-299aeafe7157

[Lobby.zip](https://github.com/user-attachments/files/16681148/Lobby.zip)

By the way, you can see another issue with the entity shadowing in the video above (0:02-0:03), but that's unrelated.

Altirix:
This is a Vanilla bug to some extent. its not the same behaviour as sodium but neither vanilla or sodium have a guarentee on rendering these entitites. This does not seem to be a result of entity culling.

[...]

PR #2839, opened 2024-10-24
Fix entities being invisible

CyberCat319:


douira:
Doesn't this defeat the point of entity culling?

IMS212:
It does. The proposed change makes it so occlusion culling has no effect on entities.

douira:
If you are experiencing problems with entity culling, please file a bug report or come to the discord server if you need support. Further, entity culling can be disabled with a setting in the video options.

Felix14-v2:
> If you are experiencing problems with entity culling, please file a bug report

A kind reminder that one already exists ._. 
https://github.com/CaffeineMC/sodium-fabric/issues/2696

PR #2879, opened 2024-11-16
Fix some cases of large BEs not rendering

douira:
Uses a model margin of 2 for neighboring sections. Do we want to also apply this to all other sections, or use a different value?

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
@@ -37,6 +37,8 @@ public class OcclusionCuller {
         while (queues.flip()) {
             processQueue(visitor, viewport, searchDistance, useOcclusionCulling, frame, queues.read(), queues.write());
         }
+
+        this.fixOriginNeighbors(visitor, viewport, searchDistance, frame);
     }
 
     private static void processQueue(Visitor visitor,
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -209,13 +211,51 @@ private static int nearestToZero(int min, int max) {
...
+    private static boolean isNeighborSectionVisible(RenderSection section, Viewport viewport, float maxDistance) {
+        return isWithinRenderDistance(viewport.getTransform(), section, maxDistance) && isWithinNeighborFrustum(viewport, section);
+    }
+
+    public static boolean isWithinNeighborFrustum(Viewport viewport, RenderSection section) {
+        return viewport.isBoxVisible(section.getCenterX(), section.getCenterY(), section.getCenterZ(),
+                CHUNK_SECTION_SIZE_NEIGHBOR, CHUNK_SECTION_SIZE_NEIGHBOR, CHUNK_SECTION_SIZE_NEIGHBOR);
+    }
+
+    private void fixOriginNeighbors(Visitor visitor, Viewport viewport, float searchDistance, int frame) {
```
jellysquid3: `fixOriginNeighbors` -> `addNearbySections`.

Probably there should also be a high-level comment here explaining that we add these nearby sections to account for oversized entities.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -209,13 +211,51 @@ private static int nearestToZero(int min, int max) {
...
+    private static final float CHUNK_SECTION_BASE = 8.0f /* chunk bounds */;
+    private static final float CHUNK_SECTION_SIZE = CHUNK_SECTION_BASE + 1.0f /* maximum model extent */ + 0.125f /* epsilon */;
+    private static final float CHUNK_SECTION_SIZE_NEIGHBOR = CHUNK_SECTION_BASE + 2.0f /* bigger model extent */ + 0.125f /* epsilon */;
 
     public static boolean isWithinFrustum(Viewport viewport, RenderSection section) {
         return viewport.isBoxVisible(section.getCenterX(), section.getCenterY(), section.getCenterZ(),
                 CHUNK_SECTION_SIZE, CHUNK_SECTION_SIZE, CHUNK_SECTION_SIZE);
     }
 
+    private static boolean isNeighborSectionVisible(RenderSection section, Viewport viewport, float maxDistance) {
```
jellysquid3: The render distance shouldn't be checked for very nearby sections, since the comparison will always be true.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -209,13 +211,51 @@ private static int nearestToZero(int min, int max) {
     // The bounding box of a chunk section must be large enough to contain all possible geometry within it. Block models
     // can extend outside a block volume by +/- 1.0 blocks on all axis. Additionally, we make use of a small epsilon
     // to deal with floating point imprecision during a frustum check (see GH#2132).
-    private static final float CHUNK_SECTION_SIZE = 8.0f /* chunk bounds */ + 1.0f /* maximum model extent */ + 0.125f /* epsilon */;
+    private static final float CHUNK_SECTION_BASE = 8.0f /* chunk bounds */;
+    private static final float CHUNK_SECTION_SIZE = CHUNK_SECTION_BASE + 1.0f /* maximum model extent */ + 0.125f /* epsilon */;
+    private static final float CHUNK_SECTION_SIZE_NEIGHBOR = CHUNK_SECTION_BASE + 2.0f /* bigger model extent */ + 0.125f /* epsilon */;
```
jellysquid3: "Neighbor" -> "Nearby". Should also have a comment explaining it's only used for nearby sections which may have oversized block models or entities.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -209,13 +211,51 @@ private static int nearestToZero(int min, int max) {
     // The bounding box of a chunk section must be large enough to contain all possible geometry within it. Block models
     // can extend outside a block volume by +/- 1.0 blocks on all axis. Additionally, we make use of a small epsilon
     // to deal with floating point imprecision during a frustum check (see GH#2132).
-    private static final float CHUNK_SECTION_SIZE = 8.0f /* chunk bounds */ + 1.0f /* maximum model extent */ + 0.125f /* epsilon */;
+    private static final float CHUNK_SECTION_BASE = 8.0f /* chunk bounds */;
```
jellysquid3: "Base" -> "Radius"

jellysquid3 (changes requested):
Some nitpicks about the terminology used. I think "neighbor" is confusing since we already use that to describe the 6-neighborhood that chunks are directly connected with.

douira:
All review items have been addressed. Rebased and force-pushed to fix an implicit conflict.

jellysquid3 (approved):

PR #2887, opened 2024-11-22
Asynchronous Graph Culling and Frame-Independent Task Scheduling

douira:
Uses an octree to generate render lists independently of the, now asynchronous, slow graph search.
- It runs the occlusion culler in a separate thread, which allows a large speedup in render list generation time. Correct culling results are ensured with a combination of different types of BFS, synchronous occlusion culling is used as a last resort if the camera teleports or moves extremely quickly.
- Tasks are ordered based on a combination score of how long they've been pending, their distance from the camera, their type, and whether they're visible in the camera frustum. How many tasks can be scheduled is now independent of the frame rate and instead are limited based on their estimated duration and size.
- There's an upload limit to ensure not too many tasks are submitted that will exceed the upload buffer's size. This isn't a hard limit and this PR doesn't implement a new way of handling task buffers, to avoid expanding the scope too far.

[Testing](https://discord.com/channels/602796788608401408/1285427417774559292) has not shown regressions and generally frame rate has improved a little if a system was not limited by render list generation, and a lot if it was. (see testing thread)

Companion PR in Iris: https://github.com/IrisShaders/Iris/pull/2539 (outdated)
Prerequisite: https://github.com/CaffeineMC/sodium/pull/3484


ghost:
Will this PR be fixed for extended render distance mods such as C2ME? Rendering engine just explodes and crashes when using it.

douira:
Please come to our [discord](https://jellysquid.me/discord) server and give a more detailed report of the issue you're experiencing in the appropriate testing thread.

ghost:
> Please come to our [discord](https://jellysquid.me/discord) server and give a more detailed report of the issue you're experiencing in the appropriate testing thread.

Discord is blocked in my country, may i report it here?

douira:
I've rebased this on top of the current `dev` and it seems to work ok for me so far. It'll require more testing and tuning since the merge conflict was rather involved.

douira:
This is ready for review. As of the recent rewrite of a lot of systems here I'm much happier with it. I'm not currently aware of any further blocking issues. I'm going to do some more testing but I think the real feedback is going to come through a beta.

PR #2893, opened 2024-11-24
Fix Culling if Origin is in an Unloaded Section

douira:
Renders the closest loaded sections if the origin section is not loaded by doing a bfs of the unloaded and empty sections starting at the origin.


douira:
Closing because #2887 will now do this better and basically for free.

PR #2907, opened 2024-12-03
Fix Fluid Occlusion For Waterlogged Blocks

### LINKED ISSUE #2699: Water is not rendered in some cases with Framed Blocks (occlusion when waterlogged)
### Bug Description

Water is not rendered in some cases with Framed Blocks (occlusion when waterlogged)

### Reproduction Steps

Place the “Framed Inner Corner Slope” frame from Framed Blocks, and waterlog it.

### Log File

https://mclo.gs/U5BAsH9

### Crash Report

*

douira:
Fixes fluid occlusion by using voxel shape comparison with the waterlogged block using BlockOcclusionCache. Using `isFaceSturdy` was wrong for occlusion purposes since it deals with hitboxes and not what is rendered. Thanks to @XFactHD for clearing this up.

Fixes https://github.com/CaffeineMC/sodium/issues/2699

I've also made a [1.21.1 branch of this](https://github.com/douira/sodium/tree/fix-fluid-occlusion) for testing purposes since FramedBlocks, which I was using to reproduce the issue, isn't available on the latest Minecraft version that `dev` uses. When backporting this to 1.21.1, you can use [the commit on that branch](https://github.com/douira/sodium/commit/8ddfd0ca634de58800b42be8d2b112590a6676f8), which slightly differs because of changes to `getFaceOcclusionShape`.

From my testing it doesn't regress the other fluid occlusion issue https://github.com/CaffeineMC/sodium/issues/2606. Visual discontinuities arising from water flowing along a row of slabs getting culled suddenly once its height goes below half a block are normal and exist in both upstream Sodium and vanilla.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockOcclusionCache.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockOcclusionCache.java
@@ -108,9 +108,9 @@ public class BlockOcclusionCache {
             var selfShape = selfBlockState.getFaceOcclusionShape(facing);
 
             // only a non-empty self-shape can occlude anything
-            if (!selfShape.isEmpty()) {
+            if (!isEmptyShape(selfShape)) {
                 // a full self-shape occludes everything
-                if (selfShape == Shapes.block() && fluidShapeIsBlock) {
+                if (isFullShape(selfShape) && fluidShapeIsBlock) {
                     return false;
                 }
 
@@ -149,13 +149,13 @@ public class BlockOcclusionCache {
         var otherShape = otherState.getFaceOcclusionShape(facing.getOpposite());
 
         // If the other block has an empty cull shape, then it cannot hide any geometry
-        if (otherShape.isEmpty()) {
+        if (isEmptyShape(otherShape)) {
             return true;
         }
 
         // If both blocks use a full-cube cull shape, then they will always hide the faces between each other.
         // No voxel shape comparison is done after this point because it's redundant with the later more accurate check.
-        return otherShape != Shapes.block() || !fluidShapeIsBlock;
+        return !isFullShape(otherShape) || !fluidShapeIsBlock;
     }
 
     private boolean lookup(VoxelShape self, VoxelShape other) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockOcclusionCache.java
```diff
@@ -88,6 +89,75 @@ private static boolean isEmptyShape(VoxelShape voxelShape) {
...
+
+        // only occlude against blocks that can potentially occlude in the first place
+        if (!otherState.canOcclude()) {
+            return true;
+        }
+
+        var otherShape = otherState.getFaceOcclusionShape(facing.getOpposite());
+
+        // If the other block has an empty cull shape, then it cannot hide any geometry
+        if (otherShape.isEmpty()) {
```
jellysquid3: `VoxelShape#isEmpty` can be slow since it involves many virtual dispatches and pointer chases. There should be a check for `shape == Shapes.empty()` before calling it.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockOcclusionCache.java
```diff
@@ -88,6 +89,75 @@ private static boolean isEmptyShape(VoxelShape voxelShape) {
...
+     */
+    public boolean shouldDrawFullBlockFluidSide(BlockState selfBlockState, BlockGetter view, BlockPos selfPos, Direction facing, FluidState fluid, VoxelShape fluidShape) {
+        var fluidShapeIsBlock = fluidShape == Shapes.block();
+
+        // only perform self-occlusion if the own block state can't occlude
+        if (selfBlockState.canOcclude()) {
+            var selfShape = selfBlockState.getFaceOcclusionShape(facing);
+
+            // only a non-empty self-shape can occlude anything
+            if (!selfShape.isEmpty()) {
```
jellysquid3: See comment about `VoxelShape#isEmpty`. Probably there should be a utility function to check if the shape is empty.

jellysquid3 (changes requested):

jellysquid3:
This patch looks good outside of the aforementioned comments.

douira:
Resolved review items by using the existing utility methods that check for empty or full shapes.

jellysquid3 (approved):

jellysquid3:
Thanks!

PR #2924, opened 2024-12-07
Fix Section Sorting by Using the Correct Section Coordinate

### LINKED ISSUE #2918: Translucency sorting uses wrong camera position
### Bug Description

Translucency sorting seems broken

image

### Reproduction Steps

with RestWorld: https://claritypack.com/restworld/RestWorld_1.21.4.zip

`/execute in minecraft:overworld run tp @s -0.80 103.29 34.34 -361.50 10.05`

### Log File

*

### Crash Report

*

douira:
Fix section and region sorting by using the correct section coordinate instead of the integer part of the camera transform which is incorrect near the origin.

Fixes #2918


jellysquid3:
Ah, that makes more sense now. Thanks!

PR #2931, opened 2024-12-10
Improve Fluid Shaping

### LINKED ISSUE #1210: Fix fluids not having bottom texture planes on the inside
### Version information

sodium-fabric-mc1.18.2-0.4.1+build.15

### Expected Behavior

Since the other five faces exist, the bottom should as well.

### Actual Behavior

It does not under any circumstances.

### Reproduction Steps

1. Place water above a pressure plate, with the water surrounded by four top slabs for easy access
2. Enter the water

### Java version

17.0.2 64bit

### CPU

4× Intel® Core™ i3-3240 CPU @ 3.40GHz

### GPU

Intel® HD Graphics 2500

### Additional information

Vanilla bug https://bugs.mojang.com/browse/MC-190053, which has persisted since 1.7. Versions prior to 1.7 did not have this bug.

This has been split from #91 due to changes made to fluid rendering in 1.18.2.

### LINKED ISSUE #1211: Fix water in waterlogged blocks not appearing "flat" when expected
### Version information

sodium-fabric-mc1.18.2-0.4.1+build.15

### Expected Behavior

The surface of the water should appear completely straight, much like it does in this geometrically equivalent slab arrangement:

Or like it does when the arrangement is surrounded by full blocks:

### Actual Behavior

The water level does not appear even, instead bulging in the center.

### Reproduction Steps

1. Create an arrangement of four stairs as to create an "offset slab"
2. Waterlog these four stair blocks
3. Stare at the water

### Java version

17.0.2 64bit

### CPU

4× Intel® Core™ i3-3240 CPU @ 3.40GHz

### GPU

Intel® HD Graphics 2500

### Additional information

Vanilla bug https://bugs.mojang.com/browse/MC-127270. It seems that 1.18.2 made this effect less intense, but it's still there.

This seems to arise because the water doesn't have any full blocks outside of it for it to appear at a higher level, despite it being contained by solid blocks in its own space.

### LINKED ISSUE #1212: Fix water trying to connect to waterlogged blocks it cannot
### Version information

sodium-fabric-mc1.18.2-0.4.1+build.15

### Expected Behavior

Since there is a physical barrier between the flowing water and still water, they should not appear connected.

### Actual Behavior

They appear connected anyway. In addition, the still water appears to flow into the flowing water.

### Reproduction Steps

1. Create the setup shown in the screenshots above, with water in all stair blocks and a single source at the end
2. Observe the shape of the flowing water

### Java version

17.0.2 64bit

### CPU

4× Intel® Core™ i3-3240 CPU @ 3.40GHz

### GPU

Intel® HD Graphics 2500

### Additional information

Vanilla bug https://bugs.mojang.com/browse/MC-127030. Split from #91 due to some changes made in 1.18.2 that fixed similar bugs mentioned in that report.

The water being taller than expected is clearly a rendering artifact, as using Spectator mode reveals that the fog and FOV changes only apply where the expected water height is.

### LINKED ISSUE #2470: Fix weird culling of fluid interior top planes
### Request Description

Build the following structure:

Now enter it in Spectator mode:

You'll notice something strange pretty quickly - the top faces of the water are visible from outside (as they should be, as they aren't coplanar with the ceiling and therefore hidden), but are completely invisible from the inside.

Things get even weirder if you knock out one of the corner blocks - one of the planes will be visible, whereas the other will not, separated by an abrupt cutoff at the block edge.

Here's another setup demonstrating much the same issue:

This is vanilla bug https://bugs.mojang.com/browse/MC-81654. May be within the scope of Sodium to fix - this is game graphics being weird and unintuitive, after all.

### LINKED ISSUE #2935: Better heuristic for culling fluid top planes in caves
### Request Description

This is a requirement if we plan on fixing #2470 at all.

Currently, for a given fluid block at elevation y in the world, the game will check if the block at y+1 and the eight blocks in its Moore neighbourhood are all solid (or contain the same fluid); if so, the interior top face is culled. If not, the face is rendered. This works fine for the most part, but causes oddities in some reasonably common situations.

My preferred solution would be the change proposed in #2472, although it has been stated that such a change may introduce other issues, so a different solution may be required; this will need to be discussed.

For cases like the following, the 16 faces underneath the stone blocks would preferably get culled; whether this also happens to the nine in the middle is a perhaps more controversial issue I may explain in a future report.

### LINKED ISSUE #2947: Rendering inner fluid faces against blocks should be consistent directionally
### Request Description

Relates to #2935.

For solid blocks, the inner faces of fluids typically are not rendered.
Image

There are, however, exceptions such as glass: while the top and four side faces will not have the fluid render against them, the bottom face will. This is inconsistent with many opaque blocks.
Image
Image

The behaviour should ideally be consistent: either fluid faces wouldn't render against any of glass's faces, or they should render against all of them. The former behaviour appears to be preferable.

Solving this issue seems tricky, however, without special-casing certain block IDs or introducing behaviour with the potential of causing equivalent or larger incongruences.

### LINKED ISSUE #3523: Water overlay texture is cutoff if texture is more than 16x16 and is an animation
### Bug Description

Starting with 0.8.6 1.21.11, the water_overlay.png texture stops rendering at 1/4 of the intended surface when the texture is more than 16x16 like the below example which is 32x32 and is animated. Non animated textures _seem_ work as intended.

Vanilla/prior Sodium versions:
Image

Sodium 0.8.6:
Image

Reason the 32x32 texture I'm using still looks 16x16 is to work around the uv map of the hardcoded model that only uses 1/4 of the actual texture, so if I put a 16x16 on the upper left corner of the texture, it will look like a normal 16x16 texture does. This may be an intended optimization, but there are conflicts with resource packs that expect vanilla behavior shown above.

### Reproduction Steps

Look at the water_overlay texture placing water against glass using a resource pack that uses an animated 32x32 texture.

### Log File

[2026-02-24-7.log](https://github.com/user-attachments/files/25534744/2026-02-24-7.log)

### Crash Report

[crash-2026-02-24_20.16.57-client.txt](https://github.com/user-attachments/files/25534746/crash-2026-02-24_20.16.57-client.txt)

douira:
Refactors the fluid renderer to improve fluid shaping in a variety of situations. Profiling shows that the performance impact appears to be negligible.

Screenshots are ordered dev and then this patch.

Regular water slopes look the same.
water 1 dev
water 1 patch
water 2 dev
water 2 patch

Water slopes are nearly identical. I experimented with changing the sloping behavior and it didn't work out well.
water 3 dev
water 3 patch

Water shaping around stairs (foreground) and between stairs (further back): Water height doesn't propagate through the closed sides of block faces, fixing issues with unnecessarily bent water faces and water connecting to waterlogged blocks where it shouldn't.
Fixes https://github.com/CaffeineMC/sodium/issues/1211
water 4 dev
water 4 patch
water 5 dev
water 5 patch

The cave heuristic has been refined to search 2 blocks around and 1 block up (compared to 1 block around and 1 block up). This improves the appearance of inward up face culling somewhat. Player builds are still somewhat weird but generally improved.
To make the presentation more self-consistent and to avoid situations where seams in the fluid are visible from air spaces that the player could inhabit, some cases now have more fluid quads than before. In other cases, less faces are shown to make glass rendering more consistent. Glass being right above water doesn't constitute enough "air" to make it render inwards facing quads, but outwards facing quads will render to ensure the fluid is visible when viewing it through the glass from above.
Fixes https://github.com/CaffeineMC/sodium/issues/2470, nearly completely, to the point where I would call it fixed.
Fixes https://github.com/CaffeineMC/sodium/issues/2935 by providing a feasible yet pretty good cull heuristic.
Fixes https://github.com/CaffeineMC/sodium/issues/2947 by hiding down-facing fluid up quads against glass in all reasonable cases.
dev
patch
water 6 dev
water 6 patch

Outward up faces are now culled with the refined cave heuristic. This slightly increases rendering performance if there's a lot of flooded visible caves. I have reports of increased fps anywhere up to around 2 to 6%.
water 8 dev
water 8 patch

The test for bottom quad occlusion to show inward water quads on air, but not on glass for example. By using the same test for overlays as the sides, this more consistently shows and hides the inward bottom face.
Fixes https://github.com/CaffeineMC/sodium/issues/1210
Fixes https://github.com/CaffeineMC/sodium/issues/1212
water 7 dev
water 7 patch

Fixes https://github.com/CaffeineMC/sodium/issues/3523


douira:
ok I'm marking this as ready, what follows will be fixing any bugs we might find in further testing.

douira:
I've fixed a small bug with the flooded cave heuristic and added detailed documentation to the PR and the fluid renderer class itself.

IMS212 (approved):
Seems fine to me. Not very knowledgeable about fluid rendering, though.

douira:
Thanks! I've done a few rounds of testing and feedback. If something significant pops up after release, we can fix it.

PR #2956, opened 2024-12-29
Remove seemingly unused graphic

muzikbike:
arrows.png has existed in assets/sodium/textures/gui since 2020, but I can't find any piece of code referencing it, nor anything that looks like it actually being used in the UI. It also wasn't split into two sprites following the 1.20.2 reworking of UI textures, making it seem even more like a leftover.


muzikbike:
According to https://discord.com/channels/602796788608401408/752637878462578778/1235782025063960598 this was indeed never used in a release version.

jellysquid3:
This little graphic has a small history and isn't hurting anyone, so we're going to keep it. :)

PR #2960, opened 2025-01-01
Faster Particle & Cuboid vertex build

MoePus:
For particles: use camera's left vector and up vector to build vertices to avoid Quat transfroms.
For cuboids: use vectors to calc vertices to avoid some muls

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/EntityRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/EntityRenderer.java
@@ -122,14 +122,27 @@ public class EntityRenderer {
     }
 
     private static void prepareVertices(PoseStack.Pose matrices, ModelCuboid cuboid) {
-        buildVertexPosition(CUBE_CORNERS[VERTEX_X1_Y1_Z1], cuboid.x1, cuboid.y1, cuboid.z1, matrices.pose());
-        buildVertexPosition(CUBE_CORNERS[VERTEX_X2_Y1_Z1], cuboid.x2, cuboid.y1, cuboid.z1, matrices.pose());
-        buildVertexPosition(CUBE_CORNERS[VERTEX_X2_Y2_Z1], cuboid.x2, cuboid.y2, cuboid.z1, matrices.pose());
-        buildVertexPosition(CUBE_CORNERS[VERTEX_X1_Y2_Z1], cuboid.x1, cuboid.y2, cuboid.z1, matrices.pose());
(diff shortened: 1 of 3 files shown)


jellysquid3:
Hi. Thanks for the pull request.

Your changes look interesting. These floating point ops are a rather significant part of the runtime (as Java can't vectorize or schedule the instructions correctly) and I would expect this patch set improves meshing performance quite a bit.

MoePus:
Hi,

I’m really glad to see that my code contributes to improving Minecraft's performance. The changes I submitted originate from my addon project for the Forge Sodium port on 1.20.1. The addon includes both non-aggressive optimizations, which are designed to be safe and should be side-effect-free, and some more aggressive changes that focus on maximizing FPS improvements.

I plan to upstream all non-aggressive optimizations to Sodium, while continuing to maintain the aggressive changes in my addon. Additionally, I’ll be updating the addon to support Sodium on 1.21.x Neoforge.

Would it be possible to incorporate some of Sodium’s recent improvements into my addon to further enhance compatibility and performance for Forge users?

Thank you for your time and consideration!

Best regards

jellysquid3:
I've merged the changes you've proposed separately and with some further modification. (https://github.com/CaffeineMC/sodium/commit/7f25220d810cd002a3a72b8b0a1e881d21855ca9, https://github.com/CaffeineMC/sodium/commit/1b0f7b9150ae7dfb092a011cce12300d2d8ab248)

The entity rendering patches did not seem to improve performance on my machine, because of bad code generation from the compiler. There was a lot of work needed to optimize the existing code to help free the bottleneck and make your optimizations take effect. After everything, entity rendering is anywhere from 8 to 15% faster on my machine, with the trick you suggested being about ~40% of that runtime improvement.

As for the billboard particle rendering patches, no further changes were made, and they seemed to help the rendering performance by around ~10% (though lighting calculations are still by far the slowest thing.)

jellysquid3:
> Would it be possible to incorporate some of Sodium’s recent improvements into my addon to further enhance compatibility and performance for Forge users?

Sodium 0.6+ is licensed under the [Polyform Shield](https://github.com/CaffeineMC/sodium/blob/1b0f7b9150ae7dfb092a011cce12300d2d8ab248/LICENSE.md), which only stipulates that you cannot use our code to compete with upstream. If your addon depends on Sodium to run, then it probably wouldn't be competition.

That said, Sodium also has an official port for NeoForge, so if your addon is installed alongside Sodium, you would probably not need to incorporate anything into your mod. Of course, you still retain the original copyright of any code you submit to our project, so you can license it for your own needs however you see fit.

PR #2964, opened 2025-01-03
Hopper model texture renames to match vanilla

Apollounknowndev:


jellysquid3:
Hm, okay. Seems we missed this one when renaming the Cauldron/Composter/etc. Thanks.

PR #2966, opened 2025-01-04
Properly Fix Hopper Texture Names

douira:
Properly fix hopper texture names to make the hopper model not show a missing texture. Previous PR #2964 was incomplete.


jellysquid3:
weh

PR #2972, opened 2025-01-07
Mixed Radix Vertex Sorting

douira:
Use indirect and direct (packed) radix sorting, which according to my measurements is around 10% faster.

Screenshot_2025-01-06_at_23 23 02


douira:
There's been [work to further improve the sorting](https://discord.com/channels/602796788608401408/651120262129123330/1326331136909119488) and this code is no longer needed.

PR #2982, opened 2025-01-14
Backface culling on items/item entities

Taiyou06:
Adds backface culling on the items and item entities, as recently with the addition of display entity stuff, servers started to using those to make custom content. This PR should be 10-20% (in general) faster compared to 0.6.6

This is my first time PRing to sodium, so if i missed something or need anything or did something wrong, I will do my best to help.

One thing to note is that i have no idea on how to make `isFacingAway` any faster, so suggestions for that would be appreciated!


jellysquid3:
As discussed in the Discord server, there are some concerns this would have for mod compatibility. The biggest issue as I see it would be that we can't detect whether the render state is _actually_ configured for back-face culling, since mods can decide to only set that state _after_ writing vertex data.

Without moving the early back-face culling to right before draw call submission, it is unlikely that could be solved. And it would negate any performance improvement, while also making it impossible to write vertex data directly into GPU-visible memory (i.e. as might be used for zero-copy uploads.)

Taiyou06:
Are there any example mods that could be causing the said issues for replication? I would love to investigate solutions for this. 
Maybe some API solution through sodium to explicitly disable the behaviour etc.

Issue #2802, opened 2024-10-06
Translucent block models that extend past 1x1x1 may not render correctly


Altirix:
### Bug Description

This is also a bug within vanilla. In one case however sodium is producing the right result.

There seem to be two categories of this bug.

- models with intersecting translucent quads.
- models that extend outside of their chunk.

### Reproduction Steps

[testpack.zip](https://github.com/user-attachments/files/17268804/testpack.zip)

| Test Scene |Sodium 0.6b2 | Vanilla |
| ------------- | ------------- | ------------- |
| 3x3x3 translucent intersecting with water | ![2024-10-06_10 43 07](https://github.com/user-attachments/assets/e8b41a7b-e031-431b-a204-3b34796f8b79)  |  ![2024-10-06_10 44 29](https://github.com/user-attachments/assets/d6a2d0fc-0ce0-4dfe-ab3e-09bc5b081e64) |
| 3x3x3 translucent intersecting multiple |  ![2024-10-06_10 43 57](https://github.com/user-attachments/assets/ccf17251-2179-4e96-8376-21d17ed229d3) | N/A  |
| 3x3x3 translucent intersecting neigbour chunk | ![2024-10-06_10 45 50](https://github.com/user-attachments/assets/ef6074f8-c03c-4945-ae5b-94d6b0d08979) ![2024-10-06_10 45 52](https://github.com/user-attachments/assets/892d9e7b-859f-4ad8-ab75-209abb078aaf) | ![2024-10-06_10 44 48](https://github.com/user-attachments/assets/8ada09cc-5611-40fb-b2ac-a48679591dec) ![2024-10-06_10 44 49](https://github.com/user-attachments/assets/74e2e93c-4f21-4a31-872e-aa9c290ea28b) |

### Log File

N/A

### Crash Report

N/A

douira:
Even regular block-sized 1x1x1 models that intersect the water surface or have internal intersections are rendered wrongly. This is because order-based translucency sorting of quads that intersect is fundamentally impossible without splitting the quads into pieces along their intersection planes. Such splitting requires a bunch of math to figure out the pieces, and then writing more quads with some interpolated vertexes. I think it's possible but requires a significant conceptualization and engineering effort.
The case where geometry is wrongly sorted if it extends outside of the section would be fixed by splitting quads along the section's boundaries and rendering the quads that are then outside of the section as part of the neighboring sections. (this is another layer of complication)

btw I think one of your last screenshots got put into the wrong column.

[...]

PR #2993, opened 2025-01-24
Translucency Sorting With Quad Splitting

### LINKED ISSUE #2943: Translucency sorting for intersecting translucent planes
### Request Description

Relates to #2802. Reported to vanilla under https://bugs.mojang.com/browse/MC-277806 and will probably never be fixed there.

This is an edge case of translucency sorting which has no good solution other than clipping the intersecting planes into smaller triangles which can be sorted more easily. Along with having a performance cost, there probably isn't a very easy way to determine how this should even be done.
Image
Image

douira:
- Fixes translucency sorting for geometry that is intersecting or otherwise unsortable by splitting quads in the partition tree.
- Introduces a setting with the states "off", "safe", "unlimited" that controls to which degree quads are split. The default is "safe", where it doesn't disable normal quantization and limits how many quads can be generated compared to the original number. (i.e. the geometry amplification is capped)
- Removes some unnecessary classes from the translucent data class hierarchy
- Simplifies vertex/quad count handling and translucent data object construction
- Reorders translucent sorting to be before meshing so that modifications to meshing can be performed
- Refactors `TQuad` construction to happen mostly in its own class

A 1.21.1 version of this patch is available at https://github.com/douira/sodium/tree/1.21.1/translucent-quad-splitting (currently outdated)

Fixes https://github.com/CaffeineMC/sodium/issues/2943

overview
split


jellysquid3:
> Also fixes an old mistake in the sort triggering code in the form of a wrong equals implementation on AlignableNormal.

Would it make sense to pull this patch out of this pull request so we can backport it? I'm not sure what it is breaking exactly.

douira:
It does break things that require sort triggering on unaligned normals but it's unlikely anyone would notice unless the geometry is particularly weird. It only becomes more noticeable with this patch because previously the partition tree wouldn't generate unaligned partition (i.e. triggering) planes, and unaligned trigger planes were only generated by the super rarely used dynamic topo sorting mode.

I could split it out if you'd merge such a fix before 0.7, which is when I'd expect this to be ready for as a whole.

muzikbike:
Does this fix #2943? May be worth marking as such if so.

muzikbike:
Does this also fix #2802? Images posted in the discord thread certainly appear to imply so.

jellysquid3:
> Does this also fix https://github.com/CaffeineMC/sodium/issues/2802?

Not completely. This only splits geometry based on the information available in a per-chunk basis. There is no splitting that happens across chunk boundaries, so any over-sized model with translucent geometry will render incorrectly if it crosses into another chunk. Likely the problem will never be fixed, since it would complicate render code dramatically for what is arguably a broken model.

douira:
This is ready for review and merging. No open in-scope issues found during testing remain.

Note that the bug fix PR https://github.com/CaffeineMC/sodium/pull/3033 is included in this for testing reasons and once it gets merged into upstream I'll drop it from this.

PR #2997, opened 2025-01-28
Fix Unaligned Sort Triggering

douira:
Fixes a bug with the plane-based triggering system where unaligned trigger planes would be wrongly not be added to the triggering system if the same same normal was already present.

This was caused by a buggy equals implementation in AlignableNormal.

Since unaligned trigger planes are only used by the very rarely employed dynamic topo sorting mode, this is unlikely to have any actual effects until quad splitting https://github.com/CaffeineMC/sodium/pull/2993, which makes use of unaligned trigger planes more commonly, is added.

I have gotten reports (at least one, maybe two) of crashes within the sort triggering code (months ago), but I don't know whether those were related to this (and I can't find them anymore on discord).

This PR is a cherry pick of commit https://github.com/CaffeineMC/sodium/pull/2993/commits/d3f4953b665927d937a75f2281028a051d34198f, originally authored as part of https://github.com/CaffeineMC/sodium/pull/2993. When/if this gets merged, that PR will probably require merge conflict resolution.


jellysquid3:
Thanks, this seems good overall. And I'm happy that we aren't subclassing Vector3f anymore.

jellysquid3 (approved):

douira:
I've added two small bug fix commits. The first for fixing an old but obscure bug that only has relevance with quad splitting, but technically did also exist otherwise. The second corrects a mistake I made during the refactor in this PR. The hashmap of unaligned normals was getting corrupted because I was using a shared scratch object as the hashmap key which mutated the hashmap's key, breaking the hashmap in the process.

PR #3033, opened 2025-02-18
Fix Holes in World by Retaining Index Data During Partial Updates

douira:
Make sure the index data isn't removed when vertex data is removing for a vertex data upload but index data is kept. This is a slightly different bug from the last similar one because this concerns erroneous overwriting of index data information when removing vertex data instead of when writing vertex data info. 
Also refactored the data removal methods into one because they have significant overlap, and they can also share the same clear call if both types of data are being removed at the same time.


douira:
Marked as ready again, there don't seem to be any known issues with this code atm. I'm investigating a different issue but will make another PR when I figure out what it is.

PR #3034, opened 2025-02-19
Fix Missed Sort From Combined Rebuild Task

douira:
Fix a bug where sections that have a pending sort task don't get sorted when the sort type gets updated to rebuild but then the rebuild task may not do any sorting because it can reuse already uploaded index data.

This changes chunk update types to add an option for rebuilding with forced sorting, which is used as the combined task type when sorting and rebuilding tasks are combined. A caveat is that this expands the ChunkUpdateType by two members, which means there are also more task lists, but it's largely unproblematic as this combined update type will only get used occasionally.

The way ChunkUpdateType calculates joined types is also more straightforward now.


douira:
Note that this is also implemented in a more reasonable way by both [async culling merge-item-4](https://github.com/douira/sodium/commit/005e7e02cbbf8a515f1504110967a10c1d31b2ee) and [merge-item-1](https://github.com/CaffeineMC/sodium/compare/dev...douira:sodium:merge-item-1-task-effort-estimation#diff-e04084e8dbb4901c055859088a2f3de91c79f17a6ed355e7b0514ec45d0690fd)

PR #3051, opened 2025-02-27
Fix Translucency Sort Triggering at Very High Camera Coordinates

douira:
Fix translucency sort triggering at very high camera coordinates (like around 16 million) by using full double precision in the calculation of the NormalPlanes's dot product offset.


douira:
I haven't observed this being a problem in trunk and it was only discovered as part of testing on the [quad splitting branch](https://github.com/CaffeineMC/sodium/pull/2993).

PR #3063, opened 2025-03-11
Add endertype_clouds shaders to the shader program blacklist

### LINKED ISSUE #3062: Vanilla clouds shader should be in the shader program blacklist
### Request Description

Overriding the vanilla `rendertype_clouds` shader does nothing on sodium due to the custom cloud renderer, but it isn't listed in the shader program blacklist and thus doesn't emit a warning.

katietheqt:
Fixes #3062


jellysquid3:
Thanks for the patch.

PR #3071, opened 2025-03-17
Don't crash Minecraft if detected on Pojavlauncher 

SwiftCris:
Change the code so, instead of throwing a new error which, crashes the game, only show a warning. Which, is better since, sodium is not supported but, some Pojavlauncher users are asking why their game is crashing not knowing it's from sodium.

### DIFF
--- a/common/src/workarounds/java/net/caffeinemc/mods/sodium/client/compatibility/checks/PostLaunchChecks.java
+++ b/common/src/workarounds/java/net/caffeinemc/mods/sodium/client/compatibility/checks/PostLaunchChecks.java
@@ -21,7 +21,8 @@ public class PostLaunchChecks {
 
         // FIXME: This can be determined earlier, but we can't access the GUI classes in pre-launch
         if (isUsingPojavLauncher()) {
-            throw new RuntimeException("It appears that you are using PojavLauncher, which is not supported when " +
+            // Only warn the user that theyre using pojav, trowing a new error will crash their game. 
+            LOGGER.warn("It appears that you are using PojavLauncher, which is not supported when " +
                     "using Sodium. Please check your mods list.");
         }
     }


IMS212:
This change was made intentionally.

douira:
Why do you think a PR that simply reverts the [1-month old commit that converted the warning into a fatal error](https://github.com/CaffeineMC/sodium/commit/53bb2976effb3d6d3e785da1ddafc8f16be25721) would be accepted?

SwiftCris:
> Why do you think a PR that simply reverts the [1-month old commit that converted the warning into a fatal error](https://github.com/CaffeineMC/sodium/commit/53bb2976effb3d6d3e785da1ddafc8f16be25721) would be accepted?


Why do you have to do this though? It feels very unnecessary.

SwiftCris:
> This change was made intentionally.

Oh, so instead of a logger warning, just crash the whole game. Intresting

douira:
Refer to the explanation [here](https://github.com/CaffeineMC/sodium/issues/3030#issuecomment-2663885235).

SwiftCris:
> Refer to the explanation [here](https://github.com/CaffeineMC/sodium/issues/3030#issuecomment-2663885235).

Ok, so instead of, having people come to support and say "omg my game crashed", you...... Do the exact same thing by crashing the game? If these people are so clueless, then why do you think they won't try to get support if you do this?

douira:
This is not the venue for extended discussion of this topic.

SwiftCris:
Also 

> > Refer to the explanation [here](https://github.com/CaffeineMC/sodium/issues/3030#issuecomment-2663885235).
> 
> Ok, so instead of, having people come to support and say "omg my game crashed", you...... Do the exact same thing by crashing the game? If these people are so clueless, then why do you think they won't try to get support if you do this?

And, to add to this, instead of people coming to sodium asking for support. They will come to pojav asking the exact same thing, which basically is the same exact issue only that sodium doesn't do anything and it's in the hands of pojavlauncher support to tell people that there is code that kills any instance running pojavlauncher.

jellysquid3:
We are not going to hear your case. The team's opinion is very firm on this -- The crash logic will stay as it is. Besides, we've been shipping the code for weeks now and haven't had any "confused users".

jellysquid3:
And, side note: It's going to be **very** difficult to convince us to show any leniency in this matter after the PojavLauncher team banned two moderators and their community manager for sexually inappropriate behavior with minors.

PR #3084, opened 2025-03-29
fix: don't show fabulous on unsupported systems

contariaa:
Vanilla hides "Fabulous!" on unsupported GPU's, this simply copies the check used by vanilla.
Works correctly even when the value is set to fabulous through options.txt, the sodium menu will simply display "Fabulous!", then when pressed skip to fancy -> fast -> fancy -> fast -> ...

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/CyclingControl.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/CyclingControl.java
@@ -50,6 +50,12 @@ public class CyclingControl<T extends Enum<T>> implements Control<T> {
         }
     }
 
+    public CyclingControl(Option<T> option, T[] allowedValues, Component[] names) {
+        this.option = option;
+        this.allowedValues = allowedValues;
+        this.names = names;
+    }
+
     @Override
     public Option<T> getOption() {
         return this.option;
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/CyclingControl.java
```diff
@@ -50,6 +50,12 @@ public CyclingControl(Option<T> option, Class<T> enumType, T[] allowedValues) {
         }
     }
 
+    public CyclingControl(Option<T> option, T[] allowedValues, Component[] names) {
```
jellysquid3, line 61: There should be validation that `allowedValues` and `names` are the same length, and that `allowedValues` is an array of the correct type.

jellysquid3 (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/CyclingControl.java
contariaa: In the current implementation  `allowedValues` and `names` are not guaranteed to be of the same length, the only requirement is that `allowedValues.length <= names.length`. Specifically `allowedValues` is an array of length 3 (default) or 2 (on systems that do not support fabulous graphics), while `names` is always an array of length 3 containing all the translations.
Would you like me to validate that `allowedValues.length <= names.length` or change the passed `names` to always match the length?

contariaa: Also, validation seems to only be present in one out of the two already existing constructors, would you like me to add that in this PR?

contariaa: Taking another look at it, the existing constructor `public CyclingControl(Option<T> option, Class<T> enumType, T[] allowedValues)` also initializes `names` with the length of the `universe` (`enumType.getEnumConstants()`) rather than the length of `allowedValues`.
In `CyclingControl.CyclingControlElement#render`, the `names` array is queried not with the index of the value in `allowedValues`, but rather the values enum ordinal, so it seems appropriate to validate `names` to always match the length of the enums universe, not just its allowed values.

contariaa: I've pushed validation that makes sense to me, `names` should always be the same length as the entire enum universe because its accessed with the enum ordinal, not the index of the value in the `allowedValues` array

PR #3087, opened 2025-03-30
fix: immediately upload sprites when they become visible again

### LINKED ISSUE #2514: Animated GUI textures with repeated frames can desync/freeze when Only Animate Visible Textures is enabled
### Bug Description

When Only Animate Visible Textures is enabled, animated GUI sprites can freeze on frames they shouldn't depending on when you close the GUI.  I believe this has to do with using a frame list in the .png.mcmeta, and opening/closing the GUI at specific times based on when repeated frames are playing.

### Reproduction Steps

1. Have an animated GUI sprite with repeat frames listed in the .png.mcmeta.  In my case, the animation holds on frame 0 for a while before playing the rest on each loop.  
2. Open the GUI and wait until the animation has progressed past the repeated frames.
3. Close the GUI when the animation is on a frame that is not repeated
4. Wait a bit until you think the repeated frames would be playing if the GUI were open
5. Open the GUI and observe the texture.  It will be frozen on the frame you closed the GUI on, and fix itself when the animation loops again.

Here's what it should look like:

https://github.com/CaffeineMC/sodium-fabric/assets/69820300/24ee5047-56dc-45c8-a766-3bd9cbf303ad

Here's what actually happens:

https://github.com/CaffeineMC/sodium-fabric/assets/69820300/d70efe7d-8987-4e37-8925-9643ce4882ff

A few other notes - 
This may not be limited to just GUI sprites, but that is all I had time to test.  
A way to have specific textures ignore Only Animate Visible Textures would be nice, if feasible.  Perhaps an additional optional field in the .png.mcmeta

### Log File

[latest.log](https://github.com/CaffeineMC/sodium-fabric/files/15387623/latest.log)

### Crash Report

[crash-2024-05-21_05.33.54-client.txt](https://github.com/CaffeineMC/sodium-fabric/files/15387830/crash-2024-05-21_05.33.54-client.txt)

contariaa:
When a sprite becomes active again after skipping uploads, it won't immediately update to its correct state, instead it will only do so the next time the frame index changes.
I initially fixed this by just ensuring it will always upload the first tick after becoming active again, however that still leaves the time between being marked as active and being ticked where the texture would be wrong.
The fix for that is to ensure the upload right when the sprite is marked as active, which means we have to look up and bind the atlas texture everytime that happens. I'm unsure about the performance cost of that, if it's a problem I can revert the second commit, the first one alone is still significantly better than the current behaviour.

Closes https://github.com/CaffeineMC/sodium/issues/2514

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/animations/tracking/SpriteContentsTickerMixin.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/animations/tracking/SpriteContentsTickerMixin.java
@@ -3,6 +3,7 @@ package net.caffeinemc.mods.sodium.mixin.features.textures.animations.tracking;
 import net.caffeinemc.mods.sodium.client.SodiumClientMod;
 import net.caffeinemc.mods.sodium.client.render.texture.SpriteContentsExtension;
 import net.minecraft.client.renderer.texture.SpriteContents;
+import org.jetbrains.annotations.Nullable;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Shadow;
@@ -16,15 +17,24 @@ import java.util.List;
 @Mixin(SpriteContents.Ticker.class)
 public class SpriteContentsTickerMixin {
     @Shadow
+    int frame;
+    @Shadow
     int subFrame;
     @Shadow
     @Final
     SpriteContents.AnimatedTexture animationInfo;
     @Shadow
-    int frame;
+    @Final
+    @Nullable
+    private SpriteContents.InterpolationData interpolationData;
 
     @Unique
     private SpriteContents parent;
+
+    @Unique
+    private boolean skippedUpload;
+    @Unique
+    private int uploadedFrame;
 
     /**
      * @author IMS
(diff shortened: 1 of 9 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/textures/animations/tracking/SpriteContentsTickerMixin.java
```diff
@@ -57,4 +74,28 @@ private void postTick(CallbackInfo ci) {
...
         parent.sodium$setActive(false);
     }
+
+    @Override
+    public void sodium$ensureUpload(ResourceLocation atlas, int x, int y) {
+        if (!this.skippedUpload) {
+            return;
+        }
+
+        Minecraft.getInstance().getTextureManager().getTexture(atlas).bind();
```
contariaa, line 84: Revisiting this, its probably problematic to change the bound texture here since we can't know at what point in the code a sprite is being marked active and it might mess up render state.

contariaa: This issue is not present with just the first commit, but as stated in the initial pr message the behaviour from just the first commit is not 100% correct since it leaves a delay between the texture becoming visible and when its reticked

douira: is there a safe approach to this? Does anyone know whether this is safe?

douira:
Do you have a safe and correct approach to this? it seems you expressed some doubt about the method last year. This PR has also accumulated merge conflicts. It should be profiled to see if there's a performance cost to the second change you made (and there were concerns for safety).

contariaa:
I don't know of any way to achieve instant sprite updates safely, the problem is that the hooks for marking a sprite as active may be called at any time before they are rendered, which makes the sprite upload fundamentally unsafe (it has to bind the sprite texture to upload, thus changing render state).
I do think porting and merging just the first commit is still a valueable improvement, the current code causes the image to only be re-uploaded when the frame changes, even if the current one isn't uploaded yet, example:
- a sprite with 4 frames at 20 ticks (also called sub-frames in the code) each
- frame 1 gets uploaded at 0 ticks
- sprite becomes invisible at 10 ticks
- frame 2 does not get uploaded at 20 ticks because the sprite is currently invisible
- sprite becomes visible again at 30 ticks, still on frame 1
- frame 3 gets uploaded at 40 ticks, after 10 ticks of the wrong frame being visible

If there is interest I can look into porting the first commit, while it still shows the old frame until the sprite is next ticked, it's still better than having the old frame visible until a new frame is uploaded in my opinion.

douira:
Yeah I think the first commit would still be an improvement as you said. Would be interesting see what it looks like in practice with the reproducer shown in the referenced issue.

PR #3094, opened 2025-04-01
Fix performance regressions with buffer uploads in 1.21.5

AzureAaron:
This PR essentially reverts Mojang's changes to buffer uploading in 1.21.5 where the new method of uploading buffer data via `glBufferSubData` is inefficient and causes significant performance losses across some systems (see MC-295893 for further details).

This patch on my system (macOS) pretty much fixes the significant performance loss I've experienced from this update. I'm not sure if the patch should only be applied to certain systems or not given the many users reporting performance losses (which may or may not be related to this) so that might be something worth considering.

I believe I have added the mixins and whatnot correctly, if anything needs changing please let me know. I also had to update MixinExtras for the `@Cancellable` annotation (Fabric and Neo ship this version so it should be fine).

### DIFF
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -50,8 +50,8 @@ dependencies {
         }
     })
 
-    compileOnly("io.github.llamalad7:mixinextras-common:0.3.5")
-    annotationProcessor("io.github.llamalad7:mixinextras-common:0.3.5")
+    compileOnly("io.github.llamalad7:mixinextras-common:0.4.1")
+    annotationProcessor("io.github.llamalad7:mixinextras-common:0.4.1")
 
     compileOnly("net.fabricmc:sponge-mixin:0.13.2+mixin.0.8.5")
     compileOnly("net.fabricmc:fabric-loader:${BuildConfig.FABRIC_LOADER_VERSION}")
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/data/config/MixinConfig.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/data/config/MixinConfig.java
@@ -63,6 +63,7 @@ public class MixinConfig {
 
         this.addMixinRule("features.render.immediate", true);
         this.addMixinRule("features.render.immediate.buffer_builder", true);
+        this.addMixinRule("features.render.immediate.buffer_upload", true);
         this.addMixinRule("features.render.immediate.matrix_stack", true);
 
         this.addMixinRule("features.render.model", true);
--- a/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_upload/GlBufferAccessor.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_upload/GlBufferAccessor.java
@@ -1,0 +1,12 @@
+package net.caffeinemc.mods.sodium.mixin.features.render.immediate.buffer_upload;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+import com.mojang.blaze3d.opengl.GlBuffer;
+
+@Mixin(GlBuffer.class)
+public interface GlBufferAccessor {
+
+    @Accessor
+    void setInitialized(boolean value);
+}
--- a/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_upload/VertexFormatMixin.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_upload/VertexFormatMixin.java
@@ -1,0 +1,34 @@
+package net.caffeinemc.mods.sodium.mixin.features.render.immediate.buffer_upload;
+
+import java.nio.ByteBuffer;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import com.llamalad7.mixinextras.injector.ModifyReceiver;
+import com.llamalad7.mixinextras.sugar.Cancellable;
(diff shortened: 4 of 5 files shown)


djmrFunnyMan:
I'm seeing up to a 8% performance drop in 1.21.5 on my Ryzen 7600X + Intel B580 system. This pr does not remedy that at all.

However interestingly enough it brings GPU utilization way down to what it was in 1.21.4 (80-50% in 0.6.12, 20-40% with this patch)

BlueGradientHorizon:
On my laptop, r5 4600h + gtx1650m it seems that this PR sadly does not affect FPS in any way. For me, GPU utilization always has been at 99-100% w/shaders and 30-50% w/o on 1.21.4 and is the same on 1.21.5, but with lower FPS. I'm experiencing a ~30 FPS drop in general compared to 1.21.4, both with Sodium only and Sodium+Iris. If you need any additional info, let us know.

djmrFunnyMan:
With Iris shaders on this PR introduced insane frametime instability



It also does nothing to fix the regressions (20% lower FPS since 1.21.4)

BlueGradientHorizon:
> insane frametime instability

For me the frametime graph itself is perfectly fine, and I also use Complementary shader. Maybe driver-specific problem arised?

djmrFunnyMan:
Here's some spark profiles:

Release Sodium: https://spark.lucko.me/88dmo9pZvH
This Patch: https://spark.lucko.me/MvAPNsiJZj

douira:
Using Sodium's `1.21.4/stable` and `1.21.5/stable` branches I wasn't able to find a performance regression on my macOS system. (intel 10900K, RX 6900 XT) It should be noted that my hardware configuration is very uncommon.

douira:
Testing Sodium's `1.21.4/stable` and `1.21.5/stable` branches and this PR I wasn't able to find a change in any direction beyond measurement noise on my macOS system. (intel 10900K, RX 6900 XT) It should be noted that my hardware configuration is very uncommon. (deleted my previous comment due to measurement error)

mrjasonn:
> Testing Sodium's `1.21.4/stable` and `1.21.5/stable` branches and this PR I wasn't able to find a change in any direction beyond measurement noise on my macOS system. (intel 10900K, RX 6900 XT) It should be noted that my hardware configuration is very uncommon. (deleted my previous comment due to measurement error)

@douira Maybe try testing with lots of entitys, for example, Hypixel bedwars lobby 1

douira:
I've tested this with 1662 item frames and I can in fact now measure a fps regression between the `1.21.4/stable` and `1.21.5/stable` branches (75 fps down to 57 fps). However, applying this patch makes it worse, it goes down to 51 fps.

`1.21.4/stable`:
1 21 4 stable

`1.21.5/stable`:
1 21 5 stable

this patch:
1 21 5 3094 patch

mrjasonn:
> I've tested this with 1662 item frames and I can in fact now measure a fps regression between the `1.21.4/stable` and `1.21.5/stable` branches (75 fps down to 57 fps). However, applying this patch makes it worse, it goes down to 51 fps.
> 
> `1.21.4/stable`: 1 21 4 stable
> 
> `1.21.5/stable`: 1 21 5 stable
> 
> this patch: 1 21 5 3094 patch

@douira Oh! I have noticed that even on a Mac, you're not using an Apple designed GPU. You will notice the performance difference with the many entities on an Apple designed GPU, as fast buffer upload actually makes Apple GPUs lag.

mrjasonn:
@douira go see RaphiMC/ImmediatelyFast#279, it may give some useful information on this, as ImmediatelyFast also had their fast buffer upload optimisation enabled for Apple GPUs at some point before Mojang added it themselves.

Kalanyr:
I definitely have the performance regression issue ( about 75% FPS for 1.21.5 vs 1.21.4 ) and while this has an effect it's pretty minor ( it's roughly x1.1 FPS on 1.21.5 as compared to without it ) so I don't think this is the actual major cause of regression at least for my setup.

djmrFunnyMan:
@Kalanyr what is your setup?

Kalanyr:
Intel x86_64, Windows 11, Laptop NVIDIA 3080 ( it's a Performance version
since it's a Desktop Replacement Laptop but still a Laptop Mode)

RaphiMC:
Hello, ImmediatelyFast dev here.
The buffer upload regression in 1.21.5 does not affect all systems. It mostly affects Apple GPUs. Avoiding glBufferSubData on all systems causes a performance decrease on most GPU drivers. ImmediatelyFast <=1.21.4 in fact even has an optimization which replaces glBufferData with glBufferSubData because its measureable faster on most systems. I although had a check, which checked for an Apple GPU and then don't apply that optimization because Apple's OpenGL driver is horribly optimized. I have now added a fix for the buffer upload regression in ImmediatelyFast 1.21.5: https://github.com/RaphiMC/ImmediatelyFast/commit/f6de9a693d551129a9b0ed7bd90020be8be1278a (Release 1.9.4)

It should be noted that 1.21.5 has multiple other performance regressions, even ones which affect all GPU vendors, so direct comparisons between 1.21.4 and 1.21.5 even with the buffer upload fix are not possible.

BlueGradientHorizon:
> 1.21.5 has multiple other performance regressions

I just don't understand how mojang could release such a completely broken update... Almost the same set of mods of the latest versions (at the moment), the same set of resource packs, the same shader with the same settings, and the difference is -50% frames.




But what's more frightening is that some mod developers are already stopping support for 1.21.4 or are moving it to the category of receiving only critical updates.

douira:
I have a apple silicon user who wants to test this but the branch doesn't even build because it cannot download the FRAPI build from @IMS212's maven. Could this maybe be updated to apply on `dev` or can IMS fix the build?

redactedontop:
This should be marked as a draft.

djmrFunnyMan:
I'd like to note that the big regression with shaders was unrelated to this after all https://github.com/IrisShaders/Iris/issues/2940

redactedontop:
Yeah, this is just for Apple GPUs and their regression caused by flushing when using subdata uploads.

PR #3099, opened 2025-04-06
Faster section aabb test

MoePus:
This PR precomputes section size into SimpleFrustum and skips far plane check to make it faster during RenderSectionManager.createTerrainRenderList
In Sparks time per tick mode createTerrainRenderList's time decreases from 10.75ms to 7.81ms and fps improves from 400 to about 500 when continuously moving mouse around with RD 40.
Reflection is used in this pr to access planes in FrustumIntersection


douira:
I've measured the actual time it takes to perform the occlusion culling on an average test scene with and without this patch after plenty of warmup, averaged over 5000 samples, and over multiple runs.
`dev`: 1164µs
this patch: 1015µs
This comes out to an occlusion culling time reduction of around 12.8%. As a nitpick we might want to adjust some of the method's names but it looks good to me in principle and practice.

douira:
I'm interested in finally merging this, but there's a minor merge conflict. Also, would it be possible to use a (mixin) Accessor instead of reflection?

PR #3117, opened 2025-04-20
Implement linear forwards&backwards VisGraph

Moulberry:
Uses two linear propagation passes instead of BFS
Reduces visible section count by 2-3% (after fix)


Moulberry:
Unfortunately there are some edge cases which I don't think are solvable

PR #3132, opened 2025-04-30
fix shader ressource loading

coco875:
this modification fix error with nvidium when loaded through connector

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/ShaderLoader.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/ShaderLoader.java
@@ -25,7 +25,7 @@ public class ShaderLoader {
     public static String getShaderSource(ResourceLocation name) {
         String path = String.format("/assets/%s/shaders/%s", name.getNamespace(), name.getPath());
 
-        try (InputStream in = ShaderLoader.class.getResourceAsStream(path)) {
+        try (InputStream in = ShaderLoader.class.getClassLoader().getResourceAsStream(path)) {
             if (in == null) {
                 throw new RuntimeException("Shader not found: " + path);
             }


coco875:
for more context, this is the message of Su5eD https://ptb.discord.com/channels/1141048834177388746/1141069780225380523/1177749063609892924 it's was for embeddium in 1.20.1

jellysquid3:
I don't have access to the mentioned Discord channel, can you please explain why this is necessary?

coco875:
> I've investigated Nvidium compatibility and to my surprise, it seems to actually work? That is, except for one small problem which requires a change in embeddium to be fixed.
Embeddium uses the class object for locating resources, which behaves differently in a modular environment. Using this method, SJH's classloader will only search for resources in the caller's module, making it impossible to load shaders from other mods, including nvidium. (not sure if any of this *should* be used by addons though).
The fix is quite simple - moving `getResourceAsStream` to be called on the original class's classloader instead, which searches all loaded mods for the specified resource.
```diff
@@ -25,7 +25,7 @@
     public static String getShaderSource(Identifier name) {
         String path = String.format("/assets/%s/shaders/%s", name.getNamespace(), name.getPath());
 
-        try (InputStream in = ShaderLoader.class.getResourceAsStream(path)) {
+        try (InputStream in = ShaderLoader.class.getClassLoader().getResourceAsStream(path)) {
             if (in == null) {
                 throw new RuntimeException("Shader not found: " + path);
             }
```

the original message, I hope it help

douira:
There's already a compatible nvidium beta. Why is this necessary?

coco875:
because it crash because of `java.lang.RuntimeException: Shader not found: /assets/nvidium/shaders/terrain/task.glsl` (the full log: https://mclo.gs/kVbGl1P) (I use the fork https://github.com/drouarb/nvidium) and I use neoforge, not fabric

drouarb:
Would be interested into getting that merged to have cross compilation inside nvidium. On neoforge ressources are isolated per mods, this would allow to load shader with directives such as `#import <nvidium:occlusion/scene.glsl>` otherwise sodium fails to load anything outside of sodium namespace.

jellysquid3:
I've gone ahead and merged this, reluctantly. We really don't support mods swapping out our shader sources since it is extremely fragile, and the code was originally written to *prevent* this (assuming classloader isolation ever existed, which it didn't.)

mreduar:
When will these changes be published?

douira:
This was merged into 1.21.1/stable but never made it into a release because by that point we were past 1.21.1. Is this change still relevant on current versions of Minecraft (1.21.11) and sodium (0.8)?

PR #3135, opened 2025-05-02
Change Primitive Iteration in TS System Away From Enhanced For Loops

douira:
Change enhanced for loops over primitive lists in the translucency sorting system to use regular for loops or primitive iterators instead, potentially improving performance by avoiding (un)boxing (I have not measured).

Based on @MCRcortex's [suggestion](https://discord.com/channels/602796788608401408/651120262129123330/1367856874577330316).


jellysquid3:
This looks like a good change. It is valuable to avoid primitive boxing even if the JIT may already be eliding the allocations.

PR #3142, opened 2025-05-09
Upgraded IsPojavLauncher() to include link with id and reason

RimakiTaema:
So i implement upgraded IsPojavLauncher class to be more longer and more detail.

This is my first pull request, not good at java.

(Not ready for review yet as no tests code)


jellysquid3:
Your changes don't make sense and we have no intention of downgrading the crash to a passive warning.

RimakiTaema:
Thanks for information.
I get it downgrading to logging isn't ideal, But instead i want to implement that warn user then crash it.
I usually run and see no log message about it in pojav. 
Just only see exception message in crash.

RimakiTaema:
(2)
Instead i want to be warn user with message like
"Pojav isn't supported, Closing Game"
In typical log, Since some pojav player would check live logs

PR #3150, opened 2025-05-13
Update fabric.mod.json

TimofeyFilkin:
tlskincape break


Samalando:
I'm guessing this won't get merged because of https://github.com/CaffeineMC/sodium/issues/3030#issuecomment-2663071835

jellysquid3:
Why? We're not the police.

TimofeyFilkin:
> Why? We're not the police.

because TLauncher just reuploads mods. Without any credits, links to authors, etc. And mod reuploads are bad.

douira:
This is not the venue for this discussion

PR #3249, opened 2025-09-03
this shows triangle strips can still reduce resource usage

Mr-Squid42:
seed: -4745202004167156036
settings: default graphics mostly except clouds turned off + vsync disabled + uncapped framerate
gamerules: respawnspawnradius set to 0 + all mob spawning turned off
java version 21.0.7
jdk flags: -Xms4096M -Xmx4096M

with triangle lists 12 chunks:
12chunkswithtrianglelists
with triangle lists 12 chunks:
12chunkswithtrianglestrips

geometry buffer ~ 1.7 % reduction in max capacity size needed in MB
index buffer ~ 18.2 % reduction in max capacity size needed in MB

with triangle lists 32 chunks:
32chunkswithtrianglelists
with triangle strips 32 chunks:
32chunkswithtrianglestrips

geometry buffer no reduction in max capacity size needed in MB
index buffer ~ 17.4 % reduction in max capacity size needed in MB

from what i can tell the frame rate seems to be about the same

full disclosure, i'm not that good at programming, i did need to use a coding agent to fix quite a few errors, i'm still not sure how the translucency is rendering as far as i can tell it is currently in my version using incorrect winding order, but i don't know what to do to fix this and it seems to be rendering fine


MCRcortex:
In your second example at 32 rd you set your fps limit to 30

douira:
Interesting work, though I would like to note some caveats that call into question the practical use of this patch, aside from the engineering concerns. I've performed some more useful measurements on a test world that contains a very large amount of geometry.

patch:
strip 2
dev:
dev 2

The patch has no effect on fps in any scenario I tested. In this scene we can see a 16.6% decrease in index buffer memory usage (not arena size or what you call "max capacity size", that's a bad metric to consider). In normal scenes like this one, the delta is negligible:

patch:
strip

dev:
dev

You can see here the index buffer memory usage only changed from 5 to 4MB, which is percentage-wise a similar amount but irrelevant in absolute terms. The arena size here is not a good thing to measure because you can see in the other test that the amount of unused but allocated memory goes down to almost zero once you're using more memory overall.

I've also previously prototyped involved approaches to reducing index buffer memory usage, but we ultimately decided their complexity is not worth the minimal benefit given how contrived the scenes need to be for them to be useful. A much more relevant area would be reducing vertex buffer usage, but that's a hard problem to solve.

As for the engineering concerns, it's pretty obvious that chatgpt or whatever agent you used did some stupid things in the code, and there's also a bunch of unnecessary changes thrown in around the place. I couldn't see any broken translucency sorting or vertex corruption, but with these kinds of changes to _the most fragile piece of code we have in the repo_ we'd need to think about it really long and do a lot of thorough testing. Whether it's even worth addressing those concerns is another matter since I'm not sure this patch has practical justification to be merged.

jellysquid3:
> geometry buffer ~ 1.7 % reduction in max capacity size needed in MB
> index buffer ~ 18.2 % reduction in max capacity size needed in MB

This is inconsequential. The _actual memory usage_ is practically unchanged (~1 MB) even at high render distance, because most of the index sequences for opaque geometry are shared. If we cared about the size of the index pool, we'd implement a compaction strategy rather than try to reduce the size of the index data.

_Primitive Restart_ massively reduces performance on older GPUs, most notably on older AMD architectures. Since the memory usage here is not changed in any meaningful manner, there is no reason to use it.

PR #3254, opened 2025-09-12
Swap face indexes to achieve parity with vanilla

2190303755:
Current face rendering behavior of sodium is different from vanilla due to swapped indexes. There is a demo mod that uses manual face culling at [2190303755/Cube](https://github.com/2190303755/Cube) for testing purpose.

Vanilla:
Vanilla

Sodium:
Sodium

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ModelCuboid.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ModelCuboid.java
@@ -27,9 +27,9 @@ public class ModelCuboid {
     public static final int
             FACE_NEG_Y = 0, // DOWN
             FACE_POS_Y = 1, // UP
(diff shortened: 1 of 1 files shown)


douira:
The indexing in question was [introduced here](https://github.com/CaffeineMC/sodium/commit/adad380b98cd768e84572676ce6ae1a589f2657f) and this change does indeed change the indexes correspond to Vanilla. Does this break core shaders that operate on entity vertices? I remember at some point we had issues with that.

2190303755:
I am sorry that I am not very familiar with shaders. I think this change only affects the visibility of faces because the only usage of `getFaceIndex` is in `createCullMask`.

jellysquid3:
I thought we had fixed this in a follow up patch but I guess it's still broken. Thanks for the patch.

jellysquid3 (approved):

PR #3255, opened 2025-09-12
Improved Immediate Chunk Presentation

### LINKED ISSUE #2768: Pistons and blocks moved by them flicker while the piston retracts or extends (Beta 0.6.0)
### Bug Description

Starting in Sodium 0.6.0 beta.1, pistons/sticky pistons will briefly flicker while extending or retracting. This also applies to blocks pushed/pulled by the pistons. While extending, the head/arm of the piston and block it is pushing will not render for one frame. While retracting, the whole piston and the block it is pulling will not render for one frame.

Example video (taken at 60fps):

https://github.com/user-attachments/assets/bf59e741-70a5-46df-8b0d-3496d9635489

### Reproduction Steps

- Use Sodium 0.6.0 beta.1 or beta.2 on Fabric 1.21 or 1.21.1. (I have not tested NeoForge.)
- Enter a world and place a sticky piston. (Placing it in any direction should work, but I mostly tested it facing upwards.)
- Place a block on top of the sticky piston.
- Use a button or lever to power the piston. To see the bug in full effect, the piston should be powered on and off slowly, as spamming it seems to lessen the flickering.
- The flickering should occur. Lowering your game framerate may make it more obvious.
- Repeat the above steps, but use Sodium 0.5.11 as to compare it to Sodium 0.6.0. The piston flickering does not seem to occur in 0.5.11.

### Log File

Uploaded in case you need extra information
https://mclo.gs/PwZcEDt
[latest.log](https://github.com/user-attachments/files/17085478/latest.log)

### Crash Report

Uploaded in case you need extra information
https://pastebin.com/a3ALha73

douira:
### [Remove redundant code for implementation of the sort task defer mode.](https://github.com/CaffeineMC/sodium/commit/19d591e2883996270976ae7fe829d6e52258c2d1)

There was previously an implementation for controlling sort task deferring as part of the sorting code. A different implementation of essentially the same concept but for rebuild tasks was added later. Now they both use the newer approach, which simplifies the task submit code.

### [Improve immediate presentation of meshes generated in blocking mode.](https://github.com/CaffeineMC/sodium/commit/8b10c2dde539135de8178ad92a1992e8ef1d1cbf)

Previously even when meshes were generated in the blocking defer mode "Immediate" (zero frame) they would not become visible if they were part of a previously invisible section and thus not in a render list already. This change makes it so that sections that receive results from blocking rebuilds are added to the render list.

When the "Immediate" chunk update defer mode is selected this fixes the issue of holes in the world when breaking a block that reveals a previously invisible section without needing to perform a second graph search pass. This is especially an issue on slower hardware and lower fps limits.

Also fixes #2768 by removing the delay between uploading chunk meshes and the render lists getting updated to include the presence of block entities in the chunk section.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/SodiumWorldRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/SodiumWorldRenderer.java
@@ -247,6 +247,10 @@ public class SodiumWorldRenderer {
             }
         }
 
+        profiler.popPush("chunk_render_lists");
+        
+        this.renderSectionManager.finalizeRenderLists(viewport);
+
         profiler.popPush("chunk_render_tick");
 
         this.renderSectionManager.tickVisibleRenders();
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/ChunkUpdateTypes.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/ChunkUpdateTypes.java
@@ -37,7 +37,7 @@ public class ChunkUpdateTypes {
         return (isRebuild(type) || isInitialBuild(type)) && isSort(type);
     }
 
-    public static TaskQueueType getQueueType(int type, TaskQueueType importantRebuildQueueType) {
+    public static TaskQueueType getQueueType(int type, TaskQueueType importantRebuildQueueType, TaskQueueType importantSortQueueType) {
         if (isInitialBuild(type)) {
             return TaskQueueType.INITIAL_BUILD;
         }
@@ -45,7 +45,7 @@ public class ChunkUpdateTypes {
             if (isRebuild(type)) {
                 return importantRebuildQueueType;
             } else { // implies important sort task
-                return TaskQueueType.ZERO_FRAME_DEFER;
+                return importantSortQueueType;
             }
         } else {
             return TaskQueueType.ALWAYS_DEFER;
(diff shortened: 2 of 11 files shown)


douira:
This is currently broken

douira:
It appears I've fixed it. Briefly tested by Adre and me. It needed more code than I wanted but ok, it seems to work.

douira:
I received some testing from a user in our discord server:
> @douira apologies for the ping out of nowhere, but I wanted to thank you for your work on the flickering piston bug. It's already a big improvement from prior. However, I also wanted to alert you to a number of edge cases where the old behavior (or worse) still occurs, hopefully while the technical details are still somewhat fresh to you.
> The old flickering behavior still occurs often when blocks are moved across subchunk boundaries, or when the blocks are in a different subchunk and the camera is sufficiently far (about 11) blocks away. The new related behavior I noticed is that when a piston pushes a non-full block upwards through a subchunk border and there are no higher opaque blocks in that column, the block entity just does not render at all. The "Chunk Updates" option doesn't seem to help with any of these, unfortunately.

> Just loaded up one of my 1.20 instances, and no distance flicker, even despite having updates deferred. With updates deferred I do get subchunk boundary flicker, but only that.
> So Sodium 0.5.11 has none of these bugs without deferred updates, and only one with.

> Vanilla has no flickering at all, up to full BE render distance, and not on subchunk borders either.
> (I just double checked everything in vanilla 1.21.8)

(quoted from applestruck on discord)

More context and further conversation [here](https://discord.com/channels/602796788608401408/752637878462578778/1423223580849864745).

It seems this PR fixes some but not all problems with flickering with pistons. I don't know why 0.5.x doesn't have these issues but maybe it does more things synchronously, or the ordering changes stuff? I susepct this might be caused by switching from uploading the data from one frame at the start of the next frame instead of trying to do it all in one frame. That would effectively just delay all frames by one, thereby avoiding flickering (at the cost of latency)

douira:
I can reproduce the flickering bug where a piston pushes in/out of an otherwise empty chunk on any version of sodium (0.5, 0.6, 0.7). See the debugging [here](https://discord.com/channels/602796788608401408/651120262129123330/1424826582123417621). The summary of my analysis is that the task for presenting the now empty chunk completes much faster than the meshing task that has to actually do work. Actually modeling the dependency between these two meshes is impossible on a technical level in my opinion. Fixing this in practice requires blocking on updates, which is infeasible in the distance and we generally don't want for non-player caused changes anyway.

Vanilla does not have flickering issues until you move far away (I assume the distance at which BEs no longer render?). Does it just block on these updates?

jellysquid3 (approved):

PR #3277, opened 2025-10-03
Replace estimator floats with doubles

MCRcortex:
Given the size of the numbers that are being handled (nanoseconds and megabytes, both of which are in the millions) it makes sense to use doubles instead of floats for computation as floating point precision may become an issue otherwise

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
@@ -48,7 +48,7 @@ public abstract class Average1DEstimator<C> extends Estimator<C, Average1DEstima
         private boolean hasRealData = false;
         private double average;
 
-        public Average(double newDataRatio, float initialValue) {
+        public Average(double newDataRatio, double initialValue) {
             this.average = initialValue;
             this.newDataRatio = newDataRatio;
         }
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
@@ -5,10 +5,10 @@ import net.caffeinemc.mods.sodium.client.util.MathUtil;
 import java.util.Locale;
 
 public abstract class Average1DEstimator<C> extends Estimator<C, Average1DEstimator.Value<C>, Average1DEstimator.ValueBatch<C>, Void, Long, Average1DEstimator.Average<C>> {
-    private final float newDataRatio;
+    private final double newDataRatio;
     private final long initialEstimate;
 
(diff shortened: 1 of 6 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/ExpDecayLinear2DEstimator.java
```diff
@@ -90,20 +90,20 @@ public void update(ClearingLinearRegressionBatch<C> batch) {
...
-                var xDelta = data.x() - xMean;
-                var yDelta = data.y() - yMean;
+                double xDelta = data.x() - xMean;
+                double yDelta = data.y() - yMean;
                 covarianceSum += xDelta * yDelta;
                 varianceSum += xDelta * xDelta;
             }
 
-            if (varianceSum == 0) {
+            if (Math.abs(varianceSum)<=Double.MIN_NORMAL) {
```
douira: nitpick: formatting

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
```diff
@@ -5,10 +5,10 @@
 import java.util.Locale;
 
 public abstract class Average1DEstimator<C> extends Estimator<C, Average1DEstimator.Value<C>, Average1DEstimator.ValueBatch<C>, Void, Long, Average1DEstimator.Average<C>> {
-    private final float newDataRatio;
+    private final double newDataRatio;
     private final long initialEstimate;
 
-    public Average1DEstimator(float newDataRatio, long initialEstimate) {
+    public Average1DEstimator(double newDataRatio, long initialEstimate) {
```
douira, line 11: Should the constants in `JobDurationEstimator` also be doubles?

common/src/main/java/net/caffeinemc/mods/sodium/client/util/MathUtil.java
```diff
@@ -54,6 +54,10 @@ public static float exponentialMovingAverage(float oldValue, float newValue, flo
         return newValueContribution * newValue + (1 - newValueContribution) * oldValue;
     }
 
+    public static double exponentialMovingAverage(double oldValue, double newValue, double newValueContribution) {
```
douira, line 53: I think the float-based exponentialMovingAverage is now unused.

douira (changes requested):
thanks for the PR!

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/estimation/Average1DEstimator.java
MCRcortex: Dont believe it matters (as afak its an exact conversion) however have changed both `JobDurationEstimator` and `UploadDurationEstimator` to doubles

douira (approved):

jellysquid3 (approved):

PR #3297, opened 2025-10-15
Perspective-Based Visibility Data Selection

douira:
Performs the generation of visibility data for a set of 8 perspectives, which each enable a different set of 3 allowed traversal directions through the chunk, and then combines them based on the camera's relative position to each chunk. As a side effect, this also prevents visibility paths through sections that contain opposing step directions from being accepted. Such paths are always invalid since they can never contain a straight view ray.

This implementation has some optimization potential, but I have also not measured the performance of the visibility data generation so I don't know how necessary it is. Feel free to review and suggest (or make) improvements. I'm going to keep this as a draft until we've tested and discussed it more thoroughly.

See discussion of my recent ideas on valid and invalid paths [here](https://discord.com/channels/602796788608401408/651120262129123330/1427793297413247056). This is the implementation of [this concept](https://discord.com/channels/602796788608401408/651120262129123330/1294721219312226325) I suggested last year.

Screenshot_2024-10-12_at_20 11 21

In my brief test, this reduced the section count from 4820 to 4512 in a simple overworld scene. More testing of its effectiveness and correctness is required.

Screenshot_2025-10-15_at_03 39 12
Screenshot_2025-10-15_at_03 41 16

A demonstration of how it can cull sections behind this section where no view ray is possible, which was previously not culled. (with all non-air blocks turned "opaque" for visualization)

Screenshot_2025-10-15_at_03 47 17

Prerequisite: https://github.com/CaffeineMC/sodium/pull/3484


douira:
Merged as part of https://github.com/CaffeineMC/sodium/pull/2887

PR #3303, opened 2025-10-22
Play click sound from SliderControl

JFronny:
The slider control in vanilla Minecraft plays the click sound when it is released.
This modifies the custom slider control in Sodium to do the same.


douira:
Thanks for having taken initiative with this PR!

PR #3307, opened 2025-10-26
add slope-refinement based occlusion culling to cull ~25% more chunks

rmmh:
visualization to build intuition is here: https://mod.ifies.com/f/251025_raycast_vis_v4.html

~~I was hoping that Vector3f would help more but at this point I might as well just bite the bullet and add 12 float member variables.~~

this currently breaks when outside of the world bounds (like too high up)-- no chunks render


douira:
Merged as part of https://github.com/CaffeineMC/sodium/pull/2887

PR #3313, opened 2025-10-31
More Accurate Buffer Sizing

### LINKED ISSUE #3300: Implement More Accurate Buffer Sizing
Previous to 0.7, we were allocating too large buffers on the GPU because of a bug in our size calculation. Once [we fixed this](https://github.com/CaffeineMC/sodium/commit/43f8b253ccdd71789429862f96376031d879768f), more appropriately sized buffers were being used once again, but Intel drivers seem to have issues when the buffers are resized very often, leading to terrible performance during chunk loading.
Previous issue that seems to be related to this: https://github.com/CaffeineMC/sodium/issues/3295 (new issue since the old issue is incomprehensible)
The goal is an implementation that sizes buffers such that they are resized as few times as possible while avoiding excessive (video) memory usage. I have some ideas on how to do this, which we discussed on discord [here](https://discord.com/channels/602796788608401408/651120262129123330/1429864772941123604).

douira:
Improved buffer size estimation leading to far fewer buffer allocations and much less resizing.

- Reuses buffers instead of freeing and reallocating them
- Uses the first 16 allocated segments in a region's buffer to estimate the total required size of the region buffer to avoid resizing it again in the future
- Resizes small buffers where size estimation would be imprecise with a growth factor instead of a small increment
- Avoids attempting to upload a queue of segments when the total remaining space is less than the total required for the segments in the upload queue. Instead it immediately performs resizing, thereby avoiding some CPU-side insertion effort and unnecessary copy commands.
- Adds a bunch of buffer-related debug counters to the F3 screen, these will likely get removed before this PR is merged.
- The buffer reuse list is a global field on `GlBufferArena` which means that any buffers allocated with it will share the same reuse pool, which currently is only the terrain. We may want to change this architecture if this class should be able to serve multiple different allocation contexts, which it currently does not.

Fixes https://github.com/CaffeineMC/sodium/issues/3300, but without implementing a "young gen" buffer because the current approach also seems to work well. My testing indicates that having a larger sample of sections to get the average size from does not improve the result much. Together with buffer reuse, this largely avoids the small initial buffers from going to waste, which is what a young gen buffer would primarily be for. It would still reduce the total number of allocations, but for now I think this is a good approach. (and less complex)

I may investigate adding a young gen buffer for reducing the number of very small index buffer allocations in the future.

Does not do https://github.com/CaffeineMC/sodium/issues/2683. There's no buffer sharing going on here and the points in that issue are still valid.

This will be a draft PR until it's been validated to actually fix the problem and not cause other problems.

If you want to test this and give feedback as a user, [please do so here](https://discord.com/channels/602796788608401408/1433967424222527519/1433967424222527519) on [our discord server](https://jellysquid.me/discord).

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
@@ -11,15 +11,21 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class GlBufferArena {
     static final boolean CHECK_ASSERTIONS = false;
 
-    private static final GlBufferUsage BUFFER_USAGE = GlBufferUsage.STATIC_DRAW;
+    // how many segments we require to be present before we calculate an average size
+    public static final int MIN_SEGMENTS_FOR_AVG = 16;
+    // growth factor to use when we have too few segments present
+    public static final float FEW_SEGMENTS_GROWTH_FACTOR = 1.5f;
+    // factor to use when we are allocating with an expected size
+    public static final float EXPECTED_SIZE_TARGET_FACTOR = 1.5f;
+    // how much bigger than requested a buffer can be to be considered for reuse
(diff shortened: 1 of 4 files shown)


douira:
Force pushed to remove instrumentation and debug, content should not have changed compared to the previous version, which we did user tests on. A brief test on my end indicated no regression against the tested commit. I'm converting this to a non-draft PR as I think it's ready.

PR #3351, opened 2025-11-24
Update neoforge.mods.toml dependencies

### LINKED ISSUE #3350: Fix Dependency Declarations on NeoForge
The current NeoForge dependency declaration on 1.21.4 doesn't make any sense. We should be using a proper version range, and consider actually setting a firm version requirement that makes sense.

https://github.com/CaffeineMC/sodium/blob/dev/neoforge/src/mod/resources/META-INF/neoforge.mods.toml

See https://discord.com/channels/602796788608401408/602805971391217683/1441064005920886856

Abbie5:
Fixes #3350

### DIFF
--- a/neoforge/src/mod/resources/META-INF/neoforge.mods.toml
+++ b/neoforge/src/mod/resources/META-INF/neoforge.mods.toml
@@ -26,14 +26,14 @@ provides = ["indium"]
 [[dependencies.sodium]]
 modId = "minecraft"
 type = "required"
-versionRange = "1.21.4"
+versionRange = "[1.21.10,)"
 ordering = "NONE"
 side = "CLIENT"
 
 [[dependencies.sodium]]
 modId = "neoforge"
 type = "required"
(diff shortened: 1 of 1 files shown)


IMS212:
so... I don't think any of these checks are working, given people can use Sodium. I would like to find out why before merging this.

douira:
The `versionRange = "1.21.4"` check apparently means `>=1.21.4`, so it working would be reasonable. Did you test whether it works as expected after the patch?

Abbie5:
I had not tested until now, sorry.

I have now tested and found that both with and without the change, the game starts but silently does not load sodium when on a version below the one it is built against, and loads the mod normally otherwise.

This leads me to believe that the dependencies declared in the file `neoforge/src/mod/resources/META-INF/neoforge.mods.toml` do not have any effect at all, and may be removed.

douira:
This is probably because of the nested mod within the service. Maybe we need to define the constraints on the top-level (service) mod instead?

PR #3374, opened 2025-12-07
Faster RGSS Sampling

douira:
I've measured that this improves performance in a fragment-heavy scene by around 4-5%.

Patch written by @GeForceLegend https://mojira.dev/MC-304778, used with permission.

Neoforge curently broken, so the neoforge part is disabled.

### DIFF
--- a/common/src/main/resources/assets/sodium/shaders/blocks/block_layer_opaque.fsh
+++ b/common/src/main/resources/assets/sodium/shaders/blocks/block_layer_opaque.fsh
@@ -68,10 +68,6 @@ vec4 sampleRGSS(sampler2D source, vec2 uv, vec2 pixelSize) {
 
     float mipLevelExact = max(0.0, log2(effectiveDerivative / minPixelSize));
 
-    float mipLevelLow = floor(mipLevelExact);
-    float mipLevelHigh = mipLevelLow + 1.0;
-    float mipBlend = fract(mipLevelExact);
-
     const vec2 offsets[4] = vec2[](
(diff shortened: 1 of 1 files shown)


GeForceLegend:
Vanilla transparent terrain uses RGSS/AF too. I don't know if sodium applied that on water too, but looks like you only edited solid/cutout terrain shader according to file name?

douira:
The file name is somewhat of a misnomer. This fragment shader runs for all three render passes, including translucent. (I just checked)

douira:
Performance testing (see [here](https://discord.com/channels/602796788608401408/1447334433752617022)) has yielded inconclusive results. I personally observed improvements, but multiple other users have observed slight regressions of around the same magnitude. There seems to be a tentative correlation with AMD GPUs doing better and Nvidia GPUs doing worse, but we don't have enough data to confirm this trend. I think this behavior is weird because doing less sampling should intuitively make it faster or at least change nothing.

GeForceLegend:
I'm using NVIDIA GPU and vanilla resourcepack test shows reducing samples by fractional mip level obviously improves performance. With 16x16x16 trial spawner aligned to render chunk and player standing on the top looking down, performance on my device (R7 7700X, RTX4070) is: No filtering 1000FPS, Vanilla RGSS 550FPS, my RGSS 700FPS.

GPU trace capture provided in that mojira ticket also tested with same scene, which shows it almost make terrain fragment shader runs 2x faster. No clue why it will lost performance with your user test on NVIDIA.

Issue #3318, opened 2025-11-02
[AMD/Windows] Since the v25.10.2 update, multiple launchers (Lunar, Feather, etc) cannot render the world


jellysquid3:
## Synopsis

When some third-party launchers (Feather, Lunar, etc) are used with Sodium, the world stops rendering correctly due to what we believe is our [shared index buffer](https://github.com/CaffeineMC/sodium/blob/dev/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/SharedQuadIndexBuffer.java) being corrupted. This issue first appeared with version 25.10.2 of the AMD Graphics Driver, and user reports have rapidly increased in frequency as the update has rolled out to more computers over the last week.

We have not observed this behavior with any of the other launchers we support (Prism, Modrinth, and Mojang's launcher.)

## Analysis

The sequence of calls into the driver for allocating this buffer and writing the index data is especially trivial: 

```c
GLuint buffer;
glGenBuffers(1, &buffer);
glBindBuffer(GL_ARRAY_BUFFER, buffer);
glBufferData(GL_ARRAY_BUFFER, length, NULL, GL_STATIC_DRAW);
void* ptr = glMapBufferRange(GL_ARRAY_BUFFER, 0, length, GL_MAP_WRITE_BIT | GL_INVALIDATE_BUFFER_BIT);

// ... write data into (ptr)

glUnmapBuffer(GL_ARRAY_BUFFER);
```

We're very confident this code is correct, and it has never caused problems before. 
Yet for some reason, calling `glMapBufferRange` produces nonsensical errors from the AMD graphics driver:

```
Driver function parameter <target> has an invalid enum '0x3' (GL_INVALID_ENUM)
```

(_Note that even more strangely, the "invalid enum" `0x3` is an integer that increases by one each time `glMapBufferRange` is called -- this makes no sense, and this isn't even a valid error to return._)

jellysquid3:
At this point, seeing as we cannot find a defect in our code, and this problem only occurs with launchers we do not provide support for anyway, we are considering using the compatibility checks module in Sodium to blacklist the specific combination of these launchers and driver version.

Icyzel:
whens this gonna be fixed?

douira:
[...]

PR #3391, opened 2025-12-11
Add AMD Workarounds

mjwells2002:
This is more or less a direct copy of the NVIDIA workarounds in function and form.

Fixes issue #3318 by applying a workaround that prevents game detection when an AMD GPU is detected on windows.

In an ideal world this would also detect the driver version and only apply if the version is newer then the version this started, but the WDDM version scheme is not easily decodable and the NVIDIA workarounds do not seem to check for a driver version

some testing notes

the AMD drivers appear apply there patches when 3 conditions are met.
executable name is `javaw.exe`,
process arguments contain `\.minecraft`
and, process arguments contain `--version` with a valid modern game version like `--version 1.21.11`

This parsing of the version is why this issue does not effect users of fabric with the default launcher is it will specify a version string like `--version fabric-loader-0.18.2-1.21.11` which does not parse correctly, whereas other launchers with the issue use `--version 1.21.11-fabric0.18.2`

This also appears to be the first time AMD have done detection for Minecraft, and the optimization is broken with sodium. So anyone in a situation where the game is working currently is also in a situation where this change makes no difference. Meaning this should make no difference in performance for users who do not have the issue. as the AMD optimization being applied and the game working with sodium are mutually exclusive.

If you wish to force the game to be detected and apply the optimization using a launcher like prism launcher you can add a JVM argument like `-Damd=" \\.minecraft --version 1.21.11 "` this ensures the instance satisfies the `--version`  and `\.minecraft` requirements.

This may not be the ideal way of solving this, but it is similar to how NVIDIA drivers are already handled in sodium.

### DIFF
--- a/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/workarounds/Workarounds.java
+++ b/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/workarounds/Workarounds.java
@@ -1,6 +1,7 @@
 package net.caffeinemc.mods.sodium.client.compatibility.workarounds;
 
 import net.caffeinemc.mods.sodium.client.compatibility.environment.OsUtils;
+import net.caffeinemc.mods.sodium.client.compatibility.workarounds.amd.AmdWorkarounds;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.intel.IntelWorkarounds;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.nvidia.NvidiaWorkarounds;
 import org.slf4j.Logger;
@@ -39,6 +40,10 @@ public class Workarounds {
 
         if (NvidiaWorkarounds.isNvidiaGraphicsCardPresent()) {
             workarounds.add(Reference.NVIDIA_THREADED_OPTIMIZATIONS_BROKEN);
+        }
+
+        if (AmdWorkarounds.isAmdGraphicsCardPresent()) {
+            workarounds.add(Reference.AMD_GAME_OPTIMIZATION_BROKEN);
         }
 
         if (IntelWorkarounds.isUsingIntelGen8OrOlder()) {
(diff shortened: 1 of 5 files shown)


vogon-poetry:
> If you wish to force the game to be detected and apply the optimization using a launcher like prism launcher you can add a JVM argument like -Damd=" \.minecraft --version 1.21.11 " this ensures the instance satisfies the --version and \.minecraft requirements.

I can confirm this works to trigger the issue but it needs to be foreword slash in my case, i.e. /.minecraft 
no idea why

jellysquid3 (approved):
The implementation here looks fine, but it needs verification for functionality.

mjwells2002:
> > If you wish to force the game to be detected and apply the optimization using a launcher like prism launcher you can add a JVM argument like -Damd=" .minecraft --version 1.21.11 " this ensures the instance satisfies the --version and .minecraft requirements.
> 
> I can confirm this works to trigger the issue but it needs to be foreword slash in my case, i.e. /.minecraft no idea why

What driver version, i tested and confirmed this on a system running driver 25.20.21.01/25.10.2 (driver version/adrenaline software version)

jellysquid3:
Honestly, it might make sense to just apply this workaround *all the time* regardless of vendor, because seemingly none of them can be trusted to not completely break themselves in the name of juicing benchmark scores.

mjwells2002:
> The implementation here looks fine, but it needs verification for functionality.

This should be tested on more systems, i tested and confirmed this on windows 11 with driver version 25.10.2/25.20.21.01 (2 version numbers for the same driver, depending on what component you look at)
on a system with a 9070xt

TheDGOfficial:
Is there any reason that the logs have this in a Linux Mesa system? From looking at both the original issue and the workaround code, the workaround is not actually applied unless on windows, so this log message seems misleading.

```
[20:33:14] [main/WARN]: Sodium has applied one or more workarounds to prevent crashes or other issues on your system: [AMD_GAME_OPTIMIZATION_BROKEN]
[20:33:14] [main/WARN]: This is not necessarily an issue, but it may result in certain features or optimizations being disabled. You can sometimes fix these issues by upgrading your graphics driver.
```

jellysquid3:
It shouldn't try to enable any workaround for this on Linux. Since the PEB structure doesn't exist on Linux, it would just crash if it tried, but evidently it never gets to that point. But yes, there is probably a bug if you're seeing that in your log file.

Issue #3404, opened 2025-12-17
Prevent Pushing Mismatching Tags


douira:
We should use git hooks to check that when creating a tag it matches the currently set version in the buildscript.

douira:
Obviated by automated publishing

PR #3410, opened 2025-12-18
pre-push hook for checking for mismatched versions

IThundxr:


douira:
Fixes #3404

douira:
Since apparently IMS uses git on Windows, we'd need a Windows implementation of this too to use it.

douira:
Obviated by #3575

PR #3475, opened 2026-01-15
Fix unecessary workaround being applied on XWayland and on GLFW >= 3.4

redactedontop:


redactedontop:
Closed due to not following contributing guidelines.

Issue #3446, opened 2025-12-31
Allow No Error Context to be used on Linux/Wayland when newer versions of GLFW are installed


redactedontop:
### Request Description

There's no need for this anymore. Modern GLFW has this fixed.

jellysquid3:
GLFW may have fixed this, but has Minecraft actually upgraded to the latest version of GLFW?

redactedontop:
When using wayland, you wanna use system GLFW anyways - that's the point. Either use patched GLFW or WayFix with system GLFW. System GLFW works decently now iirc but is lacking. Bundled GLFW is a no-go on Wayland. Atleast give us an option to disable the workaround, please.

douira:
This still doesn't answer the question whether this PR is valid.

redactedontop:
This isn't a PR. And I gave the answer: not sure, but no one in their right mind will use bundled GLFW with wayland.

jellysquid3:
> ... but no one in their right mind will use bundled GLFW with Wayland.

While I agree the experience on Linux otherwise is terrible, we need to make sure the game doesn't outright crash for people who are not using additional mods.

https://github.com/glfw/glfw/commit/eeeb56eb23c9a23866e8042fb23d7b0b4d96d106 fixes the problem, which looks like it was published with v3.4.0. We can add logic to the existing workaround to not apply it when the newer version is being used.

redactedontop:
Yeah, you're right. Should I PR?

douira:
That would be great

douira:
I'm seeking to test & publish 0.8.3 soon. What is the status of this?

redactedontop:
PRing RN.

redactedontop:
#3477

douira:
Closed by https://github.com/CaffeineMC/sodium/commit/1d2942a247f87a3cb1ae268166190b6ec267046e

jellysquid3:
For context on why we are now unconditionally disabling this: https://github.com/CaffeineMC/sodium/issues/3488#issuecomment-3797382965

There doesn't to be any reliable method for detecting whether the user has installed a "patched" version of GLFW. We could maybe detect if they have other certain mods installed or if the runtime search paths are strange, but it's all very fragile.

PR #3478, opened 2026-01-15
Improve NO_ERROR_CONTEXT_UNSUPPORTED workaround criteria

redactedontop:


redactedontop:
Works for me. A bit more FPS.

redactedontop:
Fixes #3446

barnabwhy:
Looks like there is a logic error in this PR
You're checking `!(glfwMajor[0] >= 3 && glfwMinor[0] >= 4)`
Which would work as intended for any version 3.x.x that is at least 3.4.x but in the future would fail on 4.x.x below 4.4.x

You need to instead check as follows
```java
!(
    glfwMajor[0] > 3 ||
    glfwMajor[0] == 3 && glfwMinor[0] >= 4
)
```

Maybe there is some way to write this cleaner but the > and == paths need to be different because for major > 3 you want it to always match and skip the minor check

umbrageodotus:
Alrighty, thanks.

douira:
Has the logic error been resolved?

douira (approved):

PR #3480, opened 2026-01-15
Fix Quad Ordering Stability in Dynamic TS Fast Path

douira:
Ensure quads are sorted by quad index when their dot products are same in the SNR fast path, which directly transforms the points array, for generating multi-leaves in the BSP tree. this fixes a bug with mesh-baked banners in BBE as reported by @ceeden


douira:
Branch accidentally deleted, nothing to see here

PR #3482, opened 2026-01-17
Add a text highlight on page which is hovered.

7777777-4547:


douira:
can confirm this works

douira (approved):

PR #3484, opened 2026-01-18
Fix Rare Wrongly Culled Sections When Above World

douira:
Fix a rare bug where sections would be missed during occlusion culling when above the world.

This was caused by the above-world init adding every section of the top of the world to the queue at once, resulting in some sections being visited too early such that they are not traversed through with all potentially incoming directions but marked as finished.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
@@ -19,6 +19,10 @@ public class OcclusionCuller {
 
     private final DoubleBufferedQueue<RenderSection> queue = new DoubleBufferedQueue<>();
 
+    private int outOfWorldRadius;
+    private int outOfWorldHeight;
+    private int outOfWorldDirection;
+
     public OcclusionCuller(Long2ReferenceMap<RenderSection> sections, Level level) {
         this.sections = sections;
         this.level = level;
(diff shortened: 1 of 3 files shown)


Kroppeb:
Rip #1393

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -19,6 +19,10 @@ public class OcclusionCuller {
 
     private final DoubleBufferedQueue<RenderSection> queue = new DoubleBufferedQueue<>();
 
+    private int outOfWorldLayer = -1;
```
Kroppeb: `outOfWorldLayer` is a confusing name IMHO. I think `outOfWorldRadius` would fit better? (or `...currentRadius` or `...nextRadius`)

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
```diff
@@ -33,9 +37,22 @@ public void findVisible(RenderSectionVisitor visitor,
...
+        }
 
         while (queues.flip()) {
+            if (this.outOfWorldLayer > 0) {
+                this.initOutsideWorldHeight(queues.write(), viewport, searchDistance, frame);
+                this.outOfWorldLayer++;
+            }
+
             processQueue(visitor, viewport, searchDistance, useOcclusionCulling, frame, queues.read(), queues.write());
         }
```
Kroppeb, line 57: I wonder if you should insert this here? I'd assume it's rare for there to be loaded chunks, then a gap of unloaded chunks and then a gap again of loaded chunks so it's probably overkill.

```
    if (this.outOfWorldLayer > 0) {
            while (initWriteQueue.isEmpty() && this.initOutsideWorldHeight(initWriteQueue, viewport, searchDistance, frame)) {
                this.outOfWorldLayer++;
            }
    }
```

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
Kroppeb, line 31: Can this be called multiple times? Or only once per instance?

Kroppeb (commented):
Looks good to me

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
douira: Previously this was called `layer` before I dissolved the loop in `initOutsideWorldHeight`. I could rename it.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
douira: I worry that could introduce path length inequality when there's a region of sections that are not loaded but also aren't diamond-shaped such that we initialize some adjacent columns in the same cycle, which is what this PR aims to prevent.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/occlusion/OcclusionCuller.java
douira: I've fixed that the out of world init fields were missing a reset. This method can be called multiple times.

PR #3490, opened 2026-01-28
Fix incorrect UV mapping for flowing fluid side textures

### LINKED ISSUE #2471: Fix incorrect mapping of fluid side textures
### Request Description

The side textures for fluids are not mapped correctly in vanilla or Sodium:
- While the top faces correctly map to the center of the texture file, the side faces map to the top left of it, for some reason. (https://bugs.mojang.com/browse/MC-267861)
- In addition, the texture is mapped horizontally flipped from what it should be. (https://bugs.mojang.com/browse/MC-267863)

Since fluids don't use proper json models even ten years later, this cannot be fixed via resource packs, and requires code changes. Since there are other issues with vanilla fluid rendering that I have reported and am still to report, I figured this report would fit nicely among them.

This issue is quite hard to see with vanilla's water and lava textures, so I'd recommend using the resource pack attached to MC-267861 to see the issue in action.

Note: when fixing the flipped textures issue, the interior of the fluid should be made to render horizontally flipped instead. This is to ensure that every pixel on the outside corresponds to the same pixel on the inside, as is the case for blocks such as powder snow, and in 1.20.3+, spawners and trial spawners.

ccristopher:
## Description
This PR fixes visual artifacts on the side faces of flowing fluids (water/lava) where the texture mapping appeared incorrect.

## Changes
Modified `DefaultFluidRenderer.java` to adjust the UV coordinates for horizontal fluid faces.
* Updated `u1` and `u3` sampling offsets to center the texture correctly.
* Adjusted `v` coordinate calculations to match the correct texture quadrant.

Screenshot 2026-01-27 at 11 52 01 PM

Screenshot 2026-01-27 at 11 51 15 PM

## Related Issues
Fixes #2471

## Testing
- [x] Verified that flowing fluid sides map to the center of the texture file.
- [x] Verified that flowing fluid sides are no longer mapped mirrored.
- [x] Ensured every pixel on the outside of fluid corresponds to the same pixel on the inside.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/DefaultFluidRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/DefaultFluidRenderer.java
@@ -316,11 +316,11 @@ public class DefaultFluidRenderer {
                     }
                 }
 
-                float u1 = sprite.getU(0.0F);
-                float u2 = sprite.getU(0.5F);
-                float v1 = sprite.getV((1.0F - c1) * 0.5F);
-                float v2 = sprite.getV((1.0F - c2) * 0.5F);
-                float v3 = sprite.getV(0.5F);
+                float u1 = sprite.getU(0.75F);
+                float u2 = sprite.getU(0.25F);
+                float v1 = sprite.getV(0.25F + (1.0F - c1) * 0.5F);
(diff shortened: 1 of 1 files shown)


douira (approved):
Tested it, looks good to me.

douira:
It appears this PR has produced this regression: https://github.com/CaffeineMC/sodium/issues/3523

Issue #3506, opened 2026-02-13
GeometryPlanes crash


gyurix:
### Bug Description

I got this crash report when using the latest sodium 0.8.4 with Fabric 1.21.11.
Please fix it.

### Reproduction Steps

Unsure, it just happened randomly, try to figure it out based on the crash report.

### Crash Report

```
java.lang.NullPointerException: Cannot invoke "net.caffeinemc.mods.sodium.client.render.chunk.translucent_sorting.trigger.GeometryPlanes.getPlanesForNormal(net.caffeinemc.mods.sodium.client.render.chunk.translucent_sorting.trigger.NormalList)" because "geometryPlanes" is null
	at knot//net.caffeinemc.mods.sodium.client.render.chunk.translucent_sorting.trigger.GFNITriggers.integrateSection(GFNITriggers.java:75)
	at knot//net.caffeinemc.mods.sodium.client.render.chunk.translucent_sorting.trigger.SortTriggering.integrateTranslucentData(SortTriggering.java:215)
	at knot//net.caffeinemc.mods.sodium.client.render.chunk.RenderSectionManager.processChunkBuildResults(RenderSectionManager.java:452)
	at knot//net.caffeinemc.mods.sodium.client.render.chunk.RenderSectionManager.uploadChunks(RenderSectionManager.java:379)
	at knot//net.caffeinemc.mods.sodium.client.render.SodiumWorldRenderer.setupTerrain(SodiumWorldRenderer.java:249)
	at knot//net.minecraft.class_761.method_74752(class_761.java:10743)
	at knot//net.minecraft.class_761.method_22710(class_761.java:497)
	at knot//net.minecraft.class_757.method_3188(class_757.java:785)
	at knot//net.minecraft.class_757.method_3192(class_757.java:497)
	at knot//net.minecraft.class_310.method_1523(class_310.java:1393)
	at knot//net.minecraft.class_310.method_1514(class_310.java:966)
	at knot//net.minecraft.client.main.Main.main(Main.java:250)
	at net.fabricmc.loader.impl.game.minecraft.MinecraftGameProvider.launch(MinecraftGameProvider.java:514)
	at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:72)
	at net.fabricmc.loader.impl.launch.knot.KnotClient.main(KnotClient.java:23)
```

douira:
Please come to [our discord server](https://jellysquid.me/discord) to receive help with this support request.

PR #3512, opened 2026-02-16
Fix NPE in GFNI trigger integration when geometry planes are discarded

gyurix:
## Summary
Fixes a null-pointer crash reported in #3506 during translucent sorting trigger integration.

## Cause
`DynamicData` can discard its `geometryPlanes` after integration (`discardGeometryPlanes()` sets it to `null`), but `GFNITriggers.integrateSection(...)` always dereferenced `data.getGeometryPlanes()` without a null check.

## Fix
- Added a null guard in `GFNITriggers.integrateSection(...)` and return early when `geometryPlanes` is `null`.
- Marked `DynamicData#getGeometryPlanes()` as `@Nullable` to make the lifecycle contract explicit to callers.

## Validation
- Full project build completed successfully:
  - Command: `./gradlew build`
  - Result: `BUILD SUCCESSFUL`


douira:
The first part of this change is invalid because sections are only integrated once so the null would never cause this crash. We know that this crash is typically caused by using other incompatible mods such as nvidium which violate our state constraints.
If you had properly followed our issue template and posted a gamelog I could have explained that this issue is typically caused by incompatible mixins to sodium.
Marking the field as nullable is a good idea, but not particularly useful given this is the only place its used.

PR #3516, opened 2026-02-19
FrustumIntersection Accessor

JustAHuman-xD:
#3099 used a method handle instead of accessor mixin, this just switches it to use an accessor


douira:
Turns out we can't do this on NeoForge.

PR #3522, opened 2026-02-24
Round capacity estimates to multiples of 4 to make sure allocations are always aligned

### LINKED ISSUE #3509: Incorrect assumption that buffer size matches the stride of primitive vertex count
### Bug Description

Starting Sodium 0.7.3, it is not guaranteed anymore that `(gl_VertexID % 4)` matches exactly with the quad boundary. This does not seem to visibly affect Sodium, but affects Iris shaders that rely on this. 
After investigating, [this commit](https://github.com/CaffeineMC/sodium/commit/7fd731ecd3a9589cc557d2f9799ce1df8082f208) exposed an underlying issue with the way Sodium handles rendering. In particular, the size estimate may not be aligned to a multiple of 4 anymore. While this shouldn't matter, as the geometry still fits in this buffer, it seems that Sodium relies on the byte size of the buffer instead of the actual geometry count. 

A bandaid fix is to round up the new capacity estimate to make sure it is divisible by 4: however this does not fix the underlying issues.

### Reproduction Steps

N/A

### Log File

N/A

### Crash Report

N/A

xirreal:
Closes https://github.com/CaffeineMC/sodium/issues/3509

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
@@ -379,6 +379,15 @@ public class GlBufferArena {
         } else {
             newCapacity = (long) (requiredTotalSize * FEW_SEGMENTS_GROWTH_FACTOR);
         }
+        // round up to the next multiple of 4
+        // since the new capacity is estimated using non-integers factors, it may end up not being a multiple of 4
+        // this causes three separate issues:
+        // 1. new segments are always allocated at the end of the free segment, but since the tail is calculated from the capacity,
+        //    segments may end up misaligned
+        // 2. terrain is rendered with glDrawElementsBaseCount, and since baseCount may not be even, gl_VertexID % 4 would
+        //    would not be reliable to detect quads as baseCount is added to the vertex ID
+        // 3. misaligned segments on NVIDIA GPUs seem to confuse the driver into splitting quads across workgroups,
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
```diff
@@ -379,7 +379,7 @@ private long estimateNewCapacity(float regionFillFractionInv, List<PendingUpload
         } else {
             newCapacity = (long) (requiredTotalSize * FEW_SEGMENTS_GROWTH_FACTOR);
         }
-        return newCapacity;
+        return (newCapacity + 3) & ~3;
```
douira, line 391: Please add the elaboration you gave on discord and the fact that due to https://github.com/CaffeineMC/sodium/blob/18aa0d1649cd4a4c4b8e6e4b2d7b57aa514d8494/common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java#L248, an unaligned capacity causes an unaligned tail and that causes an unaligned segment.

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/gl/arena/GlBufferArena.java
xirreal: Added explaination of the three issues

douira (approved):

PR #3527, opened 2026-02-28
Improve LWJGL wrong version error message

Moulberry:
Improves the advice message by attempting to figure out the culprit, pointing the user to the mod/launcher that caused it rather than sodium support.

If the jar containing the LWJGL class is `/mods/*`, then the advice says that it is likely caused by the mod.

If the path contains `/lwjgl/`, or the filename starts with `lwjgl-`, then it is likely that the jar is being provided by the launcher itself, so we direct the user to the launcher for support.

If neither, we still put the name of the launcher in the error message to help with diagnosing the cause, even if we aren't sure that it is caused by the launcher.

### DIFF
--- a/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/checks/PreLaunchChecks.java
+++ b/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/checks/PreLaunchChecks.java
@@ -2,12 +2,21 @@ package net.caffeinemc.mods.sodium.client.compatibility.checks;
 
 import net.caffeinemc.mods.sodium.client.platform.PlatformHelper;
 import org.lwjgl.Version;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URL;
+import java.security.CodeSource;
+import java.security.ProtectionDomain;
+import java.util.Locale;
 
 /**
  * Performs OpenGL driver validation before the game creates an OpenGL context. This runs during the earliest possible
  * opportunity at game startup, and uses a custom hardware prober to search for problematic drivers.
  */
 public class PreLaunchChecks {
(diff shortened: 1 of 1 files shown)


common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/checks/PreLaunchChecks.java
```diff
@@ -23,13 +32,61 @@ private static void checkLwjglRuntimeVersion() {
...
+                    This is usually controlled by the settings for a profile or instance in your launcher.
+
+                    If you need assistance fixing the LWJGL version, you should contact ###LAUNCHER###, not Sodium."""
+                    .replace("###LAUNCHER###", launcher);
+        } else if (isCustomLauncher) {
+            advice = """
+                    You seem to be using ###LAUNCHER###.
+
+                    You must change the LWJGL version in your launcher to continue. \
+                    This is usually controlled by the settings for a profile or instance in your launcher.""";
```
jadeposting: `.replace("###LAUNCHER###", launcher);` was forgotten here

jellysquid3:
Can you provide the context for these changes?

Moulberry:
Improves the advice message by attempting to figure out the culprit, pointing the user to the mod/launcher that caused it rather than sodium support.

If the jar containing the LWJGL class is `/mods/*`, then the advice says that it is likely caused by the mod.

If the path contains `/lwjgl/`, or the filename starts with `lwjgl-`, then it is likely that the jar is being provided by the launcher itself, so we direct the user to the launcher for support.

If neither, we still put the name of the launcher in the error message to help with diagnosing the cause, even if we aren't sure that it is caused by the launcher.

IMS212 (approved):
Looks good to me.

PR #3551, opened 2026-03-25
Fix redundant item side quads generated since 1.21.11

Argon4W:
This PR provides a completely rewrite version of `ItemModelGenerator` to avoid generating redundant per-pixel side quads by merging neighbor faces together while keeping the baked model seamless.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
@@ -18,30 +18,19 @@ import org.joml.Vector3f;
 
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


douira:
does the issue this fixes also exist on 1.21.11 or only on 26.1?

Argon4W:
Also appears on 1.21.11

douira (approved):

douira:
vanilla issue: https://mojira.dev/MC-307065

douira:
This commit doesn't work on 1.21.11. If you port it to 1.21.11, which also has this problem according to your comment, we'll merge it for the 1.21.11 version of sodium.

Argon4W:
Okay

Argon4W:
I have seen the conversation on sodium-dev channel. Maybe I could use bitset to rewrite the merge logic to make it less expensive when I have time after I port this to 1.21.11

douira:
The extra time doesn't seem all that bad, but if it was faster that would be nice. My measurements showed 35% more time spent in that method for a 512x resource pack with lots of mergable faces.

PR #3553, opened 2026-03-26
Check for Windows OS before applying AMD workaround

### LINKED ISSUE #3533: Windows-only workaround prints warning on Linux
### Bug Description

A Windows-only workaround prints warning on Linux.

```
[09:55:58] [main/WARN]: Sodium has applied one or more workarounds to prevent crashes or other issues on your system: [NO_ERROR_CONTEXT_UNSUPPORTED, AMD_GAME_OPTIMIZATION_BROKEN]
[09:55:58] [main/WARN]: This is not necessarily an issue, but it may result in certain features or optimizations being disabled. You can sometimes fix these issues by upgrading your graphics driver.
```

https://github.com/CaffeineMC/sodium/issues/3391#issuecomment-3725062264

Most probably needs a OS check before adding AMD_GAME_OPTIMIZATION_BROKEN to the list of Workarounds. Note that the game does not crash so the workaround does not actually try to apply (see comment above), so this a misleading warning and a low priority issue. 

I am in a AMD GPU system, but given that it does not care about OS type before printing warning, it might as well not care about GPU vendor, so additional checks might be needed to ensure warning only gets printed if the workaround is actually applied.

The no error context unsupported warning existed previously so is not the focus of this issue.

### Reproduction Steps

Launch MC 1.21.11 with Sodium 0.8.6 in a Linux Mesa system

### Log File

irrelevant

### Crash Report

irrelevant

TheDGOfficial:
Note that, the workaround is only added to list of Workarounds applied and the message falsely claims the process environment is going to be modified; but the modification was not actually done. So this is mostly a visual only fix.

Fixes #3533

### DIFF
--- a/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/workarounds/Workarounds.java
+++ b/common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/workarounds/Workarounds.java
@@ -41,7 +41,7 @@ public class Workarounds {
             workarounds.add(Reference.NVIDIA_THREADED_OPTIMIZATIONS_BROKEN);
         }
 
-        if (AmdWorkarounds.isAmdGraphicsCardPresent()) {
+        if (AmdWorkarounds.isAmdGraphicsCardPresent() && OsUtils.getOs() == OsUtils.OperatingSystem.WIN) {
             workarounds.add(Reference.AMD_GAME_OPTIMIZATION_BROKEN);
         }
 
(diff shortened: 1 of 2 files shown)


IMS212 (approved):

common/src/boot/java/net/caffeinemc/mods/sodium/client/compatibility/workarounds/Workarounds.java
```diff
@@ -41,7 +41,7 @@ private static Set<Reference> findNecessaryWorkarounds() {
             workarounds.add(Reference.NVIDIA_THREADED_OPTIMIZATIONS_BROKEN);
         }
 
-        if (AmdWorkarounds.isAmdGraphicsCardPresent()) {
+        if (AmdWorkarounds.isAmdGraphicsCardPresent() && OsUtils.getOs() == OsUtils.OperatingSystem.WIN) {
```
IMS212: `operatingSystem` field is usable right above...

PR #3555, opened 2026-03-26
Replace all critical memory code with Unsafe

### LINKED ISSUE #3554: Severe Performance issue on 26.1
This is easily reproduced with Iris but Vogon was able to reproduce with just sodium [here](https://discord.com/channels/602796788608401408/651120262129123330/1486811044067414167).
> so turns out I am able to trigger the problem with only sodium, and in Vanilla too. Its muuuuch worse in vanilla, but with sodium it can still take up like half the frametime. Here is vanilla at 48FPS with <50 entities. Normal FPS is around 900.

Image

> with Sodium it drops from >1K FPS to like 500 in this scene

Image

This is probably related to the switch to Java 25 where `VarHandle` and related mechanisms are apparently broken. This is in part, but not fully, a vanilla issue. (and also an iris issue)

IMS212:
Should fix #3554. Not a nice fix, but a fix.

### DIFF
--- a/buildSrc/src/main/kotlin/BuildConfig.kt
+++ b/buildSrc/src/main/kotlin/BuildConfig.kt
@@ -2,7 +2,7 @@
 
 object BuildConfig {
     val MINECRAFT_VERSION: String = "26.1"
-    val NEOFORGE_VERSION: String = "26.1.0.0-alpha.0+rc-3.20260323.185344"
+    val NEOFORGE_VERSION: String = "26.1.0.3-beta"
     val FABRIC_LOADER_VERSION: String = "0.18.4"
     val FABRIC_API_VERSION: String = "0.144.0+26.1"
     val SUPPORT_FRAPI : Boolean = false
--- a/common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
+++ b/common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
@@ -85,7 +85,7 @@ public class MemoryIntrinsics {
 
     public static long getAddress(long address) {
         if (BITS32) {
-            return UNSAFE.getInt(address);
+            return UNSAFE.getInt(address) & 0xFFFF_FFFFL;
         } else {
             return UNSAFE.getLong(address);
         }
--- a/common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
+++ b/common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
@@ -1,11 +1,13 @@
 package net.caffeinemc.mods.sodium.api.memory;
 
+import org.lwjgl.system.Pointer;
 import sun.misc.Unsafe;
 
 import java.lang.reflect.Field;
 
 public class MemoryIntrinsics {
     private static final Unsafe UNSAFE;
+    private static final boolean BITS32 = Pointer.BITS32;
 
     static {
         try {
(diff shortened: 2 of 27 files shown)


neoforge/src/mod/java/net/caffeinemc/mods/sodium/mixin/platform/neoforge/EntrypointMixin.java
```diff
@@ -17,7 +17,7 @@
  */
 @Mixin(Minecraft.class)
 public class EntrypointMixin {
-    @Inject(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/Options;loadSelectedResourcePacks(Lnet/minecraft/server/packs/repository/PackRepository;)V"))
+    @Inject(method = "<init>", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/Options;<init>(Lnet/minecraft/client/Minecraft;Ljava/io/File;)V", shift = At.Shift.AFTER))
```
douira, line 20: is this change related to this PR?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/workarounds/context_creation/WindowMixin.java
```diff
@@ -17,7 +17,7 @@
 import org.lwjgl.glfw.GLFW;
 import org.lwjgl.glfw.GLFWNativeWin32;
 import org.lwjgl.opengl.WGL;
-import org.lwjgl.system.MemoryUtil;
+import net.caffeinemc.mods.sodium.api.memory.MemoryIntrinsics;
```
douira: I don't think this import makes any sense?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/workarounds/context_creation/RenderSystemMixin.java
```diff
@@ -18,6 +18,7 @@
 import net.minecraft.resources.Identifier;
 import org.lwjgl.glfw.GLFWNativeWin32;
 import org.lwjgl.opengl.WGL;
+import net.caffeinemc.mods.sodium.api.memory.MemoryIntrinsics;
```
douira: this one also not

common/src/main/java/net/caffeinemc/mods/sodium/client/util/NativeBuffer.java
```diff
@@ -6,6 +6,7 @@
 import net.caffeinemc.mods.sodium.client.SodiumClientMod;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import net.caffeinemc.mods.sodium.api.memory.MemoryIntrinsics;
```
douira: unused import

common/src/main/java/net/caffeinemc/mods/sodium/client/gl/shader/ShaderWorkarounds.java
```diff
@@ -19,6 +19,7 @@
 import org.lwjgl.PointerBuffer;
 import org.lwjgl.opengl.GL20C;
 import org.lwjgl.system.MemoryStack;
+import net.caffeinemc.mods.sodium.api.memory.MemoryIntrinsics;
```
douira: unused

buildSrc/src/main/kotlin/BuildConfig.kt
```diff
@@ -2,7 +2,7 @@
 
 object BuildConfig {
     val MINECRAFT_VERSION: String = "26.1"
-    val NEOFORGE_VERSION: String = "26.1.0.0-alpha.0+rc-3.20260323.185344"
```
douira: is this change related to this PR?

common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
```diff
@@ -1,11 +1,13 @@
 package net.caffeinemc.mods.sodium.api.memory;
 
+import org.lwjgl.system.Pointer;
 import sun.misc.Unsafe;
 
 import java.lang.reflect.Field;
 
 public class MemoryIntrinsics {
     private static final Unsafe UNSAFE;
+    private static final boolean BITS32 = Pointer.BITS32;
```
douira, line 10: do we have validation that sodium actually works in 32 bit mode?

douira (changes requested):

neoforge/src/mod/java/net/caffeinemc/mods/sodium/mixin/platform/neoforge/EntrypointMixin.java
IMS212: No, but required for the NeoForge version to run. Was forgotten in 0.8.7.

common/src/main/java/net/caffeinemc/mods/sodium/mixin/workarounds/context_creation/WindowMixin.java
IMS212: Find/replace bug. Will fix.

common/src/api/java/net/caffeinemc/mods/sodium/api/memory/MemoryIntrinsics.java
IMS212: No, however the behavior here is copied identically from old MemoryUtil.

buildSrc/src/main/kotlin/BuildConfig.kt
douira: would you mind making a separate commit for the neoforge update?

douira (approved):

PR #3565, opened 2026-03-30
Try make improved item meshing faster

Argon4W:
The improved item meshing is about 35% slower than vanilla for 512x resource packs with lots of merge-able faces. This draft aims to speed up the improved item meshing by using separate bitset map storage for each side face direction. The storage merges the faces after all per-quad face "bits" have been set, using a continuous for-loop This results in fewer `SideFace` allocations and eliminates remove operations on HashSet. This solution still requires further testing to verify its effectiveness.


douira:
Have you done some profiling on the speedup? It might take some time for me to get around to it. In the mean time if you want, you could also measure how long the relevant methods take before/after this patch using a profiler and this resource pack: [everything upscaled to 512x.zip](https://github.com/user-attachments/files/26337116/everything.upscaled.to.512x.zip)
My protocol was to start the game, then start the profiler, and then reload resources three times, before capturing the profile.

Argon4W:
The average time of `bakeItemQuads` in this solution takes about 13-14ms while the original takes about 17-19ms on my machine. But the result varies greatly between single tests on my machine so I can't tell if this solution is good enough by myself.

douira:
I did some profiling and the total time spent on `compute` with this resource pack over three reloads went from 4491ms to 5762ms (a reduction by 22%).

douira:
Sorry I got my numbers messed up. It actually appears that this patch results in the _slower_ figure of these two. (Unfortunately IJ has no ergonomic way of labeling profiler results so I confused them)

douira:
I did another longer profile with 10 reloads each and unfortunately indeed the patch seems to make it slower. It went from 19649ms to 21584ms. I also noticed my benchmarking pack only works if I unzip it. I don't know why.

douira:
On the other hand I have found a pathological case ([download from discord](https://discord.com/channels/602796788608401408/651120262129123330/1488196382333468713)) where the current beta is broken. If I randomize the alpha values of all pixels in a 512x resource pack, it never finishes (it's been running for 10 minutes) with the current beta, but with this patch it does finish. So the algorithmic complexity of the new algorithm seems to be much better but slightly worse in typical cases.

Argon4W:
Weird, so what’s your point, switching to the new algorithm? I have no idea why this is slower than the previous one

Argon4W:
Theoretically I eliminated the operations on HashSet and the indefinite while-loop. I have totally no idea why this is slower, probably lambda things on computeIfAbsent? Could you help me test it. I’m gonna replace the computeIfAbsent.

Argon4W:
I can't even tell why this is slower, it's sometimes faster sometimes slower in my profiler capture, the visualvm shows that `isTransparent` takes most of the time but the count of `isTransparent` calls is the same between to approaches.

douira:
We've discussed some potential further optimizations that I will mirror here, but I'm going to merge the PR since we want to release soon.
- One idea is to avoid creating any SideFace objects. This would require some significant refactoring, but it might be possible to directly emit quads instead of writing them into a list of SideFaces and then converting those into quads on the builder. I'm not sure how much this would help, it would eliminate writing/reading the array of objects.
- (Completed:) `faceOutput` can be a list instead of a set. I've confirmed empirically that duplicate items never seem to be inserted, and I also don't see a reason for why it needs to be a set.

douira:
Do you have a suggestion how we could mitigate https://github.com/CaffeineMC/sodium/issues/3583 which this improved meshing system regresses on? There's z fighting because the quads are no longer exactly the same so there's floating point errors in the depth calculation.

PR #3569, opened 2026-03-31
Update template_anvil.json

DreamStar-cn:
Add the display setting “on_shelf”.


douira (changes requested):
You've changed the indentation from spaces to tabs. Please change it back so that the diff isn't the whole file.

PR #3571, opened 2026-04-01
Fixes everything being cutout

MCRcortex:
there are probably other issues still with the frapi pr but this fixes 2 of the most important issues


common/src/main/java/net/caffeinemc/mods/sodium/client/render/model/AbstractBlockRenderContext.java
```diff
@@ -238,8 +238,8 @@ public void bufferDefaultModel(BlockStateModelPart part, Predicate<Direction> cu
                 final BakedQuad q = quads.get(j);
                 editorQuad.fromBakedQuad(q);
                 editorQuad.setCullFace(cullFace);
-                editorQuad.setRenderType(q.materialInfo().layer());
                 editorQuad.setAmbientOcclusion(ao.toTriState());
+                editorQuad.setShadeMode(SodiumShadeMode.VANILLA);
```
IMS212: The default ShadeMode should be enhanced, pretty sure

IMS212 (changes requested):
one issue

Issue #3253, opened 2025-09-09
Crash with Blaze3D Validation Layer in NeoForge


AlexInCube:
### Bug Description

I install Sodium in my dev environment and i have crashes. 
The NeoForge maintainer say me to disable Blaze 3D validation layer. 
To do it you need go to `config/neoforge-client.toml` and set `enableB3DValidationLayer` to `false`

### Reproduction Steps

1. Go to `config/neoforge-client.toml` 
2. Set `enableB3DValidationLayer` to `true`

### Log File

[2025-09-09-3.log.gz](https://github.com/user-attachments/files/22224765/2025-09-09-3.log.gz)

### Crash Report

[crash-2025-09-09_04.58.42-client.txt](https://github.com/user-attachments/files/22224748/crash-2025-09-09_04.58.42-client.txt)

douira:
Indeed, it currently does not work with the dev-only validation layer that neoforge adds.

awildergoose:
This could be fixed by changing the casts in these functions for GlTexture and GlDevice to make them use the getRealTexture/getRealDevice functions: if the objects are ValidationGpuTexture/ValidationGpuDevice
https://github.com/CaffeineMC/sodium/blob/472e82a61c0d66fbb4c77e26e97250f806827151/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/shader/DefaultShaderInterface.java#L106
https://github.com/CaffeineMC/sodium/blob/472e82a61c0d66fbb4c77e26e97250f806827151/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/ShaderChunkRenderer.java#L91

I've temporarily fixed this with mixins in my mod until this is fixed on upstream :)

PR #3581, opened 2026-04-07
Port improved item model fix and documentation to 1.21.11 (#3565)

Argon4W:


douira:
This user reports that the PR crashes on neoforge: https://github.com/CaffeineMC/sodium/issues/3583#issuecomment-4200761638

Argon4W:
> This user reports that the PR crashes on neoforge: [#3583 (comment)](https://github.com/CaffeineMC/sodium/issues/3583#issuecomment-4200761638)

I just tested and the neoforge version did not crash. Has the user turned off the `enableB3DVValidationLayer`？(#3253) The validation layer is on by default in dev environment.

douira:
Yeah I don't know what exactly their crash was. I'll validate it on my own and merge if it doesn't break.

Argon4W:
Okay

douira:
It crashed for me but I was able to fix that by doing a gradle clean. Thanks for the contribution!

PR #3585, opened 2026-04-08
Revise hardware compatibility and macOS support details

nzc0der:
Updated hardware compatibility section and added notes for macOS users regarding OpenGL support.


douira:
I don't really see the necessity of this but the other contributors are free to comment.

nzc0der:
Its just to give a heads up to whoever is using it on a Mac (I do) just so they know

PR #3587, opened 2026-04-08
Translatable donation promtp message

StarmanMine142:
Although it may be useless, I would like this message to be translatable.


douira:
Thank you for your effort, but due to the sensitive nature of the donation prompt we aren't currently looking to make this text translatable.

PR #3597, opened 2026-04-15
Support Minecraft 26.1.2 and NeoForge 26.1.2.10-beta

Mikulasz12:
## Summary
- update Sodium to Minecraft `26.1.2`
- update NeoForge support to `26.1.2.10-beta`
- bump Fabric and Loom versions
- fix `NeoForgeLevelRenderHooks` for the `AddSectionGeometryEvent.SectionRenderingContext` change in NeoForge `26.1.2.10-beta` due to https://github.com/neoforged/NeoForge/pull/3090
- add a temporary compatibility shim so the changed NeoForge hook still works on older NeoForge builds
- update the NeoForge Gradle plugin to `2.0.141`
- update Gradle to `9.4.1`

## Main Changes
- update `NeoForgeLevelRenderHooks` for the `SectionRenderingContext` constructor change in NeoForge `26.1.2.10-beta`
- add a compatibility path for both the new constructor and the older `BlockPos` constructor so previous NeoForge builds still work - although this could be removed as adds complexity etc

### DIFF
--- a/fabric/build.gradle.kts
+++ b/fabric/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
     id("multiloader-platform")
 
-    id("net.fabricmc.fabric-loom") version ("1.15.4")
+    id("net.fabricmc.fabric-loom") version ("1.16-SNAPSHOT")
 }
 
 base {
--- a/fabric/build.gradle.kts
+++ b/fabric/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
     id("multiloader-platform")
 
-    id("net.fabricmc.fabric-loom") version ("1.16-SNAPSHOT")
+    id("net.fabricmc.fabric-loom") version ("1.16.1")
 }
 
 base {
--- a/neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelRenderHooks.java
+++ b/neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelRenderHooks.java
@@ -1,12 +1,15 @@
 package net.caffeinemc.mods.sodium.neoforge.level;
 
-import com.mojang.blaze3d.vertex.PoseStack;
 import com.mojang.blaze3d.vertex.VertexConsumer;
+import java.lang.reflect.Constructor;
+import java.util.List;
+import java.util.function.Function;
 import net.caffeinemc.mods.sodium.client.services.PlatformLevelRenderHooks;
 import net.caffeinemc.mods.sodium.client.world.LevelSlice;
 import net.minecraft.client.Camera;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.LevelRenderer;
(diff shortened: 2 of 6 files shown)


neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelRenderHooks.java
```diff
@@ -17,10 +20,15 @@
...
 import org.joml.Matrix4f;
 
-import java.util.List;
-import java.util.function.Function;
-
 public class NeoForgeLevelRenderHooks implements PlatformLevelRenderHooks {
+    /**
+     * NeoForge 26.1.2.10-beta removed the broken section-origin PoseStack from
+     * AddSectionGeometryEvent.SectionRenderingContext (NeoForge PR #3090).
+     * Keep one compatibility path here so Sodium can run against both the new
```
douira: Is there a good reason to support the old neoforge beta?

douira (dismissed):

jellysquid3:
Thanks for the pull request!

Your patch looks reasonable to me. That said, I'd rather just bump the version requirement for NeoForge than keep around a compatibility shim for the older versions. After all, these beta versions are moving quickly, which is the point of them.

fabric/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 plugins {
     id("multiloader-platform")
 
-    id("net.fabricmc.fabric-loom") version ("1.15.4")
+    id("net.fabricmc.fabric-loom") version ("1.16-SNAPSHOT")
```
jellysquid3: Can we pin against a specific version of Loom here, instead of using a SNAPSHOT build? We try to avoid that since it means the behavior of the build script is constantly changing.

Mikulasz12:
Yes true, good point, I thought the compatibility shim would be unnecessary but just wanted to include just in case. I'll remove the compatibility shim and I'll use a specific version of Loom.

jellysquid3 (approved):
Still needs to be verified for functionality, otherwise looks fine.

neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/level/NeoForgeLevelRenderHooks.java
Mikulasz12: No, I mainly added it to avoid forcing users onto the newest NeoForge immediately and to preserve compatibility in the rare instance that another mod depended on this path and hadn’t been updated yet which would result them in having to use a lower version of NeoForge. I treated it as a temporary band-aid, but that was probably unnecessary.

Mikulasz12:
> Still needs to be verified for functionality, otherwise looks fine.

Thank you! i didn't test it much, only made sure the game doesn't crash on world load anymore.

SpacEagle17:
I can also confirm that it no longer crashes on world load :)
Tested with the new ATM11 modpack, which was crashing on world load: https://www.curseforge.com/minecraft/modpacks/all-the-mods-11

Issue #3164, opened 2025-06-18
Multipart blockstates do not correctly inherit block model rendertype


iron431:
### Bug Description

When a multipart blockstate (ie fence blocks) has a part model that utilizes different rendertype than the base model (ie base model has a solid rendertype, and a part model specifies the cutout rendertype) then the part model will incorrectly inherit the base model rendertype, being solid.

### Reproduction Steps

1. Have a base block model which does not specify a rendertype (defaulting to solid)
```{
  "parent": "minecraft:block/cube_all",
  "textures": {
    "all": "irons_spellbooks:block/voidstone"
  }
}
```
2. Have a part model which specifies a rendertype
```
{
	"render_type": "minecraft:cutout",
	"textures": {
		"0": "irons_spellbooks:block/voidstone_trim",
		"particle": "irons_spellbooks:block/voidstone_trim"
	},
	"elements": [
		{
			"from": [0, 16.001, 0],
			"to": [16, 16.001, 16],
			"faces": {
				"up": {"uv": [0, 0, 16, 16], "texture": "#0"}
			}
		},
		{
			"from": [0, 0, -0.001],
			"to": [16, 16, -0.001],
			"faces": {
				"north": {"uv": [16, 0, 0, 16], "texture": "#0"}
			}
		}
	]
}
```
3. Have a multipart blockstate utilizing said models
```
{
  "multipart": [
    {
      "apply": {
        "model": "irons_spellbooks:block/voidstone"
      }
    },
    {
      "when": {
        "north": "false"
      },
      "apply": {
        "model": "irons_spellbooks:block/voidstone_trim",
        "uvlock": false
      }
    },
    {
      "when": {
        "east": "false"
      },
      "apply": {
        "model": "irons_spellbooks:block/voidstone_trim",
        "y": 90,
        "uvlock": false
      }
    },
    {
      "when": {
        "south": "false"
      },
      "apply": {
        "model": "irons_spellbooks:block/voidstone_trim",
        "y": 180,
        "uvlock": false
      }
    },
    {
      "when": {
        "west": "false"
      },
      "apply": {
        "model": "irons_spellbooks:block/voidstone_trim",
        "y": 270,
        "uvlock": false
      }
    },
    {
      "when": {
        "north": "false",
        "east": "false"
      },
      "apply": {
        "model": "irons_spellbooks:block/voidstone_trim_corner",
        "uvlock": false
      }
    },
    {
      "when": {
        "north": "false",
        "west": "false"
[...]

PR #3599, opened 2026-04-19
Fix multipart block model rendering

### LINKED ISSUE #3400: Atmospheric's dragon roots have a weird texture when the dragon fruits are grown.
### Bug Description

this happens with Sodium 0.6.13 in 1.21.1, when a dragon root has dragon fruits / is bonemealed its texture becomes a 2d square with the fruits still visible, no idea why

### Reproduction Steps

bone meal / let grow dragon roots from Atmospheric

Image

### Log File

https://mclo.gs/BnvX999

### Crash Report

https://mclo.gs/bMdU8J5

jamesgreen26:
Fixes NeoForge multipart model rendering so each child part is evaluated with the correct ModelData and render-pass information instead of being rendered through a path that loses top-level pass filtering. 

Fixes https://github.com/SuperMartijn642/Fusion/issues/227 . Likely fixes #3400 and/or #3164 , not tested yet.

### DIFF
--- a/buildSrc/src/main/kotlin/BuildConfig.kt
+++ b/buildSrc/src/main/kotlin/BuildConfig.kt
@@ -2,7 +2,7 @@ import org.gradle.api.Project
 
 object BuildConfig {
     val MINECRAFT_VERSION: String = "1.21.1"
-    val NEOFORGE_VERSION: String = "21.1.143"
+    val NEOFORGE_VERSION: String = "21.1.219"
     val FABRIC_LOADER_VERSION: String = "0.16.9"
     val FABRIC_API_VERSION: String = "0.110.0+1.21.1"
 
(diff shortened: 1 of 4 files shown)


jellysquid3:
Thanks for your contribution.

I have asked that @IMS212 review your pull request, as they were the last person to work on this code, and had previously tried to resolve this exact issue.

jamesgreen26:
Sounds good! Here's #3400 reproduction and fixed:

Screenshot 2026-04-18 at 5 15 08 PM

--

Screenshot 2026-04-18 at 5 17 49 PM

jellysquid3:
I'm not sure the solution proposed here is ideal, ignoring that it is using mixins to patch Sodium's own code.

According to a contributor of the Fabric project:

> The self mixin solves the problem of multipart models that enter the FRAPI pipeline not correctly passing model data to submodels, but what the self mixin does is disable FRAPI for multipart models completely. This obviously breaks functionality but also does not catch all cases (like a model wrapper that uses FRAPI and delegates to a multipart model).

> The proper solution for the FRAPI pipeline is to update Forgified FRAPI (and require the updated version) and copy the Indigo-side changes from https://github.com/Sinytra/ForgifiedFabricAPI/commit/b71c3a6c5e9446ed54da5f18af4f44e597ada9ca. 

Following their guidance here seems like the correct solution, and if you'd like to make a pull request doing so, we'd be happy to accept it. Otherwise, we will try to get to this in the near future.

jamesgreen26:
This makes sense! Thanks for the feedback, I'm not very familiar with FRAPI or Indigo, so I apologize for these oversights. I'll look into your suggested solution when I get a chance.

PR #3600, opened 2026-04-19
Fix multipart block model rendering v2

jamesgreen26:
Follow up on https://github.com/CaffeineMC/sodium/pull/3599

In my testing this seems to resolve the original issue, but I’m not confident that the fix is complete given my limited understanding of FRAPI and Indium. Opening this as a draft for feedback or handoff.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/AbstractBlockRenderContext.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/AbstractBlockRenderContext.java
@@ -33,6 +33,7 @@ import net.minecraft.world.level.block.state.BlockState;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
+import java.util.Stack;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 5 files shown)


douira:
I don't know any of the specifics of this code so I can't give a review beyond that it's syntactically acceptable. I'll request reviews.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/AbstractBlockRenderContext.java
```diff
@@ -102,6 +103,7 @@ public void emitDirectly() {
      * The current model's model data.
      */
     protected SodiumModelData modelData;
+    private final Stack<SodiumModelData> modelDataStack = new Stack<>();
```
SuperMartijn642, line 106: Nothing is ever put onto this stack 🤔

douira:
It appears I've fixed this in my experimental port by means of a somewhat similar implementation: https://github.com/douira/sodium/commit/491aa90020aed0e90d31fd4f0e70a7032c01f2b8

I'll close this as I don't think we're going to be basing future 1.21.1 releases off of the current state of the stable branch, and instead will do something based on the port experiment that includes the linked commit. 
Thank you for taking the first step towards fixing this! @SuperMartijn642 and other reviewers are welcome to review the linked commit as a patch that's adjacent to this work.

Issue #2769, opened 2024-09-22
Add hotkeys for Undo/Apply/Cancel buttons


mk-pmb:
### Request Description

Could you please add a way to reliably activate the Undo/Apply/Cancel buttons via keyboard? Maybe like those underlined characters on buttons when you press the Alt key in most traditional desktop UIs. I tried Alt+U, Alt+A, Alt+C to see if maybe it's secretly implemented, but it had no effect.

I'm trying to automatically adjust some settings via simulated keystrokes. One of my problems is navigating to the undo/apply/cancel buttons after I managed to change the settings, especially because the number of focus-enabled (implies: visible) buttons changes based on whether the new settings actually differ from the old ones. My fallback strategy would be to analyze screenshots and simulate mouse input, but I hope this is a very small change for you and could save a lot of code on my side. It may also have minor usability benefits for other users.

jellysquid3:
I'm fine with the idea of underlining a character on the buttons when ALT is held, and allowing the user to jump to that in the focus tree when ALT+LETTER is pressed. This is typical of Windows applications, even if Minecraft doesn't provide it.

mk-pmb:
> allowing the user to jump to that in the focus tree

Thanks! Even better would be if the Alt key combination would not just jump to the button, but actually fire it, like in Windows or Gnome dialogs. That way, I could reliably fire Apply (if available) and then Done. Currently, the different behavior depending on whether I have to press one or two buttons to save and go back to the Options menu, requires [a workaround activating and cancelling ingame chat.](https://github.com/mk-pmb/minecraft-nether-roof-biomes-marker/blob/861d8e282e8c4ed2f1e9c9fbb522edfd9e3d1412/util/save_or_discard_sodium_settings.sh#L34)

> even if Minecraft doesn't provide it.

Indeed it would be very helpful In the Minecraft menus, but there it's less of a problem, because its menu buttons can reliably be focussed with unchanging arrow key sequences.

marco-fernandes2006-nofc:
Hello, I've been following this issue and the discussion around it.

From what I understand, the goal is to improve interaction with the Undo, Apply, and Cancel buttons, and there was some mention of Alt-based navigation or focus highlighting depending on the system.

I’d be interested in working on this, as I’d like to contribute more to the project.

Before starting, I wanted to check if this is still something that should be implemented, and whether anyone is already working on it.

[...]

PR #3604, opened 2026-04-21
Implement Alt shortcuts and ESC discard behavior (#2769)

marco-fernandes2006-nofc:
## Summary

This PR implements keyboard shortcuts for the main actions in the Video Settings screen, as discussed in the issue.

Users can now use Alt-based shortcuts for Apply, Done, and Undo, with the corresponding letter underlined in each button to make the interaction more discoverable.

For the ESC behavior, it was implemented as an immediate “abort” action: it discards any pending changes and exits the screen.

## Changes

* Added Alt + key shortcuts for:

  * Apply
  * Done
  * Undo
* Underlined the first letter of each button label to reflect the shortcut
* Implemented ESC to discard pending changes and exit the Video Settings screen
* Extracted `applyChanges()` to centralize applying pending configuration changes

## Notes

A confirmation prompt on ESC (e.g. asking whether to apply changes before exiting, optionally with a “don’t show again” checkbox) was considered, I decided not to include it in this PR to keep the scope focused and avoid introducing additional UI/UX complexity here, that behavior could be explored separately as a follow-up improvement if needed

## Issue

#2769

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
@@ -198,15 +198,15 @@ public class VideoSettingsScreen extends Screen implements ScreenPromptable, Scr
             reserveBottomSpace = true;
         }
 
-        this.closeButton = new FlatButtonWidget(new Dim2i(this.getLimitX() - Layout.BUTTON_LONG - ifNotInsetX(Layout.INNER_MARGIN), this.getLimitY() - (ifNotInsetY(Layout.INNER_MARGIN) + Layout.BUTTON_SHORT), Layout.BUTTON_LONG, Layout.BUTTON_SHORT), Component.translatable("gui.done"), this::onClose, true, false);
+        this.closeButton = new KeyBoundButtonWidget(new Dim2i(this.getLimitX() - Layout.BUTTON_LONG - ifNotInsetX(Layout.INNER_MARGIN), this.getLimitY() - (ifNotInsetY(Layout.INNER_MARGIN) + Layout.BUTTON_SHORT), Layout.BUTTON_LONG, Layout.BUTTON_SHORT), Component.translatable("gui.done"), this::onClose, true, false);
         this.addRenderableWidget(this.closeButton);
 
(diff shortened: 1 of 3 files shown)


common/src/main/resources/assets/sodium/lang/en_us.json
```diff
@@ -77,6 +77,7 @@
   "sodium.options.use_no_error_context.tooltip": "When enabled, the OpenGL context will be created with error checking disabled. This slightly improves rendering performance, but it can make debugging sudden unexplained crashes much harder.",
   "sodium.options.buttons.undo": "Undo",
   "sodium.options.buttons.apply": "Apply",
+  "sodium.options.buttons.done": "Done",
```
douira: Is it necessary to have our own translation string for "done"? There's still usage of Minecraft's `gui.done` if there is a good reason to have our own. We'd want to at least be consistent.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
```diff
@@ -324,13 +324,37 @@ public void extractRenderState(@NonNull GuiGraphicsExtractor graphics, int mouse
         }
     }
 
+    // Changes the first letter of a label to be underlined when showKeybinding is true.
+    private Component buildLabel(String labelKey, boolean showKeybinding) {
```
douira: I feel it would be more elegant if be abstracted the label building into its own class of "keybound button" or something of the sort.

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
```diff
@@ -366,6 +390,10 @@ private void undoChanges() {
         ConfigManager.CONFIG.resetAllOptionsFromBindings();
     }
 
+    private void applyChanges() {
+        ConfigManager.CONFIG.applyAllOptions();
```
douira: this method appears redundant to me when we've already established in this class that we're just using `ConfigManager.CONFIG.*` methods directly

douira (changes requested):

PR #3610, opened 2026-04-28
(1.21.11) Upload the correct jar file

UltimatChamp:
All Fabric <26.1 versions require the jar from the `remapJar` task to be uploaded.
The jar from the normal `jar` task can be used for Neoforge and 26.1+


douira:
Thanks for taking the initiative! #3622 is going to replace this.

Issue #3385, opened 2025-12-09
Tracking Issue for Config UI & API Improvements


douira:
PR that fixes some of these aspects: https://github.com/CaffeineMC/sodium/pull/3397

Potential Improvements to be made
- [x] Indicate pages that open external screens more clearly in the page list, by an icon, text, or both
- [x] Make scrollbars thicker for easier use
- [x] Hide search bar clear button when nothing is searched
- [x] Make search bar text have more contrast
- [x] Inset the whole UI into a centered container of reasonable width on large screens
- [x] When inset, space between search bar/close button and page list/search bar?
- [x] Disabled tickboxes don't show the right enablement state
- [x] per-option and global reset by holding down shift
- [x] Clicking on mod headers in the page list should jump to the start of their options in the option list
~~- [ ] Clicking on mod headers in the page list should collapse the page list, in conflict with the previous suggestion. This could also collapse the version information. However, it's bad UX to have two similar elements (mod headers and page headers) in the page list cause such different behaviors.~~
- [x] moving the window from one monitor to the other while the settings ui is open causes a crash because it tries to read a video mode with an index higher than the number of video modes on the new screen
- [x] Adjust spacing in multiple places in the UI to accomodate horizontal and vertical external margins on the content area (separately)
- [ ] Use left/right bouncy text or widen the options to account for some locale requiring more width. Bouncy text is just not nice to look at while widening the options is problematic since on small screens doing so is impossible. We could potentially widen the options slightly only when there's enough space (i.e. instead of adding an inset to the sides, we first widen the options by maybe 40% at most)
- [x] Pressing the escape key when there are pending changes should perform "undo", and then pressing it again should exit the screen.
- [ ] Option text is cut off when italic
- [ ] Reconsider the size of the reset button's click target. On iris it's the whole option width, though I don't find this behavior intuitive.

API Improvements
- [x] Custom update flags (like renderer reload) and registering flag-related runners globally
- [x] Post-save hooks that are syntactic sugar for a unique flag per option [example](https://discord.com/channels/602796788608401408/1447940748036014222/1448278728617492543)
- [x] API for mods to open the config screen and jump to a certain page for use in mod menu integrations. [example](https://discord.com/channels/602796788608401408/1447940748036014222/1448282906144542810)
- [x] Document Option Overlay and Replacement API usage
- [ ] Graphics presets to match vanilla
- [ ] Customization of presents? Adding new presets? adding options from mods to presets through the api?

[...]

PR #3614, opened 2026-04-29
Add performance preset buttons and localization strings for options

chi20253032:


common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
```diff
@@ -183,7 +186,29 @@ private void rebuild() {
         int topBarHeight = Layout.BUTTON_SHORT;
         this.searchWidget = new SearchWidget(this::onSearchResults, new Dim2i(x, y, w, topBarHeight));
 
-        int topBarClear = topBarHeight + ifInsetY(Layout.INNER_MARGIN);
+        int presetButtonW = Layout.BUTTON_LONG;
```
douira, line 189: I'm not convinced adding more buttons to the button stack is a good idea, in particular not for presets since there can be potentially arbitrarily many presets if we allow such customization in the API.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/VideoSettingsScreen.java
```diff
@@ -322,6 +347,57 @@ private void hideDonationButton() {
...
+
+    @SuppressWarnings("unchecked")
+    private static <V> void setOptionValue(Identifier id, V value) {
+        Option option = ConfigManager.CONFIG.getOption(id);
+        if (option instanceof StatefulOption<?> statefulOption) {
+            ((StatefulOption<V>) statefulOption).modifyValue(value);
+        }
+    }
+
+    private void applyPresetValues(boolean maxPerformance) {
```
douira, line 366: I'm not in favor of this way of doing the preset values since it requires us to manually extract preset values from vanilla, and requires a code change any time new presets are added or they change any of the values. If we were to hardcode preset values, we'd do so on the options themselves and not in one giant method like this. This ties into the fact that we want to eventually work on a system that automatically extracts the default values for options from vanilla instead of hardcoding them into our option builder.

douira (changes requested):
As per our tracking issue #3385, we should investigate more generally how we want presets to work within our UI and whether/how mods may interact with them. I think we need more discussion in general on this problem both from a UI/UX design perspective as well as an API perspective.

PR #3616, opened 2026-04-29
Configurable filter mode

pajicadvance:
Adds a config option to allow switching between nearest and linear filtering modes, with the default mode being nearest instead of linear in vanilla. The option name and tooltip are placeholders and should be updated.


douira:
Apparently if I merge with changes it doesn't merge the PR and I instead need to commit to your branch and then merge on github, sorry. PR merged by [f63b055](https://github.com/CaffeineMC/sodium/commit/f63b05510aef76243aadf589bc11a2ab0db07d51).

PR #3631, opened 2026-05-06
Fixed #3603 (incorrect AO and sky light level around light sources)

### LINKED ISSUE #3603: Lack of AO and low sky light level around opaque light sources
### Bug Description

In 26.1, Sodium lacks AO around opaque light sources. And, even after vanilla fixed MC-161150 (tho it's still open, I think they never tracked it) in 26.1, opaque light sources still reducing smoothed sky light level around them, just like vanilla do before 26.1.

### Reproduction Steps

No AO around sea lanterns and crying obsidian, and we can see yellow light map around crying obsidian caused by reduced smooth sky light level (sea lanterns also reduce sky light, but it's not very visible since block light level is nearly maximum)

Image

In vanilla 26.1, everything looks fine

Image

In vanilla 1.21.11, light map around crying obsidian is same as Sodium 26.1, tho with correct AO

Image

### Log File

Nothing related to this issue in log. Only sodium 0.8.9 is loaded on fabric 26.1.

### Crash Report

No crash.

GeForceLegend:
This is a draft because:
- There is comments in [LightDataAccess.java#L92](https://github.com/CaffeineMC/sodium/blob/dev/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/data/LightDataAccess.java#L92) which shows the lack of AO around opaque light sources might be work as intended. However, imo sky brightness during day is much stronger than blocks, so it should create AO at least during day.
- Changes for [AoFaceData.java](https://github.com/CaffeineMC/sodium/blob/dev/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java) can be optimized like split sky and block light level outside that function, so we will only have to do 9 splits instead of 16. But I'm not sure if it meet your tastes.

Changes for AO in LightDataAccess.java is simple enough so I guess I don't have to explain for it.

Changes for smoothed light levels is similar to how vanilla handle it in [net/minecraft/client/renderer/block/BlockModelLighter#L166-169](https://mcsrc.dev/1/26.1.2/net/minecraft/client/renderer/block/BlockModelLighter#L166-169), the [smoothBlend()](https://mcsrc.dev/1/26.1.2/net/minecraft/util/LightCoordsUtil#L61) function do sky and block light levels separately, to avoid the 0 sky light level inside opaque light sources reducing smoothed sky light level.

Fixes #3603

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
@@ -237,16 +237,38 @@ class AoFaceData {
     private static int calculateCornerBrightness(int a, int b, int c, int d, boolean aem, boolean bem, boolean cem, boolean dem) {
         // FIX: Normalize corner vectors correctly to the minimum non-zero value between each one to prevent
         // strange issues
-        if ((a == 0) || (b == 0) || (c == 0) || (d == 0)) {
+        int ab = a & 255;
+        int bb = b & 255;
+        int cb = c & 255;
+        int db = d & 255;
+        if ((ab == 0) || (bb == 0) || (cb == 0) || (db == 0)) {
             // Find the minimum value between all corners
-            final int min = minNonZero(minNonZero(a, b), minNonZero(c, d));
+            final int min = minNonZero(minNonZero(ab, bb), minNonZero(cb, db));
 
             // Normalize the corner values
-            a = Math.max(a, min);
-            b = Math.max(b, min);
-            c = Math.max(c, min);
-            d = Math.max(d, min);
+            ab = Math.max(ab, min);
+            bb = Math.max(bb, min);
+            cb = Math.max(cb, min);
+            db = Math.max(db, min);
         }
+        int as = a & 16711680;
+        int bs = b & 16711680;
+        int cs = c & 16711680;
+        int ds = d & 16711680;
+        if ((as == 0) || (bs == 0) || (cs == 0) || (ds == 0)) {
+            // Find the minimum value between all corners
+            final int min = minNonZero(minNonZero(as, bs), minNonZero(cs, ds));
+
+            // Normalize the corner values
+            as = Math.max(as, min);
+            bs = Math.max(bs, min);
+            cs = Math.max(cs, min);
+            ds = Math.max(ds, min);
+        }
+        a = ab | as;
+        b = bb | bs;
+        c = cb | cs;
+        d = db | ds;
 
         // FIX: Apply the fullbright lightmap from emissive blocks at the very end so it cannot influence
         // the minimum lightmap and produce incorrect results (for example, sculk sensors in a dark room)
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
@@ -237,10 +237,10 @@ class AoFaceData {
     private static int calculateCornerBrightness(int a, int b, int c, int d, boolean aem, boolean bem, boolean cem, boolean dem) {
         // FIX: Normalize corner vectors correctly to the minimum non-zero value between each one to prevent
         // strange issues
-        int ab = a & 255;
-        int bb = b & 255;
-        int cb = c & 255;
-        int db = d & 255;
+        int ab = a & 0xFF;
+        int bb = b & 0xFF;
+        int cb = c & 0xFF;
+        int db = d & 0xFF;
         if ((ab == 0) || (bb == 0) || (cb == 0) || (db == 0)) {
             // Find the minimum value between all corners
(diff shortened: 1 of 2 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
```diff
@@ -237,16 +237,38 @@ private static int packLight(float sl, float bl) {
...
-            a = Math.max(a, min);
-            b = Math.max(b, min);
-            c = Math.max(c, min);
-            d = Math.max(d, min);
+            ab = Math.max(ab, min);
+            bb = Math.max(bb, min);
+            cb = Math.max(cb, min);
+            db = Math.max(db, min);
         }
+        int as = a & 16711680;
```
douira: What does this magic number `16711680` do? Can we format this in a more comprehensible way?

douira (changes requested):
> split sky and block light level outside that function, so we will only have to do 9 splits instead of 16

I don't understand what this means. What are we splitting? What is a split? Where do we see this in the code?

common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
GeForceLegend: It's 0xFF0000 in [net/minecraft/util/LightCoordsUtil#L64](https://mcsrc.dev/1/26.1.2/net/minecraft/util/LightCoordsUtil#L64), I just copied it from locally remapped jar. I'll format it later.

GeForceLegend:
In `calculateCornerBrightness()`, you can see I do `& 0xFF` (block light level) and `& 0xFF0000` (sky light level) for every passed in encoded light levels. Also, we do 4 `calculateCornerBrightness()` call at the end of `initLightData()`, that's the meaning of 16 light splits: we split block and sky light level 4 time in `calculateCornerBrightness()`, and totally 4*4 in `initLightData()`.

There is only 9 raw light levels in `initLightData()`, so if we do the split in `initLightData()`, and pass splitted block light level and sky light level into `calculateCornerBrightness()`, we only have to do 9 splits (9x `& 0xFF` and 9x `& 0xFF0000`). I would guess there isn't much performance change, so it's just some suggestion.

common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
GeForceLegend: `ab` means block light level of a, and `as` means sky light level of a. It just let the `minNonZero` applied on block light and sky light separately.

douira:
I see what you mean now. I don't have a particular preference for when to do the decoding of the sky and block light. It seems fine to me as it is. Though the alternative might be slightly faster, it would also significantly bloat the parameters.

One thing that confuses me about this code is that previously we were doing `a = Math.max(a, min);` which would imply `a` is a regular number and not some sort of packed encoding. Was this performing a special trick that somehow the `max` was supposed to only effectively compare the skylight of the values unless the skylight is zero? Or is this the mistake your are correcting with this patch?

GeForceLegend:
If all of the 4 corners have some lighting, it will just be fine as they will all be themselves without change. If one of the corner has no both sky and block light level, then it will be the `minNonZero` value which will consider least sky light level first, and then least block light level (but not all 0). I'm not sure if it correct or not, and there seems nothing wrong when block light and sky light are all fading to 0, but at least my approch seems more correct ideally as it won't let the 0 sky light level inside opaque light sources reducing nearby sky light.

(This is the image of current Sodium when block and sky light levels are both fading to 0. Seems nothing wrong?)

douira:
I've looked at this in game a little bit. This seems fine to me. If there's no objections, I'll merge this.

douira:
Are these changes also applicable to 1.21.11 and/or 1.21.1?

GeForceLegend:
I am not sure since this bug is fixed in vanilla 26.1 but not 1.21.11. If Sodium does not change these functions when updating to 26.1 then I guess it's fine

douira:
In 1.21.11 vanilla doesn't have this problem but Sodium does. It also seems like vanilla is just generally much darker, though I don't think that's better.

Sodium:
sodium 1 21 11

Vanilla:
vanilla 1 21 11

GeForceLegend:
Magma block is the specail one: it does not decrease sky light level around it. Instead, it increases sky light level. It's considered as FULL_BRIGHT which means the raw light data of magma block is always full sky and full block. If you want to test the sky light level thing, you may want something like crying obsidian, or lit redstone ore.

GeForceLegend:
The reduced sky light level around common opaque light sources should get fixed in 26.1-snapshot11 with MC-275014, tho it seems like MOJANG never tracked it so MC-161150 is still open.

c4t4strofe:
Captura de tela 2026-05-26 140125
Captura de tela 2026-05-27 122845
So, there's a bug happening in the latest version of the "SEUS" shader. I've already removed all mods (except Iris, Sodium, and the Fabric API) and resource packs, but it's still bugged, giving this error. I'm on version 1.21.1.

PR #3632, opened 2026-05-06
Add numeric framerate limit input

### LINKED ISSUE #3556: Replace the framerate slider with a numeric input for custom FPS caps
### Request
Would you consider replacing Sodium's framerate limit slider with a numeric text box and allowing arbitrary positive FPS caps?

### Why this would help
- The current 10-step slider is awkward when someone wants an exact cap like 144, 165, 180, 240, 360, or 480.
- The old 260 ceiling means anything above that effectively becomes uncapped instead of staying at a real user-defined limit.
- A typed value makes more sense here than a slider because this is usually an exact-match setting, not something people tune by feel.

### Expected behavior
- The framerate option accepts any positive integer cap.
- Values above the current 260 limit are treated as real caps instead of "max" or uncapped.
- The chosen value saves through the normal options flow and stays in sync anywhere the vanilla framerate option is changed.

### Extra context
I ended up building a small addon for this behavior on my own setup, and after actually using it for a while it feels much better than the slider in practice.

I also noticed draft PR #2618 touches the same area, so this might overlap with work you've already seen.

Riqqqque:
Closes #3556.

Adds a numeric input for the framerate limit so exact caps can be entered instead of being limited to 10 FPS slider steps.

Changes:
- uses a text box for Sodium's framerate limit option
- replaces Minecraft's framerate limit option with an exact integer range before options are loaded
- keeps the existing Sodium apply/save flow

The standalone addon this was ported from is Unlicense.

Validation:
- `./gradlew.bat build`
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks`


common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/IntegerOption.java
```diff
@@ -59,6 +61,10 @@ Integer validateValue(Integer value) {
 
     @Override
     Control createControl() {
+        if (FramerateLimit.OPTION_ID.equals(this.id)) {
+            return new IntegerTextBoxControl(this);
```
douira: I don't think it's a good idea to hardcode this here. We should make adding Integer text box controls part of the api so that other mods can also use it.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/IntegerTextBoxControl.java
```diff
@@ -0,0 +1,255 @@
...
+            this.setText(String.valueOf(this.option.getValidatedValue()));
+        }
+
+        private void setText(String text) {
+            this.updatingText = true;
+            this.textBox.setValue(text);
+            this.updatingText = false;
+        }
+
+        private static String sanitize(String text) {
```
douira, line 250: I feel like sanitization could also be handled by a simple regex?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/FramerateLimit.java
```diff
@@ -0,0 +1,15 @@
+package net.caffeinemc.mods.sodium.client.gui.options;
+
+import net.minecraft.resources.Identifier;
+
+public final class FramerateLimit {
+    public static final Identifier OPTION_ID = Identifier.parse("sodium:general.framerate_limit");
```
douira: as stated above, this should probably not exist

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/gui/FramerateLimitOptionMixin.java
```diff
@@ -0,0 +1,56 @@
...
+                "options.framerateLimit",
+                OptionInstance.noTooltip(),
+                FramerateLimitOptionMixin::formatFramerateLimit,
+                new OptionInstance.IntRange(FramerateLimit.MIN, FramerateLimit.MAX),
+                Codec.intRange(FramerateLimit.MIN, FramerateLimit.MAX),
+                FramerateLimit.VANILLA_DEFAULT,
+                FramerateLimitOptionMixin::setFramerateLimit);
+    }
+
+    private static Component formatFramerateLimit(Component caption, Integer value) {
```
douira, line 46: is it necessary to have this when you're not going to be seeing the vanilla option anyway? or is this a provision for when somebody uses the shift + P bypass?

douira (changes requested):
If you don't have a suggestion for changes to the API to accomodate a second type of integer option, I can also do that myself during merge.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/FramerateLimit.java
Riqqqque: Removed this class. The constants are now local to the two FPS call sites.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/IntegerTextBoxControl.java
Riqqqque: Switched the paste cleanup to a compiled regex.

common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/IntegerOption.java
Riqqqque: Changed this to a builder option. Integer options can now call `setControl(IntegerOptionControl.TEXT_BOX)`, and the FPS option opts into it without checking the option id.

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/gui/FramerateLimitOptionMixin.java
Riqqqque: Yeah, it is still needed for the backing option, not the vanilla screen. Sodium writes through `Options.framerateLimit()`, and the vanilla value set still rounds via `/10`, so exact values like 399 become 390 without replacing it.

douira (changes requested):
I've pushed some changes to this branch to clean up some things, faster than going back and forth here.
- However, I've noticed that once I enter a number into the text field, it doesn't let me change it.
- Focusing it by keyboard navigation works, but it doesn't let me enter text
- In the video the reset button shows up sometimes, but that's just because pressing shift + tab activates it because of the shift key
- Clicking on the text field doesn't activate the cursor once it's been used
- When the text field is activated and there's an outline around the option, the outlines touch which is not nice
- The text in the textbox seems to be around 1-2 pixels too high and not vertically centered

https://github.com/user-attachments/assets/3495bb8c-c752-45d7-83f5-32053684309d

Riqqqque:
Pushed a fix for the textbox focus/click handling and tightened the textbox height/vertical positioning.

Checked with:
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks`
- `./gradlew.bat build`

Riqqqque:
Added one more follow-up for the textbox behavior.

Tab focus now routes input to the actual text field, so typing/backspace should work instead of only showing the cursor. The click focus, focus-out commit, and spacing tweaks are still in place.

Checked with:
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks`
- `./gradlew.bat build`

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/options/control/IntegerTextBoxControl.java
```diff
@@ -0,0 +1,310 @@
...
+    static class IntegerTextBoxControlElement extends StatefulControlElement implements ContainerEventHandler {
+        private static final Pattern NON_DIGIT_PATTERN = Pattern.compile("[^0-9]");
+        private static final int TEXT_BOX_WIDTH = Layout.SLIDER_WIDTH;
+        private static final int TEXT_BOX_HEIGHT = Layout.BUTTON_SHORT - 6;
+
+        private final IntegerOption option;
+        private final EditBox textBox;
+
+        private @Nullable GuiEventListener focusedElement;
+        private boolean dragging;
```
douira: can you explain why we have this `dragging` flag?

douira (changes requested):
It seems to work now but I'm somewhat confused why we need `dragging`. This flag doesn't seem to be doing anything. Maybe it would be appropriate to instead implement `AbstractParentWidget` rather than reimplementing a bunch of its code?
Thanks for improving this patch thus far!

Riqqqque:
Cleaned up the textbox focus handling.

`ControlElement` now has the parent-widget focus support, and the textbox registers its `EditBox` as a child instead of carrying its own focused-child/dragging state. The textbox control only keeps the focus-out commit hook now.

Checked with:
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks`
- `./gradlew.bat build`

CoolAtlantis:
Yea, we need this! I play with VRR and i would like to have my framerate capped exactly at 177fps. I currently use a mod which makes the slider go by steps of 1 instead of 10. But having an input box built in to sodium itself, sounds awesome.

douira (changes requested):
It appears that tab navigation between options is broken on this branch.

Riqqqque:
Fixed the tab navigation issue. Normal option controls were returning their own focus path even while already focused, so Tab kept selecting the same option. The fallback now only returns the control when it is not already focused; controls with children still use parent traversal.

Checked with:
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks --console=plain`
- `./gradlew.bat build --console=plain`

Commit: `1db98b99`

Riqqqque:
﻿Rebased onto current `dev`. Tab navigation fix is still in.

Checked with:
- `./gradlew.bat :common:compileJava :fabric:compileJava :neoforge:compileJava --rerun-tasks --console=plain`
- `./gradlew.bat build --console=plain`

PR #3635, opened 2026-05-07
Sort Entity Quads by Closest Point

### LINKED ISSUE #3343: Stained glass block displays rendered aint rendered inside each other
### Bug Description

This bug even is in vanilla (i tested it in vanilla, my default client side fabric modpack, fabric+sodium only) and happens with translucent stained blocks for block displays... E.g. stained glass but also stuff like nether portal
Other blocks work the as expected in block displays

Nothing is changed between these screenshots except player location:
Image
Image
Image
Image

### Reproduction Steps

Summon different sized block displays e.g. of stained glass and then of a normal block inside each other

### Log File

[latest.log](https://github.com/user-attachments/files/23637691/latest.log)

### Crash Report

[crash-2025-11-19_22.25.09-client.txt](https://github.com/user-attachments/files/23637715/crash-2025-11-19_22.25.09-client.txt)

douira:
Adds a default-off option to sort entity quads by their closest point to the camera. This improves the sorting order for entities that contain many differently sized quads arranged in specific unfortunate ways.

Fixes #3343


douira:
Does anyone have a suggestion for a better place to check the current state of the option? I'm not sure if there is maybe a performance impact of checking it every time we need to sort something, which may be many times per frame.

PR #3640, opened 2026-05-09
Fixed FULL_BRIGHT blocks increasing nearby sky light level

GeForceLegend:
Mostly magma and active sculk sensor, I'm not sure if there is any FULL_BRIGHT blocks in vanilla. They will consider themselves have full sky and block brightness, and then increasing nearby sky light level even inside dark caves if smooth lighting is on. This will result in lighting around them brighter during day, and darker during night.

Day:

Night:

After only apply full block light level on them, they no longer contribute sky light level to nearby blocks, and keeps the same lighting no matter what daytime it is. (And more yellow to be block light color instead of white sky light color right?)

Quads of those block themselves still have full sky light strength, which might similar to `light_emission`. Vanilla `light_emission` also increase the minumum value of sky light, but I'm not sure if injecting into [net/minecraft/util/LightCoordsUtil#L51](https://mcsrc.dev/1/26.1.2/net/minecraft/util/LightCoordsUtil#L51) is a wise thing so it does not included in this pr.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/model/light/smooth/AoFaceData.java
@@ -273,16 +273,20 @@ class AoFaceData {
         // FIX: Apply the fullbright lightmap from emissive blocks at the very end so it cannot influence
         // the minimum lightmap and produce incorrect results (for example, sculk sensors in a dark room)
         if (aem) {
-            a = LightCoordsUtil.FULL_BRIGHT;
+            a &= 0xFF0000;
+            a |= 0xF0;
         }
         if (bem) {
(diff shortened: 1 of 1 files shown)


douira:
Is this the fix for https://github.com/CaffeineMC/sodium/issues/3639 ?

GeForceLegend:
Nope. I'm not sure how sodium handles things related to 3639 so I haven't decide how to fix it, this is just some further fix for wrongly spreaded vanilla sky light level.

Lolothepro:
is this fix https://github.com/CaffeineMC/sodium/issues/2442 ?

GeForceLegend:
If yor issue is related to AO, it should be fixed in #3631. This pr will actually make magma blocks darker than vanilla when sky light level is low, as before this pr, magma block will be considered as full sky light, and increase nearby sky light level if smooth lighting is on.

PR #3644, opened 2026-05-10
Fix Light Emission Type Error

### LINKED ISSUE #3639: Incorrect `light_emission` handle for item models
### Bug Description

In [net/caffeinemc/mods/sodium/client/render/immediate/model/BakedModelEncoder.java#L33](https://github.com/CaffeineMC/sodium/blob/dev/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/BakedModelEncoder.java#L33), the second param of `getLightCoordsWithEmission()` requires vanilla `light_emission` which would be in range [0, 15]. But Sodium passes an encoded light coordinate for it, which would be at least 0x100010 (1048592) if `light_emission` of this quad is not zero. This will lead to any non-zero light_emission value becomes full bright on item models.

### Reproduction Steps

Create any item model with only 1 light emission, and see it's as bright as 15.

### Log File

Nothing useful in log

### Crash Report

No crash

douira:
Fix type error where instead of the quad's light emission a packed coordinate is passed into lightCoordsWithEmission.

This is a draft PR because before this is merged I would like to document this code so that it becomes somewhat more comprehensible. Currently this is completely opaque and impossible to properly work with without prior knowledge.

@IMS212 @jellysquid3 or anyone else who happens to know about this, please supply a description of the following:

1. What is `instance` and where does it come from?
2. What is `quad` and where does it come from?
3. In which contexts and why is `BakedModelEncoder.writeQuadVertices` called?
4. What was `int newLight = instance.getLightCoordsWithEmission(i, quad.getMaxLightQuad(i));` supposed to be doing and how is it different from doing `int newLight = quad.getMaxLightQuad(i);`?

If you feel we can solve this issue in another way than adding a method to BakedQuadView, such as adding a `lightEmission`parameter to `writeQuadVertices` [like this](https://discord.com/channels/602796788608401408/651120262129123330/1502841918466424985).

Closes https://github.com/CaffeineMC/sodium/issues/3639

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/model/quad/BakedQuadView.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/model/quad/BakedQuadView.java
@@ -10,4 +10,6 @@ public interface BakedQuadView extends ModelQuadView {
     boolean hasShade();
 
     boolean hasAO();
+
+    int getLightEmission();
 }
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/BakedModelEncoder.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/BakedModelEncoder.java
@@ -16,6 +16,14 @@ import org.lwjgl.system.MemoryStack;
 public class BakedModelEncoder {
     private static final boolean USE_COLOR_MULTIPLICATION = PlatformRuntimeInformation.getInstance().usesBakedQuadColorMultiplication();
 
+    /**
+     * Encodes the given quad into the provided writer, applying the transformations and combining the data from {@code quad} and {@code instance}, where {@code instance} is modified dynamically and {@code quad} comes from the baked model.
+     *
+     * @param writer The writer to write the vertex data into.
+     * @param matrices The current transformation matrices to apply to the vertex data.
+     * @param quad The quad to encode, providing the base vertex data.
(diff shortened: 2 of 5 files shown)


GeForceLegend:
1. Seems to be `private final QuadInstance quadInstance;` of `net.minecraft.client.renderer.feature.BlockFeatureRenderer` or `ItemFeatureRenderer` in same package. I guess it's the instance of current item or block?

2,3. IDEA shows it's only called in [common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_builder/intrinsics/BufferBuilderMixin.java#L67](https://github.com/CaffeineMC/sodium/blob/dev/common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_builder/intrinsics/BufferBuilderMixin.java#L67). It passed a `BakedQuad` into `BakedModelEncoder.writeQuadVertices`. Those quads should be quads of related item and block models.

4. `quad.getMaxLightQuad(i)` will just be the light_emission encoded to both sky and block, because the [`getLight()`](https://github.com/CaffeineMC/sodium/blob/dev/fabric/src/main/java/net/caffeinemc/mods/sodium/mixin/core/model/quad/BakedQuadMixin.java#L78) in [`getMaxLightQuad()`](https://github.com/CaffeineMC/sodium/blob/dev/fabric/src/main/java/net/caffeinemc/mods/sodium/mixin/core/model/quad/BakedQuadMixin.java#L123) always return 0 for BakedQuad, so it can just be correctly replaced with simple `light_emission` values.

PR #3648, opened 2026-05-13
Refactor `ImprovedItemModelBuilder` for better mod compatibility

pajicadvance:
Instead of replacing the whole `ItemModelGenerator` class with an instance of `ImprovedItemModelBuilder`, this changes `ImprovedItemModelBuilder` to be an utility class and replaces only the relevant method calls in `ItemModelGenerator` with methods from `ImprovedItemModelBuilder`. This keeps the improvements from https://github.com/CaffeineMC/sodium/pull/3551 while keeping compatibility with [Glowtone](https://modrinth.com/mod/glowtone) and possibly other mods.


douira:
I'll give this a review later. Could you also do a backport to 1.21.11? The original PR that added this code was also produced for 1.21.11 but I recall there was some effort needed to adapt to Minecraft changes.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
```diff
@@ -147,12 +96,8 @@ private static void bakeSideQuads(
 			var toY = minY;
 
 			switch (faceFacing) {
-				case UP -> {
-					toX = minX + length;
-				}
-				case LEFT -> {
-					toY = minY + length;
-				}
+				case UP -> toX = minX + length;
```
douira: nit: this part of the diff is just formatting, I think it was good the way it was.

pajicadvance:
Yes I'll get to it shortly, the mixin targets will likely need to be adjusted for 1.21.11.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
```diff
@@ -227,22 +172,6 @@ private static Collection<SideFace> buildSideFaces(SpriteContents sprite) {
 		return storage.buildSideFaces();
 	}
 
-	private record ItemLayerKey(BakedQuad.MaterialInfo quadMaterial, ModelState modelState) implements ModelBaker.SharedOperationKey<@NotNull QuadCollection> {
```
douira, line 230: was item layer key not necessary in the first place or did these changes obsolete it?

douira (changes requested):
looks good to me! I haven't tested it yet. Do you happen to have any ideas what we could do about armor trim overlays having z-fighting?

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
pajicadvance: The original PR replaced the entire record only to change one method call in the `compute` method. This is now done via a redirect in the mixin: https://github.com/CaffeineMC/sodium/pull/3648/changes#diff-4ba4cf616c2ef496dade3318a278473c2217ec11ebc34de5715aabc44e67c0a9R31

douira (approved):

common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilder.java
pajicadvance: OK, to explain this better:
- The original code replaced the `ItemModelGenerator.bakeExtrudedSprite` method with `bakeItemQuads`. The method is identical, except it calls `bakeSideQuads` instead of `ItemModelGenerator.bakeSideFaces`.
- This method is called in `ItemLayerKey.compute`. The `compute` method was replaced only to change the `ItemModelGenerator.bakeExtrudedSprite` call to a `bakeItemQuads` call. Instead of doing this, I just mixin into `ItemModelGenerator.bakeExtrudedSprite` to replace the `ItemModelGenerator.bakeSideFaces` call with `bakeSideQuads`, achieving the same functionality.

PR #3697, opened 2026-05-31
Fix Double Driver Workaround Application

### LINKED ISSUE #3664: Driver workarounds are applied twice when FML early window control is disabled
### Bug Description

On NeoForge, when early window control is disabled, driver workarounds are applied twice, potentially causing error.

The first, in `GraphicsBootstrapper`:

https://github.com/CaffeineMC/sodium/blob/59e7586b49da6ba278912d03e8d83eef7a8b07d5/neoforge/src/main/java/net/caffeinemc/mods/sodium/service/SodiumWorkarounds.java#L23-L24

The second, in `WindowMixin`:

https://github.com/CaffeineMC/sodium/blob/59e7586b49da6ba278912d03e8d83eef7a8b07d5/common/src/main/java/net/caffeinemc/mods/sodium/mixin/workarounds/context_creation/WindowMixin.java#L40-L41

The root cause seems to be a change in NeoForge. On 1.21.1 the `glfwCreateWindow` call is [removed](https://github.com/neoforged/NeoForge/blob/869777fa2e6fc91b55a406e4b9733cb1dca92402/patches/com/mojang/blaze3d/platform/Window.java.patch#L7-L9), so the `WindowMixin` won't inject. On 1.21.11+, however, the `glfwCreateWindow` call [remains](https://github.com/neoforged/NeoForge/blob/6cb49467014c294e6cd6adae618d516a27c17cb5/patches/com/mojang/blaze3d/platform/Window.java.patch#L11) in place.

Note: This had been reported in https://github.com/CaffeineMC/sodium/issues/3636 but it was wrongly attributed to Ixeris or Async Logger.

### Reproduction Steps

1. Create a 1.21.11+ NeoForge instance and launch, for `fml.toml` to be created
2. In `fml.toml`, set `earlyWindowControl = false`
3. Install Sodium and launch the game
4. Observe that the driver workarounds are applied twice (the game might crash with message "Command line is already modified")

### Log File

Not needed (something similiar to "Modifying process environment to apply workarounds for the AMD graphics driver..." would be printed twice)

### Crash Report

See #3636

douira:
Only make environment changes if the early window will create a gl context early

Closes #3664

### DIFF
--- a/neoforge/src/main/java/net/caffeinemc/mods/sodium/service/SodiumWorkarounds.java
+++ b/neoforge/src/main/java/net/caffeinemc/mods/sodium/service/SodiumWorkarounds.java
@@ -3,8 +3,9 @@ package net.caffeinemc.mods.sodium.service;
 import net.caffeinemc.mods.sodium.client.compatibility.checks.PreLaunchChecks;
 import net.caffeinemc.mods.sodium.client.compatibility.environment.probe.GraphicsAdapterProbe;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.Workarounds;
-import net.caffeinemc.mods.sodium.client.compatibility.workarounds.nvidia.NvidiaWorkarounds;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.amd.AmdWorkarounds;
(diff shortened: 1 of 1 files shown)


decce6:
I can confirm this fixes the issue.

PR #3701, opened 2026-06-02
fix: respect ambient occlusion flag in FRAPI AltModelBlockRenderer

### LINKED ISSUE #3594: Ambient occlusion flag not respected in AltModelBlockRenderer implementation.
### Bug Description

The AltModelBlockRenderer that Sodium returns from `Renderer.get().altModelBlockRenderer()` seems to not respect the `ambientOcclusion` parameter passed in, and unconditionally emits quads which are shaded by cardinal direction, but lack any form of ambient occlusion baked into the vertex colors. This behavior differs from Indigo's AltModelBlockRenderer implementation, which bakes the AO into the vertex colors the way I would expect it to.

Screenshot when using Indigo:
Image

Screenshot when using Sodium:
Image

This issue affects `sodium-fabric-0.8.9+mc26.1.1.jar`, and possibly earlier versions that I didn't test.

### Reproduction Steps

The above screenshots were from my LOD terrain renderer, but a more minimal stand-alone class which demonstrates this issue can be found here: [AOTest.java](https://github.com/user-attachments/files/26665902/AOTest.java)

Upon calling `capture()`, it will create a Mesh containing the geometry of the blocks in an 11x11x11 block area centered around the player, and start rendering it 20 blocks above the player's location at the time of capturing. The mesh can then be inspected visually to see whether or not it has the correct AO everywhere.

### Log File

[latest.log](https://github.com/user-attachments/files/26666037/latest.log)

### Crash Report

[crash-2026-04-12_23.59.07-client.txt](https://github.com/user-attachments/files/26666040/crash-2026-04-12_23.59.07-client.txt)

nghiphaam:
The AltModelBlockRenderer implementation in FRAPI (NonTerrainBlockRenderContext) was missing ambient occlusion initialization. The prepareAoInfo() method that sets useAmbientOcclusion and defaultLightMode fields inherited from AbstractBlockRenderContext was never called, so both fields remained at their defaults (false and FLAT), causing all quads to be rendered with flat shading regardless of the ambientOcclusion parameter passed to the constructor.

Closes #3594

### DIFF
--- a/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
+++ b/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
@@ -154,6 +154,9 @@ public class NonTerrainBlockRenderContext extends AbstractBlockRenderContext imp
         this.offset.set(x + offset.x, y + offset.y, z + offset.z);
         defaultAo = allowAO && blockState.getLightEmission() == 0;
 
+        this.useAmbientOcclusion = allowAO;
+        this.defaultLightMode = useAmbientOcclusion && defaultAo ? LightMode.SMOOTH : LightMode.FLAT;
+
         this.lightDataCache.reset(pos, level);
         this.prepareCulling(enableCulling);
 
@@ -178,7 +181,7 @@ public class NonTerrainBlockRenderContext extends AbstractBlockRenderContext imp
         if (aoMode == TriState.DEFAULT) {
             lightMode = this.defaultLightMode;
         } else {
-            lightMode = this.useAmbientOcclusion && aoMode != TriState.FALSE && defaultAo ? LightMode.SMOOTH : LightMode.FLAT;
+            lightMode = this.useAmbientOcclusion && aoMode != TriState.FALSE ? LightMode.SMOOTH : LightMode.FLAT;
         }
         final boolean emissive = quad.emissive();
 


frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
douira: We already basically have the same calculation as this PR adds here. Should this one be removed or adjusted? It seems weird to me what we have this once per block and once per quad. Could you maybe point me to and compare to the reference implementation in Indigo of this if one exists?

douira (changes requested):

frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
nghiphaam: Right — removed the redundant `&& defaultAo`. It was compensating for the missing `useAmbientOcclusion` init. Now `processQuad` matches the terrain `BlockRenderer`. The two new lines in `tesselateBlock` do what `prepareAoInfo()` does for the terrain path, using `allowAO` from the constructor since FRAPI has no `LevelSlice`.

jellysquid3:
AI-generated contributions are not accepted to this project, per the contributing guidelines. Even more so if you are going to use it to dishonestly engage with others in review feedback.

PR #3735, opened 2026-06-14
Port to Minecraft 26.2

IMS212:
This includes a major rewrite to the renderer to run entirely within Blaze3D, Mojang's rendering API. This provides support for both OpenGL and Vulkan by default.

A new `DrawContext` system has been introduced for the part of the renderer that requires API specific knowledge. There are three backends.
- GLDrawContext: Replicates pre-PR behavior as-is.
- VKMultiDrawContext: The preferred path on Vulkan. Uses push constants as a replacement for uniforms, and `EXT_multi_draw` as the multidraw command.
- VKIndirectContext: Since `EXT_multi_draw` does not have guaranteed support, this turns the draw into an indirect one with an extra GPU upload, which is always supported in this Vulkan version.


common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/arena/staging/MappedStagingBuffer.java
```diff
@@ -135,16 +129,15 @@ private static List<CopyCommand> consolidateCopies(PriorityQueue<CopyCommand> qu
...
 
     @Override
-    public void delete(CommandList commandList) {
+    public void delete() {
         while (!this.fencedRegions.isEmpty()) {
             var region = this.fencedRegions.dequeue();
             var fence = region.fence();
-            fence.sync();
-            fence.delete();
+            fence.awaitCompletion(1000);
```
douira, line 136: Is there a cleaner way to wait on this fence?

common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/VKMultiDrawBatch.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public void put(int size, int elementCount, int baseVertex, long elementOffset) {
+        MemoryIntrinsics.putInt(pCommands + (size * VkMultiDrawIndexedInfoEXT.SIZEOF) + VkMultiDrawIndexedInfoEXT.INDEXCOUNT, elementCount);
+        MemoryIntrinsics.putInt(pCommands + (size * VkMultiDrawIndexedInfoEXT.SIZEOF) + VkMultiDrawIndexedInfoEXT.VERTEXOFFSET, UInt32.uncheckedDowncast(baseVertex));
+        MemoryIntrinsics.putInt(pCommands + (size * VkMultiDrawIndexedInfoEXT.SIZEOF) + VkMultiDrawIndexedInfoEXT.FIRSTINDEX, UInt32.uncheckedDowncast(elementOffset));
+    }
+
+    @Override
+    public void draw(DrawContext context) {
+        context.getPass().multiDrawIndexed(MemoryUtil.memIntBuffer(pCommands, size * 3), 1, 0, size);
```
douira: Where does this 3 come from?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -720,24 +737,11 @@ private OptionPageBuilder buildAdvancedPage(ConfigBuilder builder) {
                                 .setTooltip(Component.translatable("sodium.options.use_persistent_mapping.tooltip"))
                                 .setDefaultValue(DEFAULTS.advanced.useAdvancedStagingBuffers)
                                 .setBinding(value -> this.sodiumOpts.advanced.useAdvancedStagingBuffers = value, () -> this.sodiumOpts.advanced.useAdvancedStagingBuffers)
-                                .setEnabled(isPersistentMappingSupported)
+                                .setEnabled(false)
```
douira: If this option is now dead code, we should probably remove it.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/region/RenderRegion.java
```diff
@@ -187,22 +186,13 @@ public SectionRenderDataStorage createStorage(TerrainRenderPass pass) {
         return storage;
     }
 
-    public void refreshTesselation(CommandList commandList) {
-        if (this.resources != null) {
-            this.resources.deleteTessellation(commandList);
-            this.resources.deleteIndexedTessellation(commandList);
-        }
-
+    public void refreshTesselation() {
```
douira: Does the concept of a tesselation even still exist? If not, this should be renamed.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
douira, line 1: As noted before, this class shouldn't be concerned with the management of the gpu buffers or the chunk fades. Further, it should probably also not deal with rendering at all. Please at least move some of this out.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/SharedQuadIndexBuffer.java
```diff
@@ -17,64 +13,66 @@ public class SharedQuadIndexBuffer {
...
-    public void delete(CommandList commandList) {
-        commandList.deleteBuffer(this.buffer);
+    public void delete() {
+        if (this.buffer != null) this.buffer.close();
     }
 
-    public enum IndexType {
-        SHORT(GlIndexType.UNSIGNED_SHORT, 64 * 1024) {
+    public enum IndexFormat {
+        SHORT(IndexType.SHORT, 64 * 1024) {
```
douira: SHORT is unused. Why do we still have this?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/immediate/buffer_builder/sorting/MultiBufferSourceMixin.java
douira, line 1: If we don't need this anymore and you have good evidence of this, please remove it.

neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/block/NeoForgeBlockAccess.java
```diff
@@ -99,7 +99,7 @@ public AmbientOcclusionMode usesAmbientOcclusion(BlockStateModelPart model, Bloc
 
     @Override
     public boolean shouldBlockEntityGlow(BlockEntity blockEntity, LocalPlayer player) {
-        return blockEntity.hasCustomOutlineRendering(player);
+        return false;// TODO?- blockEntity.hasCustomOutlineRendering(player);
```
douira, line 102: What is up with this?

douira (changes requested):
I would like to avoid RSM becoming more of a mess than it already is. Surely, moving the new code from it to somewhere else more appropriate wouldn't require rewriting everything.

common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/VKMultiDrawBatch.java
IMS212: The struct for a multi draw command is 12 bytes. 12 / 4 = 3

can probably be made a constant, but it is impossible this will ever change

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
IMS212: I do want to do that; but it seems like doing that removes the only setting in the "early load", so I have to remove the Sodium group entirely, which breaks something further in the config API?

neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/block/NeoForgeBlockAccess.java
IMS212: No replacement for this exists, at least in the current Neo alpha.

common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/VKMultiDrawBatch.java
douira: Maybe this could be written down somewhere.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
douira: The only early load option we have is buildNoErrorContextOption as far as I can tell.

neoforge/src/mod/java/net/caffeinemc/mods/sodium/neoforge/block/NeoForgeBlockAccess.java
douira: I guess we should get back to this once neo exits alpha? We should make better note of this than just an unlabeled todo that I guarantee will be forgotten about.

Lolothepro:
Is it possible to disable OpenGL workarounds with Vulkan and log the selected DrawContext?

PR #3752, opened 2026-06-22
Fix rendering regions with deleted resources

### LINKED ISSUE #3744: NPE when rendering with deleted region resources
### Bug Description

so i walked around in the game for around 30 minutes, died, respawned. and got a crash after i pressed the restart button, might be gpu problem though (gt 1030), i played with 20 chunks render distance. Sodium 0.9.0, mc 26.2

### Reproduction Steps

play the game for 30 minutes, die, respawn

### Log File

[2026-06-19-1.log](https://github.com/user-attachments/files/29145313/2026-06-19-1.log)

### Crash Report

[crash-2026-06-19_22.05.28-client.txt](https://github.com/user-attachments/files/29145315/crash-2026-06-19_22.05.28-client.txt)

FlashyReese:
Fixes #3744.

I was able to reproduce this with a command-block stress test world that repeatedly rebuilds translucent glass pane sections while trying to replicate another issue.

Steps:
1. [Load the attached test world.](https://github.com/user-attachments/files/29191250/2026-06-21_23-25-29_3744.zip)
2. Stand near the test area: `/tp @p 16 82 16`
3. Teleport away from it: `/tp @p 500 90 500`
4. The game crashes during the render frame.

The crash happens because a render region can have its GPU resources deleted while an old cached draw batch still makes the renderer try to draw it.

This clears cached batches when region resources are deleted, and skips rendering regions that no longer have resources.

Edit:

I managed to reproduce another crash while retesting this world by teleporting back and forth quickly. I'm not sure about the fix for that one yet, but it looks like a separate async culling race. Seems timing-sensitive, only managed to reproduce once.

Crash log: [crash-2026-06-21_23.35.58-client.txt](https://github.com/user-attachments/files/29191814/crash-2026-06-21_23.35.58-client.txt)

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/DefaultChunkRenderer.java
@@ -82,6 +82,12 @@ public class DefaultChunkRenderer extends ShaderChunkRenderer {
                 continue;
             }
 
+            var resources = region.getResources();
+            if (resources == null) {
+                region.clearCachedBatchFor(renderPass);
+                continue;
+            }
+
             var batch = region.getCachedBatch(renderPass);
             if (!batch.isFilled) {
                 fillCommandBuffer(batch, region, storage, renderList, camera, renderPass, useBlockFaceCulling, useIndexedTessellation);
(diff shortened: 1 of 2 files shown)


douira:
Thanks for the contributions! I believe the other crash is caused by a lack of synchronization on the section map, I'll try to fix that.

PR #3756, opened 2026-06-23
Fix crashes resulting from unsafe concurrency in async culling

douira:
I don't think we have a crashlog for this as a github issue, but it's a arrayindexoutofbounds exception within the occlusion culler when it tries to read the section map but it gets modified during the hashmap probing and breaks.


FlashyReese:
I got a crashlog while working on https://github.com/CaffeineMC/sodium/pull/3752

douira:
Can you share the crashlog?

FlashyReese:
https://github.com/user-attachments/files/29191814/crash-2026-06-21_23.35.58-client.txt

douira:
Thanks. And I assume it's now fixed?

PR #3773, opened 2026-06-26
 Cache max draw size in MultiDrawBatch instead of scanning every frame

ExodusCoder9:
getIndexBufferSize() was doing a full O(n) walk through native memory every frame for every non-empty region — just to check if the shared index buffer is big enough ,  on most frames the buffer is already allocated at the right size, so all those reads are wasted.
My Pr moves the tracking into put() so the max element count is known the moment the batch is built. getIndexBufferSize() now returns the cached value in constant time.
All three backends (GL, VK multi-draw, VK indirect) had the same pattern — removed the scan from each and added the cache update call.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
@@ -1,17 +1,11 @@
 package net.caffeinemc.mods.sodium.client.gpu.device.batch;
 
-import com.mojang.blaze3d.buffers.GpuBufferSlice;
-import com.mojang.blaze3d.systems.RenderPass;
 import net.caffeinemc.mods.sodium.client.gpu.device.backend.DrawBackend;
 import net.caffeinemc.mods.sodium.client.gpu.device.context.DrawContext;
 
 public abstract class MultiDrawBatch {
     public int size;
     public boolean isFilled;
-
-    // Tracks the largest element count across all draw commands in this batch.
-    // Updated during put() so getIndexBufferSize() can return it instantly
-    // instead of scanning native memory every frame.
     private int maxElementCount;
 
     public static MultiDrawBatch newBatch(int capacity) {
@@ -32,21 +26,12 @@ public abstract class MultiDrawBatch {
         return this.size <= 0;
     }
 
-
-     //Returns the number of elements in the largest draw within this batch.
-     //Used by the caller to ensure the shared index buffer is big enough.
-    public int getIndexBufferSize() {
+    public int getMaxElementCount() {
         return this.maxElementCount;
     }
 
-
-     //Called by subclasses when appending a draw command. Keeps the cached
-    // maximum up to date so we don't have to scan the full batch later.
-
     protected final void updateMaxElementCount(int elementCount) {
-        if (elementCount > this.maxElementCount) {
-            this.maxElementCount = elementCount;
-        }
+        this.maxElementCount = Math.max(this.maxElementCount, elementCount);
     }
 
     public abstract void put(int size, int elementCount, int baseVertex, long elementOffset);
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
@@ -9,6 +9,11 @@ public abstract class MultiDrawBatch {
     public int size;
     public boolean isFilled;
 
+    // Tracks the largest element count across all draw commands in this batch.
+    // Updated during put() so getIndexBufferSize() can return it instantly
+    // instead of scanning native memory every frame.
+    private int maxElementCount;
+
     public static MultiDrawBatch newBatch(int capacity) {
         return switch (DrawBackend.BACKEND) {
             case OPENGL -> new GLDrawBatch(capacity);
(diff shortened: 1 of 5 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
```diff
@@ -498,15 +497,26 @@ public void processChunkBuilds(Viewport viewport, UniformBufferManager uniforms)
...
+        }
+
+        int camChunkX = (int) this.cameraPosition.x() >> 4;
+        int camChunkZ = (int) this.cameraPosition.z() >> 4;
+
+        if (Math.abs(section.getChunkX() - camChunkX) > 5 || Math.abs(section.getChunkZ() - camChunkZ) > 5) {
+            return false;
+        }
+
+        double distanceSquared = section.getSquaredDistance(
```
douira: Why are you adding another distance check? What is the purpose of having two different distance checks?

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
```diff
@@ -498,15 +497,26 @@ public void processChunkBuilds(Viewport viewport, UniformBufferManager uniforms)
     }
 
     private boolean isSectionFrustumVisible(Viewport viewport, RenderSection section) {
-        // unloaded sections are considered visible as to not be an impossible requirement for immediate presentation
-        return section == null || this.renderTree == null || this.renderTree.isSectionVisible(viewport, section);
+        return section == null || this.renderTree.isSectionVisible(viewport, section);
```
douira: Why is this change necessary?

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
```diff
@@ -483,7 +483,6 @@ public void processChunkBuilds(Viewport viewport, UniformBufferManager uniforms)
             return;
         }
 
-        // processing build results can cause invalidation of the render lists or change the connectivity of the graph. They don't necessarily imply each other, so they're tracked separately.
```
douira, line 486: Why did you remove this?

common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
```diff
@@ -20,13 +25,29 @@ public static MultiDrawBatch newBatch(int capacity) {
...
 
     public boolean isEmpty() {
         return this.size <= 0;
     }
 
-    public abstract int getIndexBufferSize();
+
+     //Returns the number of elements in the largest draw within this batch.
+     //Used by the caller to ensure the shared index buffer is big enough.
+    public int getIndexBufferSize() {
```
douira: Can you justify this change based on the places where we use the getIndexBufferSize method?

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
ExodusCoder9: Sorry , i accidentally removed it , i added it back in the next commit

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
ExodusCoder9: The first check is using Manhattan distance with integer chunk coordinates. it basically acts as an early fast-fail. Because it is placed before getSquaredDistance(), if a chunk is obviously too far away on the grid, it returns false instantly and completely skips the heavier floating-point math inside getSquaredDistance(). I added it to save a few CPU cycles on distant chunks, but if you feel getSquaredDistance() is already cheap enough that this just adds unwanted code or something, I'm okay with reverting this part! Anyway thanks for reviewing My PR.

common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/RenderSectionManager.java
ExodusCoder9: isSectionImmediatePresentationCandidate calls isSectionFrustumVisible up to 7 times (once for the section, and 6 times for its neighbors). By hoisting the renderTree == null check to the very top, we evaluate it exactly once and can return early. Before, it was redundantly evaluating that same null check 7 times in a row for every candidate chunk. So i thought we can avoid that by making this change.

common/src/main/java/net/caffeinemc/mods/sodium/client/gpu/device/batch/MultiDrawBatch.java
ExodusCoder9: Right Now, getIndexBufferSize() performs an O(N) scan through the native memory batch every single frame just to find the max element count. By tracking maxElementCount incrementally on the Java side during put(), getIndexBufferSize() becomes an O(1) operation. This completely gets rid of the need to scan native memory later.  Only Purpose of my change is to reduce memory scanning thats honestly unneeded. Thanks for the question tho.

douira: I'm willing to accept this change, but it needs to be put in a separate PR from the other change. You've made two PRs, and they both include the same two commits. Please separate them.

ExodusCoder9: Okay , Sorry for the delay but i have separated the two changes into two different PRs , MultiDrawBatch Change in This PR and RenderSectionManager Change in the other. Thanks for your patience.

Issue #3197, opened 2025-07-21
Water is invisible in some cases with FramedBlocks


Lolothepro:
### Bug Description

<img width="1848" height="1011" alt="Image" src="https://github.com/user-attachments/assets/846d8429-f40d-49ea-ae5a-1fb54083f505" />

Water is invisible in some cases with FramedBlocks
Here, the top texture should be rendered

### Reproduction Steps

Load this world:

[Water.zip](https://github.com/user-attachments/files/21354311/Water.zip)

### Log File

https://mclo.gs/PAsQER5

### Crash Report

*

Lolothepro:
Affects Minecraft 26.1

Auth0x78:
<img width="1911" height="1048" alt="Image" src="https://github.com/user-attachments/assets/92af92f7-1792-4c96-8fd2-89b9fe5b8ff5" />

_Fig: Left block is with property solid = true and the block on the right has property solid set to false_


Setting the block property solid to false make its render the top face of water.

Auth0x78:
Adding a additional condition (check if current block is solid) check inside the function `sFluidSelfVisible(BlockState selfBlockState, Direction facing, VoxelShape fluidShape)` seems to fix this issue

Fix:

<img width="877" height="242" alt="Image" src="https://github.com/user-attachments/assets/d133e399-ae00-4642-9b60-f6863bb67935" />

I wanted clarification regarding the fix, will this fix potentially cause any performance hit?

douira:
I don't think that patch is correct. Logically, if the block is solid, why would that mean we should render the fluid inside of it?

What seems to rather be the issue here is that we only call `isFluidSelfVisible` with `Shapes.block()` which would result in the occlusion being true, but the fluid shape is not actually a full block. The logic needs to be refined here to correctly handle blocks where there's a difference in visibility between culling against a full block fluid and an accurate fluid shape.

Auth0x78:
I agree that solid block should not render fluid inside them, then the problem seems to be that, the fluid block is a full block which causes the up face to not reneder and which is causing the issue.

Auth0x78:
I have found a better way that is to use `sampleFluidHeight()` to get the actual height of the fluid block and then inside the `isFluidSelfVisible()` function I have modified the return to return false if fluidHeight is 1.0f or greater and for heights less than 1.0f and if full block and full fluid block then we return true (visible fluid face) this check is only done for fluid face that is in the UP direction

Does this approach seem fair?

[...]

PR #3774, opened 2026-06-26
Fix rendering issue for UP face of waterlogged blocks

Auth0x78:
We check for if the actual fluid's height (waterlogged block) is less than 1.0f and if that is the case then the UP face should be occluded as there may occur the case where the sides of a waterlog block are translucent but UP face is solid.

Issue resolved:
#3197


Auth0x78:
Closed due to generic branch name

PR #3783, opened 2026-06-28
Parallel draw command buffer filling

ExodusCoder9:
Basically , this parallelizes draw command buffer filling , mostly did it to shift some work off the main thread , used a forkjoinpool , i profiled in jfr and it shaved off almost 0.7ms off of main thread cpu time , worth it for me for me personally as its a small change. set cores max as 4 because more would probably give vanishing returns because context switching might then take more time than the task does now. also made the hash map concurrent just to avoid any reading of stale data if new code needs it.


douira:
I'm not in favor of introducing threading into this highly latency sensitive part of the render loop.

ExodusCoder9:
okay understand your concern

douira:
If you put your suggestions in our dev channel before you put effort into writing PRs we can tell you whether they would get accepted beforehand. I applaud the initiative though.

ExodusCoder9:
okay ,will do from now on thanks for the feedback.

PR #3795, opened 2026-07-01
Fix ambient occlusion in non-terrain block rendering fixes #3780

### LINKED ISSUE #3780: moving blocksare very dark in some situations
### Bug Description

when a falling/moving block goes through a 1 block gap, the side become pitch black compared to vanilla where they just darken a bit

sodium:

Image

https://github.com/user-attachments/assets/3e2e9fbd-4d02-4d17-ae04-eea34c00ab29

vanilla:

Image

https://github.com/user-attachments/assets/6b952c84-e5ad-4829-82c6-2461ada840bb

looking at renderdoc, this seems to be an issue with the light coords, with sodium they are only max or min, with vanilla they also have values in between

sodium:

Image

vanilla:

Image

tested on both sodium 26.2-0.9.1 and sodium 26.1-0.8.12

### Reproduction Steps

let a sand block fall through a one block gap
for better visibility one can turn down the tick rate to 2

### Log File

[latest.log](https://github.com/user-attachments/files/29408373/latest.log)

### Crash Report

[crash-2026-06-27_14.18.00-client.txt](https://github.com/user-attachments/files/29408376/crash-2026-06-27_14.18.00-client.txt)

meta-legend:
Fixes #3780

The Fabric non-terrain block renderer was receiving the ambient occlusion setting through the constructor but wasn't applying it to the block's lighting state. Because of that, moving/non-terrain blocks could use flat lighting when they were supposed to use the smooth AO lighting, making them render too dark.

I can't reproduce the lighting issue with the setup from #3780 with the new fix, so it seems good.

### DIFF
--- a/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
+++ b/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
@@ -146,7 +146,9 @@ public class NonTerrainBlockRenderContext extends AbstractBlockRenderContext imp
 
         Vec3 offset = blockState.getOffset(pos);
         this.offset.set(x + offset.x, y + offset.y, z + offset.z);
+        this.useAmbientOcclusion = this.allowAO;
         this.defaultAo = this.allowAO && blockState.getLightEmission() == 0;
+        this.defaultLightMode = this.defaultAo ? LightMode.SMOOTH : LightMode.FLAT;
 
         this.lightDataCache.reset(pos, level);
         this.prepareCulling(this.enableCulling);


douira:
Does this also fix https://github.com/CaffeineMC/sodium/issues/3594? I made [a patch](https://github.com/douira/sodium/tree/missing-ao-non-terrain-investigation) to potentially fix that issue, which affects the same lines of code, but I was unsure about it's correctness in all situations.

meta-legend:
I think it should

meta-legend:
I looked at your patch, and the diff between yours and mine is that mine makes it so that moving light-emitting blocks don't get AO, while in yours they could get AO, but only if the quad explicitly sets AO to TriState.TRUE.

meta-legend:
my fix could be modified further if you removed the "&& this.defaultAo" inside the else block of the processQuad method of the NonTerrainBlockRenderContext class to mimic the functionality of your patch if you wanted it to

douira:
Would that improve the situation? I'm not sure what this code does or what `defaultAo` is for.

meta-legend:
My code makes it so the tesselateBlock method actually uses the global AO provided by the constructor. this.defaultAo is true when Ambient Occlusion is enabled globally in the context and if the block doesn't emit light. Removing this.defaultAo would allow light-emitting blocks that are rendered through the NonTerrainBlockRenderContext to get AO if TriState.TRUE is passed through the method. If you keep it, they will never get AO even if TriState.TRUE is passed through the processQuad method.

douira:
> Removing this.defaultAo would allow light-emitting blocks that are rendered through the NonTerrainBlockRenderContext to get AO if TriState.TRUE is passed through the method.

So doing this is necessary for correctness?

meta-legend:
It depends on if the intended functionality for TriState.TRUE is to force AO no matter whether a block is emitting light or not. If it is, then i could remove it.

douira:
I don't know what the intended functionality of this code is. I'll try to find out from other people.

douira:
Well it turns out nobody knows what this code is supposed to be doing. So I'll just merge this and if it breaks we can look into it again. Thanks for taking the time to work through this.

Issue #3726, opened 2026-06-13
[26.2/port] In some cases, the red and blue colors are inverted with Fabric on the glowing effect


Lolothepro:
### Bug Description

In some cases, the red and blue colors are inverted with Fabric on the glowing effect

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/8c620f55-ed83-4e0d-a90b-81afa96bb09d" />

### Reproduction Steps

`/summon block_display ~ ~ ~ {Glowing:1b,glow_color_override:16711680,block_state:{Name:"minecraft:bell"}}`

### Log File

*

### Crash Report

*

douira:
Is this maybe related to #3590

PR #3798, opened 2026-07-02
Remove extra ABGR conversion fixes #3779

### LINKED ISSUE #3779: sodium 0.9.0/0.9.1 moving blocks have swapped color channels
### Bug Description

the Red and Blue color channels of moving blocks are swapped around causing grass to change color when being moved (very obvious in plains biomes)

Image

https://github.com/user-attachments/assets/f3570bd3-34e8-472a-bf14-9174b89b79ee

### Reproduction Steps

- go to a plains biome
- place a grass block on a sticky piston
- for better visibility reduce the tick rate to 2 or so
- activate the piston and see the grass block turn bluish

### Log File

[latest.log](https://github.com/user-attachments/files/29408155/latest.log)

### Crash Report

[crash-2026-06-27_14.02.31-client.txt](https://github.com/user-attachments/files/29408153/crash-2026-06-27_14.02.31-client.txt)

meta-legend:
Fixes #3779 

Remove extra "ColorARGB.toABGR()" conversion in the non-terrain block render context.

I can't reproduce the color issue from #3779 with this new fix, so it seems good.

### DIFF
--- a/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
+++ b/frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/render/NonTerrainBlockRenderContext.java
@@ -121,7 +121,7 @@ public class NonTerrainBlockRenderContext extends AbstractBlockRenderContext imp
         if (this.tintCacheIndex == tintIndex) {
             return this.tintCacheValue;
         } else {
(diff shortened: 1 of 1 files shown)


douira:
Can you check that this does not regress https://github.com/CaffeineMC/sodium/commit/5c4771fedea87af75edc78fd827f793c8d5c0056 ?

meta-legend:
It shouldn't regress that. I used the command from #3726 and tested the glow of the bell with my fix, and it glows red like it is supposed to.
image

PR #3800, opened 2026-07-03
Fix non-terrain block lighting to comply with the FRAPI contract

douira:


meta-legend:
What was the intended functionality in the end?

douira:
I had [talked to Pepper in #dev-sodium](https://discord.com/channels/602796788608401408/651120262129123330/1522403868644868157) and it turns out that the behavior remained incorrect, so I had a look at the documentation on the [FRAPI interface](https://github.com/FabricMC/fabric-api/blob/26.2/fabric-renderer-api-v1/src/client/java/net/fabricmc/fabric/api/client/renderer/v1/mesh/MutableQuadView.java#L441) and implemented it based on that.

PR #3807, opened 2026-07-06
gui: make max FPS slider dynamic based on monitor refresh rate

### LINKED ISSUE #3781: Increase max fps slider range
### Request Description

Currently, the max fps slider only goes to 250, the next step after is Unlimited.

I have a 280Hz monitor, so I would like to be able to set my max fps beyond 250.
Playing with Unlimited fps is annoying because it makes my PC run at 1000+FPS, much hotter and louder than it really should.

eyadmkv:
Fixes #3781: Automatically calculate the slider's upper bound based on the current monitor's refresh rate, removing the static hardcoded cap and providing better support for high-refresh-rate displays

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
@@ -38,6 +38,8 @@ import org.joml.Vector4f;
 import org.jspecify.annotations.Nullable;
 import org.lwjgl.opengl.GL;
 import org.lwjgl.opengl.GLCapabilities;
+import org.lwjgl.glfw.GLFW;
+import org.lwjgl.glfw.GLFWVidMode;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -141,24 +143,7 @@ public class SodiumConfigBuilder implements ConfigEntryPoint {
                 .addPage(this.buildPerformancePage(builder));
     }
 
-private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
-        int monitorRefreshRate = 60;
-        try {
-            long windowHandle = Minecraft.getInstance().getWindow().handle();
-            long monitorHandle = org.lwjgl.glfw.GLFW.glfwGetWindowMonitor(windowHandle);
-            if (monitorHandle == 0L) {
-                monitorHandle = org.lwjgl.glfw.GLFW.glfwGetPrimaryMonitor();
-            }
-            org.lwjgl.glfw.GLFWVidMode vidMode = org.lwjgl.glfw.GLFW.glfwGetVideoMode(monitorHandle);
-            if (vidMode != null) {
-                monitorRefreshRate = vidMode.refreshRate();
-            }
-        } catch (Exception e) {
-            // fallback if window context is uninitialized during early config building
-        }
-        int dynamicMaxFps = Math.max(260, ((monitorRefreshRate + 9) / 10) * 10 + 10);
-        final int finalMaxFps = dynamicMaxFps;
-        
+    private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
         var generalPage = builder.createOptionPage().setName(Component.translatable("sodium.options.pages.general"));
         generalPage.addOptionGroup(builder.createOptionGroup()
                 .addOption(
@@ -320,8 +305,10 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
                                 .setStorageHandler(this.vanillaStorage)
                                 .setName(Component.translatable("options.framerateLimit"))
                                 .setTooltip(Component.translatable("sodium.options.fps_limit.tooltip"))
-                                .setValueFormatter(value -> value >= finalMaxFps ? Component.translatable("options.framerateLimit.max") : Component.literal(value + " FPS"))
-                                .setRange(10, finalMaxFps, 10)
+                                .setValueFormatter(ControlValueFormatterImpls.fpsLimit())
+                                .setValidatorProvider(
+                                        (state) -> new Range(10, this.getMaxFramerateLimit(), 10),
+                                        ConfigState.UPDATE_ON_REBUILD, ConfigState.UPDATE_ON_APPLY)
                                 .setDefaultValue(60)
                                 .setBinding(this.vanillaOpts.framerateLimit()::set, this.vanillaOpts.framerateLimit()::get)
                 )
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -328,24 +345,7 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
...
-                    }
-                })
-                .setElementNameProvider(EnumOptionBuilder.nameProviderFrom(
-                        Component.translatable("options.graphicsApi.default"),
-                        Component.translatable("options.graphicsApi.opengl"),
-                        Component.literal("Prefer Vulkan")))
-                .setDefaultValue(PreferredGraphicsApi.DEFAULT)
-                .setFlags(OptionFlag.REQUIRES_GAME_RESTART)
-                .setBinding((value) -> this.vanillaOpts.preferredGraphicsBackend().set(value), () -> this.vanillaOpts.preferredGraphicsBackend().get())));
+
```
douira, line 353: why was this removed?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -303,8 +320,8 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
                                 .setStorageHandler(this.vanillaStorage)
                                 .setName(Component.translatable("options.framerateLimit"))
                                 .setTooltip(Component.translatable("sodium.options.fps_limit.tooltip"))
-                                .setValueFormatter(ControlValueFormatterImpls.fpsLimit())
-                                .setRange(10, 260, 10)
+                                .setValueFormatter(value -> value >= finalMaxFps ? Component.translatable("options.framerateLimit.max") : Component.literal(value + " FPS"))
```
douira: why was this change made?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -303,8 +320,8 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
...
+                                .setRange(10, finalMaxFps, 10)
```
douira: this should not be static since the video mode can be changed and then the maximal refresh rate also changes.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -141,7 +141,24 @@ private void buildFullConfig(ConfigBuilder builder) {
                 .addPage(this.buildPerformancePage(builder));
     }
 
-    private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
+private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
+        int monitorRefreshRate = 60;
+        try {
+            long windowHandle = Minecraft.getInstance().getWindow().handle();
+            long monitorHandle = org.lwjgl.glfw.GLFW.glfwGetWindowMonitor(windowHandle);
```
douira: why are you doing fully qualified package imports?

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -329,7 +333,7 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
                 )
         );
         generalPage.addOptionGroup(builder.createOptionGroup().addOption(builder.createEnumOption(Identifier.fromNamespaceAndPath("sodium", "general.graphics_api"),
-                PreferredGraphicsApi.class)
+                        PreferredGraphicsApi.class)
```
douira: please clean up the diff

eyadmkv:
Hi @douira, thank you again for the feedback. I have pushed an update addressing the points you raised, I hope they meet your expectations!
FYI: The Graphics API being removed was an accident

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -346,9 +350,34 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
...
+     * this computes an appropriate maximum for the framerate limit slider based on the current
+     * monitor's refresh rate, so users on high-refresh-rate displays aren't capped at 260
+     * + falls back to the vanilla default max if the window/monitor context isn't available yet
+     * (e.g. during early config building).
+     */
+    private int getMaxFramerateLimit() {
+        int monitorRefreshRate = 60;
+        try {
+            long windowHandle = Minecraft.getInstance().getWindow().handle();
+            long monitorHandle = GLFW.glfwGetWindowMonitor(windowHandle);
```
douira, line 368: We already have a reference to the window on this class in the field `window`.

douira (changes requested):

eyadmkv:
Tell me if there's something wrong, I'm happy to learn from you!

eyadmkv:
any updates @douira ?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -346,9 +350,36 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
...
+            }
+            GLFWVidMode vidMode = GLFW.glfwGetVideoMode(monitorHandle);
+            if (vidMode != null) {
+                monitorRefreshRate = vidMode.refreshRate();
+            }
+        }
+        } catch (Exception e) {
+            // fallback if window/monitor context is unavailable
+        }
+        return Math.max(260, ((monitorRefreshRate + 9) / 10) * 10 + 10);
```
douira, line 380: What is this mysterious integer calculation about?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -304,7 +306,9 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
                                 .setName(Component.translatable("options.framerateLimit"))
                                 .setTooltip(Component.translatable("sodium.options.fps_limit.tooltip"))
                                 .setValueFormatter(ControlValueFormatterImpls.fpsLimit())
-                                .setRange(10, 260, 10)
+                                .setValidatorProvider(
+                                        (state) -> new Range(10, this.getMaxFramerateLimit(), 10),
+                                        ConfigState.UPDATE_ON_REBUILD, ConfigState.UPDATE_ON_APPLY)
```
douira, line 311: Can you explain why you've picked these flags?

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
```diff
@@ -304,7 +306,9 @@ private OptionPageBuilder buildGeneralPage(ConfigBuilder builder) {
...
                                 .setDefaultValue(60)
                                 .setBinding(this.vanillaOpts.framerateLimit()::set, this.vanillaOpts.framerateLimit()::get)
```
douira, line 313: Are you sure the value won't simply be clamped the value to 260 or something since we're writing it to vanilla's option system?

douira (changes requested):

personalized-advertising:
@douira why did you close this poor fellow's PR was it added in a different PR? Because this feature is needed.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
eyadmkv: because the maximum framerate limit is no longer a static value it now relies on this.getMaxFramerateLimit(), the valid range is dynamic and depends on other runtime or config states

eyadmkv: `ConfigState.UPDATE_ON_REBUILD`: if a user changes a dependent setting, the ui will rebuild to reflect that change when it rebuilds getMaxFramerateLimit() might return a completely different value this flag ensures that the slider/input field for the framerate limit immediately updates its min/max bounds to match the new reality which prevents the user from interacting with stale ui constraints

`ConfigState.UPDATE_ON_APPLY`: this tells the configuration system to re-evaluate the validator right after the configuration is applied/saved

applying settings often changes the actual runtime state of the game by updating on apply, the system guarantees that once the new settings take effect, the framerate limit is re-validated against the newly established runtime state.
It's a safety net to clamp or correct the framerate value if the newly applied state makes the previously selected framerate invalid for any reason.

common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
eyadmkv: Yes sir, I'm sure. the 260 limit is only enforced by the vanilla UI slider, not the underlying data model. the actual integer field in vanilla's options doesn't have a hard cap of 260

eyadmkv:
i took a break from github for a few weeks, he probably closed it due to inactivity @personalized-advertising

eyadmkv:
@douira i would appreciate it if you take another look.

douira:
Also see https://github.com/CaffeineMC/sodium/pull/3632 which is likely to conflict with this. We need to ensure the combination works and follows our code standards. I haven't had a lot of time to review prs like this recently.

eyadmkv:
@douira That makes sense re: #3632. Since that PR replaces the slider with a numeric input, the range/validator logic I have here would probably just become dead code once it lands.

I think it makes more sense to hold off on further changes to the slider approach for now. Once #3632 is merged, I can rework this PR around the numeric input instead. For example, we could use the detected monitor refresh rate as the default and show a warning if someone enters a value higher than that, rather than enforcing a hard cap.

Let me know if that sounds good, or if you'd rather I close this and pick it up as a follow-up once #3632 is merged.
Thanks for taking the time to look into this, Appreciate it.

PR #3815, opened 2026-07-09
Optimize the dried ghast model

muzikbike:
This commit deletes six faces from models/block/dried_ghast.json which do not need to exist: they can only ever be seen by intersecting the main central cube of the block. The outward appearance is entirely unchanged.

Formatting is otherwise unchanged from the model file as it exists in minecraft.jar.


muzikbike:
...any idea what that threatening message is coming from? I've never experienced it before with respect to just uploading a fixed model file.

jellysquid3:
What?

muzikbike:
"Merging is blocked
Cannot update this protected ref."

No other pull request I've checked appears to have this?

PR #3818, opened 2026-07-12
Support FRAPI on NeoForge

### LINKED ISSUE #3766: NeoForge version declares Fabric renderer but never registers it
### Bug Description

Related issue: https://github.com/Sinytra/ForgifiedFabricAPI/issues/275

The NF version of Sodium declares that it contains a renderer in `neoforge.mods.toml`, but if the Fabric rendering API is present (for example, through Forgified Fabric API), it never registers one.

This causes Indigo to disable itself, awaiting something to be registered. Then when you join a world, there is no renderer active the Fabric API can use, so the game crashes.

### Reproduction Steps

1. Install latest NeoForge (26.1.2.76) and Sodium (NeoForge version, 0.8.12+mc26.1.2)
2. Install Forgified Fabric API and Forgified Fabric Loader (0.152.1+3.0.5+26.1.2 and 2.5.75+0.18.4+26.1 respectively)
3. Try creating/opening/joining any world and the game will crash

### Log File

https://gist.github.com/leetftw/90cab33752f3ba3cbc58f7a0830ef7ba

### Crash Report
https://gist.github.com/leetftw/b168506f420a449ae7ee2d1e8582c8b8

Su5eD:
## The issue

Sodium currently declares it contains a FRAPI renderer implementation on NeoForge, but does not actually include it. See #3766 for details.

- #3766

## The proposal

This PR adds support for FRAPI rendering on NeoForge by bundling the FRAPI module currectly included on Fabric. Services and mixins provided by this module are loaded conditionally only when FRAPI is present, ensuring that existing behavior remains unchanged.

### Changes

- Bundle FRAPI module code in Sodium "mod" jar on NeoForge
- Add conditional providers for `PlatformModelEmitter` and `FRAPIRegistrar` (renamed from `FRAPIProvider`) services
	- This allows us to only load these services when FRAPI is available

## Testing

I tested the integration in a 26.1.2 NeoForge 26.1.2.77 instance with the following mods installed:
- [Forgified Fabric API v0.154.0+3.2.2+26.1.2](https://modrinth.com/mod/forgified-fabric-api/version/r7jVSmZ9)
- [Chisels & Bits v26.1.2.33](https://www.curseforge.com/minecraft/mc-mods/chisels-bits/files/8087317)

I used a Chisel to create a chiseled block, which is rendered using the `FRAPIEmitter`. The results appear to be the same as when only Sodium is installed.

## Alternatives

I tried running Sodium alongside FFAPI without the `contains_renderer` metadata attribute and everything appeared to work fine - I was able to load into the world and all blocks had the expected appearance. Therefore, as a temporary measure until this PR is merged, I'd suggest at least removing the `contains_renderer` attribute to prevent the game from crashing.

Closes #3766


douira:
Does this mean that it will route usage of FRAPI through our FRAPI implementation on NeoForge? Or is it somehow using a different FRAPI implementation and piping into our meshing system on another path?

douira:
Closes #3766 
(seems like this doesn't work with comments, fine)

Su5eD:
> Does this mean that it will route usage of FRAPI through our FRAPI implementation on NeoForge? Or is it somehow using a different FRAPI implementation and piping into our meshing system on another path?

The former is the case here, yes. The Forgified Fabric API provides the original FRAPI to mods on NeoForge, so from Sodium's perspective, nothing changes. On FFAPI's end, the FRAPI implementation uses additional context and extension methods added by NeoForge to integrate with its rendering APIs.

frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/FRAPIEmitterProvider.java
```diff
@@ -0,0 +1,18 @@
...
+
+import net.caffeinemc.mods.sodium.client.services.PlatformModelEmitter;
+import net.caffeinemc.mods.sodium.client.services.PlatformModelEmitterProvider;
+import net.caffeinemc.mods.sodium.client.services.PlatformRuntimeInformation;
+
+public class FRAPIEmitterProvider implements PlatformModelEmitterProvider {
+    @Override
+    public boolean isEnabled() {
+        return PlatformRuntimeInformation.getInstance().isModInLoadingList("fabric-renderer-api-v1") 
+            || PlatformRuntimeInformation.getInstance().isModInLoadingList("fabric_renderer_api_v1");
```
douira: Where do these two different notations come from?

douira (changes requested):
I'm in favor of this PR, it looks ok to me, I'm just wondering about the double mod presence check.

frapi/src/main/java/net/caffeinemc/mods/sodium/client/render/frapi/FRAPIEmitterProvider.java
BluSpring: NeoForge doesn't support dashes in mod IDs, so the first one would be for handling Fabric whereas the second one would be for handling NeoForge, at least that's what I'd assume

douira: Maybe we should instead have the `PlatformRuntimeInformation` itself handle this platform-specific behavior? though it's irrelevant to this PR for now, I might change it later.

douira:
Is this also relevant to 1.21.1, 1.21.11, and 26.2?

Su5eD:
It's effectively relevant to 26.1 and higher. 1.21.1 already contains support for FRAPI, not sure about 1.21.11 but we're not planning on porting to it anyway so I'd say it's a nonissue.

douira:
Is there a specific reason for this PR being targeted at 26.1.2 and not `dev`? Typically we target dev and then cherry-pick into older versions unless something is irrelevant to the current version.

Su5eD:
That's an error on my part, I'll rebase the branch and change the target to dev.

douira:
If there's any changes you need to make to switch the base that are non-trivial, I'd appreciate if they are kept around so we can use them for the 26.1.2 cherry-pick.

Su5eD:
Okay should be ready now. I've tried cherry picking the commit to 26.1.2/stable and it worked without any conflicts.

PR #3820, opened 2026-07-13
Fix generated item side-face stitching 

### LINKED ISSUE #3715: Item Stitching Is Offset/Incorrect
### Bug Description

When Sodium is installed, the stitching of held items, like torches, is offset and appears incorrect.

Vanilla:
Image

Sodium:

Image

### Reproduction Steps

Minecraft Version: 26.1.2
Fabric Version:
Sodium Version:

1. Install Sodium
2. Open a world
3. Hold an item (e.g. Torches)
4. Observe the incorrect item stitching.

### Log File

[latest.log](https://github.com/user-attachments/files/28785372/latest.log)

### Crash Report

[crash-2026-06-10_02.39.13-client.txt](https://github.com/user-attachments/files/28785414/crash-2026-06-10_02.39.13-client.txt)

meta-legend:
This fixes #3715.

### Background
Isolation and cause [analysis](https://github.com/CaffeineMC/sodium/issues/3715#issuecomment-4879772053) is present in the original issue (long story short, it was caused by this [PR](https://github.com/CaffeineMC/sodium/pull/3551)).

### Why
Previously, generated item side faces were being merged into longer quads. This reduced quad count, but it also meant "UV_SHRINK" was only applied to the outer ends of each merged run instead of each side texel (UV_SHRINK is present in the Fabric and Neoforge specific item-builders). This is most visible on simple generated item textures such as torches, levers, and other torch-like items. The stitching error is more visible on these items because they have large flat regions and long exposed side-face runs, so the shifted side-face texel boundaries are easy to compare against the front face.

### Fix
This reverts that previous optimization for generated side-faces. This PR changes the generated item side-face builder to emit each occupied side texel as its own "SideFace/quad" again. 

Torch after my fix:
image

### Tradeoff
Since this removes the previous side-face merging optimization for generated item side faces, performance might be marginally worse on low-res resource packs, but it could be more noticeable while using high-res resource packs.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilderBase.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilderBase.java
@@ -134,11 +134,11 @@ public class ImprovedItemModelBuilderBase {
         public List<SideFace> buildSideFaces() {
             var output = new ReferenceArrayList<SideFace>();
 
-            // Merges and collects all faces from different directions.
-            buildMergedFaces(output, this.up, SideDirection.UP);
-            buildMergedFaces(output, this.down, SideDirection.DOWN);
-            buildMergedFaces(output, this.left, SideDirection.LEFT);
-            buildMergedFaces(output, this.right, SideDirection.RIGHT);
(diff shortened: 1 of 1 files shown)


douira:
I'd rather have our optimized face builder be fixed than the optimization being reverted.

PR #3825, opened 2026-07-13
Fix generated item side-face stitching v2

### LINKED ISSUE #3715: Item Stitching Is Offset/Incorrect

meta-legend:
This fixes #3715.
This is a v2 of a previous PR (by me), which, instead of fixing the optimization it reverted it.

### Background
Isolation and cause [analysis](https://github.com/CaffeineMC/sodium/issues/3715#issuecomment-4879772053) is present in the original issue (long story short, it was caused by this [PR](https://github.com/CaffeineMC/sodium/pull/3551)).

### Why
Previously, generated item side faces were being merged into longer quads. This reduced quad count, but it also meant "UV_SHRINK" was only applied to the outer ends of each merged run instead of each side texel (UV_SHRINK is present in the Fabric and Neoforge specific item-builders). This is most visible on simple generated item textures such as torches, levers, and other torch-like items. The stitching error is more visible on these items because they have large flat regions and long exposed side-face runs, so the shifted side-face texel boundaries are easy to compare against the front face.

### Fix
I did a previous PR on this same issue, which just reverted the optimization. This time, instead of reverting it completely, it keeps the side-face merging optimization, but makes it conditional. Adjacent side texels are now only merged when their sprite pixel colors match across the relevant animation frames. This preserves merged quads for visually identical runs, while splitting runs where merging could shift visible texel boundaries.

Example:
A A A A -> one merged side face
A A B B -> two merged side faces
A B C D -> four side faces

### Test
Torch after my fix:
Screenshot 2026-07-13 002340

### Tradeoff
This keeps the optimization for same-color side-face runs, so it avoids fully reverting the previous optimization. The optimization is only lost where adjacent exposed side texels differ visually, which is the case where merging can produce incorrect stitching.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilderBase.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/immediate/model/ImprovedItemModelBuilderBase.java
@@ -1,5 +1,7 @@
 package net.caffeinemc.mods.sodium.client.render.immediate.model;
 
+import it.unimi.dsi.fastutil.ints.Int2IntMap;
+import it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.objects.ReferenceArrayList;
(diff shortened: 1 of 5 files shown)


Argon4W:
I think the simplest way to solve this issue is to remove the UV_SHRINK entirely and test with various resource packs.

Argon4W:
I am worrying about that adding those checks through nested map-of-maps will fail the random transparent pixel 512x resource pack test that douira used to test my PR with.

douira:
> I think the simplest way to solve this issue is to remove the UV_SHRINK entirely and test with various resource packs.

I would be interested in seeing whether this is possible.

Argon4W:
> > I think the simplest way to solve this issue is to remove the UV_SHRINK entirely and test with various resource packs.
> 
> I would be interested in seeing whether this is possible.

Can I open a draft for that?

douira:
Sure, if not everything is broken after removing it? I don't know why it exists in the first place, but surely at some point it had a reason. Is there maybe leakage of neighboring sprites without it?

Argon4W:
As far as I know, both item and block atlas now has bleeding around the sprites, so I assume that removing the UV_SHRINK won't cause the sprite leakage.

meta-legend:
I guess you could maybe use my PR as a bandage for #3715 until a broader rewrite happens?

douira:
Obviated by #3831

PR #3833, opened 2026-07-17
Avoid repeated region lookups when building render lists

NovaVanity:
The collector did a region map lookup for every visited section on every render list update. The traversal visits sections of the same region in long runs, so remembering the last region skips almost all of those lookups. The reset logic still runs once per region through the existing frame check.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/lists/VisibleChunkCollector.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/render/chunk/lists/VisibleChunkCollector.java
@@ -4,6 +4,7 @@ import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import net.caffeinemc.mods.sodium.client.render.chunk.LocalSectionIndex;
 import net.caffeinemc.mods.sodium.client.render.chunk.region.RenderRegion;
 import net.caffeinemc.mods.sodium.client.render.chunk.region.RenderRegionManager;
+import org.jspecify.annotations.Nullable;
 
 /**
  * The async visible chunk collector is passed into a section tree to collect visible chunks.
@@ -14,6 +15,12 @@ public class VisibleChunkCollector implements CoordinateSectionVisitor, RenderLi
 
     private final ObjectArrayList<ChunkRenderList> sortedRenderLists;
 
+    // sections of the same region are visited in runs, so memoizing the last region avoids most of the region map lookups
+    private int lastRegionX = Integer.MIN_VALUE;
+    private int lastRegionY;
+    private int lastRegionZ;
+    private @Nullable ChunkRenderList lastRenderList;
+
     public VisibleChunkCollector(RenderRegionManager regions, int frame) {
         this.regions = regions;
         this.frame = frame;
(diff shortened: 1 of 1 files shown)


douira:
Please show us some benchmarks to decide if this optimization has an impact.

NovaVanity:
> Please show us some benchmarks to decide if this optimization has an impact.

Since the repo doesn't have a benchmark harness for this (at least not one I could find), I wrote a quick JMH benchmark to simulate the collection loop. It uses the exact same map structures, region shifts, and traversal as the real code. Each run simulates a full render list build over a 24-section-tall world (the heavy, low-occlusion case).
Results:
result

The memoized loop is consistently around 1.6x to 2x faster, saving about 16us at RD 8, 35us at RD 16, and around 190us at RD 31. It's a nice cheap optimization for the render thread.

Benchmark:
https://gist.github.com/NovaVanity/0855031c989bf9ec72c1dbfbf999bd25

douira:
Thanks, but I was more referring to a benchmark of the in-game performance, or the measured duration of the render list generation.

NovaVanity:
Ah, I see. I'll do that now if you still want it, just give me a bit

NovaVanity:
> Thanks, but I was more referring to a benchmark of the in-game performance, or the measured duration of the render list generation.

Ok, I measured it in-game. I wrapped the render list generation (readRenderListFromTree) in a nanoTime timer that logs a 5 second average, then flew around in spectator at RD 32 on the same seed and path, once on dev and once with the patch.

There's not really any measurable difference. Dev settles around 3.4-3.7us per build once everything is warmed up and the patch sits around 3.1-3.4us, so it's within noise. Both spike to 40-70us during heavy chunk loading.

```
dev: [timing] render list build: avg=3.5us max=35.3us builds=573 (over 5s)
patched:  [timing] render list build: avg=3.1us max=13.6us builds=581 (over 5s)
```
Turns out occlusion and frustum culling keep the visible set so small in normal play that the whole render list build is only a few microseconds to begin with, and the lookups this patch removes are maybe a microsecond of that. My JMH numbers were the everything-visible worst case.

So while the collection loop is around 2x faster in isolation, that doesn't translate into anything visible in-game. I still think it's a reasonable cleanup since it's strictly fewer operations for 30 lines, similar reasoning to [#3773](https://github.com/CaffeineMC/sodium/pull/3773), but I understand if that's not enough to justify merging it.

douira:
As far as I remember render list generation from the tree takes around 150µs, so I think something in this measurement is off. Are you sure you were measuring the right thing?

NovaVanity:
The function I timed was readRenderListFromTree, so the whole traversal plus collection and region sorting. I added visible section and region counts to the timing and re-ran both branches. Reason was the build time scales with the visible section count. It's around 30-50us at a few hundred sections, around 90us at 2000 and around100-107us at 3000, so 150us would be a scene with roughly 4-5k visible sections. My first test flight only had a few hundred sections visible after culling, which is where the 3us came from.
At matched section counts there's a small difference between dev and the patch, though that could be nosie:
```
sections   dev        patched
~1550      76.3us     68.8us
~2000      91.5us     90.6us
~2950      105.3us    102.3us
```

Not a major difference, still it's strictly fewer operations per build.

Issue #2831, opened 2024-10-24
macOS: Fullscreen resolution "unsupported" on 1.21.3


Proton0:
### Bug Description

I wanted to reduce my fullscreen resolution due to macOS having retina displays causing my M1 air to lag (especially with shaders) but i noticed that it said that its unavailable and is only supported on Windows

![Minecraft 1.21.3](https://github.com/user-attachments/assets/5191c452-9f87-4073-a174-73591453f8db)

but on other versions (1.21.0) it allows me to change the fullscreen resolution
![Minecraft 1.21.0](https://github.com/user-attachments/assets/10a0e042-49da-42f3-ab2f-de3c279073db)

I launched Minecraft 1.21.3 without Sodium and the fullscreen resolution slider works and is supported
![Minecraft 1.21.3 without sodium](https://github.com/user-attachments/assets/03e92d61-c1ee-41c7-8870-cb8143d87a7e)

My Minecraft version is 1.21.3 and sodium version is 0.6.0-beta.3-fabric and my macOS Version is macOS Seqouia (15.0.1 build 24A348)

### Reproduction Steps

1. Play on Minecraft 1.21.3 on macOS
2. Check fullscreen resolution (should say unsupported)

### Log File

[11:12:51] [main/INFO]: Loading Minecraft 1.21.3 with Fabric Loader 0.16.7
[11:12:51] [main/INFO]: Loading 11 mods:
	- fabricloader 0.16.7
	   \-- mixinextras 0.4.1
	- java 21
	- minecraft 1.21.3
	- sodium 0.6.0-beta.3+mc1.21.2
	   |-- fabric-api-base 0.4.48+c47b9d431a
	   |-- fabric-block-view-api-v2 1.0.16+c47b9d431a
	   |-- fabric-renderer-api-v1 4.0.3+c47b9d431a
	   |-- fabric-rendering-data-attachment-v1 0.3.54+73761d2e1a
	   |-- fabric-rendering-fluids-v1 3.1.13+c47b9d431a
	   \-- fabric-resource-loader-v0 3.0.5+c47b9d431a
[11:12:51] [main/INFO]: SpongePowered MIXIN Subsystem Version=0.8.7 Source=file:/Users/proton/Library/Application%20Support/PrismLauncher/libraries/net/fabricmc/sponge-mixin/0.15.3+mixin.0.8.7/sponge-mixin-0.15.3+mixin.0.8.7.jar Service=Knot/Fabric Env=CLIENT
[...]

PR #3843, opened 2026-07-23
Re-enable the "Fullscreen Resolution" control on Linux

Kira-NT:
Hi! As of today, there's no longer a need to artificially disable the resolution control on Linux - it now works properly on the major Wayland desktops. I've tested it on my laptop with KDE 6.6.5 and GNOME 50 using both Vulkan and OpenGL, as well as both my Intel iGPU and Nvidia dGPU *(can it even be called that?)*. Here are some examples:

| sodium-xwayland-opengl-nvidia | sodium-xwayland-vulkan-nvidia | sodium-xwayland-opengl-intel |
|-|-|-|

I also tested changing Minecraft's fullscreen resolution on another machine running an older version of KDE (6.4.3), and there it still doesn't work, which suggests that something was fixed/improved relatively recently. However, the important thing here is that the issue is no longer present on the latest-ish versions of the said desktop environments.

And since the problem is no longer reproducible on KDE and GNOME, which make up like 99% of the modern Linux desktop, and the consensus in #2831 seems to be that it doesn't make sense to withhold this feature from everyone just because it may still be a bit junky on a handful of systems, I think it's time to retire this check.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/gui/SodiumConfigBuilder.java
@@ -15,7 +15,6 @@ import net.caffeinemc.mods.sodium.api.config.option.OptionImpact;
 import net.caffeinemc.mods.sodium.api.config.option.Range;
 import net.caffeinemc.mods.sodium.api.config.structure.*;
 import net.caffeinemc.mods.sodium.client.SodiumClientMod;
-import net.caffeinemc.mods.sodium.client.compatibility.environment.OsUtils;
 import net.caffeinemc.mods.sodium.client.compatibility.workarounds.Workarounds;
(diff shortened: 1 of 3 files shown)


douira:
We've decided to do this only on 26.3 where they've switched to using SDL3 instead of GLFW.

Kira-NT:
> they've switched to using SDL3 instead of GLFW

Not the way I find out one of my mods is gonna be completely broken in 26.3 :')

> We've decided to do this only on 26.3

Just as a note, I also tested changing the fullscreen resolution in every version up to 1.16, and whatever GLFW has been doing all this time now works properly on all of them. Given that things are unlikely to suddenly regress, it may be worth dropping this limitation for the versions you currently support as well. That way, once you stop supporting them, players revisiting/stuck on older Minecraft versions won't run into it in the future. But that's just my two cents. If you decide to only allow changing the resolution starting with the next Minecraft release, that's perfectly fine with me personally!

PR #3848, opened 2026-07-25
Restore the fast quad encoder for entity-format buffers

### LINKED ISSUE #3847: Entity-format buffers skip the quad encoder since the 26.2 port
### Bug Description

The port to 26.2 swapped `!this.fastFormat` for `!this.blockFormat` in `putBakedQuad`, but those aren't the same check. 26.1's `fastFormat` was true for BLOCK and ENTITY both, while 26.2 gives each its own flag, so what used to mean "neither of these" now means "anything except BLOCK" and every entity-format buffer drops to the vanilla path.

It reads like the rename in went positional, which would explain why the comment on that line still says "check for ENTITY". `!this.blockFormat && !this.entityFormat` gets the old behaviour back. Will open PR fix linked.

### Reproduction Steps

Not relevant

### Log File

Not relevant

### Crash Report

Not relevant

NovaVanity:
Fixes #3847. I dropped the "check for ENTITY" comment since the flags now say it anyway.


douira:
Do you think this is also applicable to 26.1?

NovaVanity:
> Do you think this is also applicable to 26.1?

No I don't think so. There vanilla computes `fastFormat = (format == ENTITY) || (format == BLOCK)`, so `!fastFormat` already meant "neither BLOCK nor ENTITY" and entity-format buffers still reached `BakedModelEncoder` fine.

PR #3850, opened 2026-07-27
Removed available buffer clear at end of each frame

GeForceLegend:
### Before merging:

- I don't know if I should do the same thing on mixin.json and accesswidener to other package. If I should, please tell me which package need it too, or do this immediately after merging.
- Vanilla will acquire a pair of staging buffer and vertex buffer for every item in gui draw, each takes 256KB at minimum size. When open creative inventory, the first frame may draw over 50+ unique item, and allocating about 20+MB of VRAM. After merging this pr, those 20+MB VRAM will not get cleaned until game close. This might be good prepaing for other item in gui draws, but if we can make item in gui draw shares the same staging and vertex buffer, that should be much better than current.

The reason why we need to fix is already described in [dev-sodium channel](https://discord.com/channels/602796788608401408/651120262129123330/1529730425298096179).

Common frame:
QQ20260727-000739

Frames that triggers buffer create:
QQ20260727-000733

This will be worse on Iris, as when using shaders, frame time will be much longer, so time spent for glFence to be sync is more unstable, and Iris is using it's on buffer pool for shadow, which means it will be double affected by it.


GeForceLegend:
Close as vanilla fix on it has been backported

PR #3866, opened 2026-08-05
Add a way to specify priority for option overrides and overlays

Kira-NT:
At the moment, if two (or more) different mods attempt to register an override or an overlay for the same option, it results in a hard crash. This is quite unfortunate, since two mods with overlapping features are not necessarily incompatible, and yet they would be unable to coexist if both of them want/need to override some of Sodium's options.

As an example, say, hypothetically, I maintain a mod which provides an option for borderless fullscreen. Then, hypothetically, a maintainer of a popular variety mod gets a bright idea of providing an option for borderless fullscreen of their own, alongside their other tweaks. They obviously, and still hypothetically, implement it incorrectly. And now our mods are incompatible, simply because we both override `sodium:general.fullscreen_mode`, meaning players must choose between normal borderless fullscreen and the features provided by the second hypothetical mod, which is not ideal.

This PR solves the issue by allowing consumers of Sodium's Config API to specify the priority of their option overrides/overlays, similarly to how we already do this for mixins and such. The override/overlay with the highest priority wins. If there are still multiple contenders with the same priority, an exception is thrown, just as it was before.

The new API is fully backwards compatible and does not break existing mods, as it simply adds new overloads for the `registerOptionReplacement` and `registerOptionOverlay` methods instead of replacing them. The original versions of these methods now simply register overrides and overlays with the default priority of `0`.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/Config.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/Config.java
@@ -80,19 +80,23 @@ public class Config implements ConfigState {
     }
 
     private void applyOptionChanges() {
-        var overrides = new Object2ReferenceOpenHashMap<Identifier, OptionOverride>();
-        var overlays = new Object2ReferenceOpenHashMap<Identifier, OptionOverlay>();
+        var overrides = new Object2ReferenceOpenHashMap<Identifier, ObjectArrayList<OptionOverride>>();
+        var overlays = new Object2ReferenceOpenHashMap<Identifier, ObjectArrayList<OptionOverlay>>();
 
-        // collect overrides and overlays and validate them, also against each other
+        // collect overrides and overlays and validate them
         for (var modConfig : this.modOptions) {
             for (var override : modConfig.overrides()) {
                 if (override.target().getNamespace().equals(modConfig.configId())) {
                     throw new IllegalArgumentException("Override by mod '" + modConfig.configId() + "' targets its own option '" + override.target() + "'");
                 }
 
-                var oldOverride = overrides.put(override.target(), override);
-                if (oldOverride != null) {
-                    throw new IllegalArgumentException("Multiple overrides for option '" + override.target() + "'! Sources: " + oldOverride.source() + " and " + override.source());
+                var existingOverrides = overrides.computeIfAbsent(override.target(), _ -> new ObjectArrayList<>(1));
+                var existingOverridePriority = existingOverrides.isEmpty() ? override.priority() : existingOverrides.getFirst().priority();
+                if (override.priority() > existingOverridePriority) {
+                    existingOverrides.clear();
+                }
+                if (override.priority() >= existingOverridePriority) {
+                    existingOverrides.add(override);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/Config.java
```diff
@@ -134,17 +145,20 @@ private void applyOptionChanges() {
...
-                                this.exchangeOption(options, i, overlaidOption, option);
-                            } catch (Exception e) {
-                                throw new IllegalArgumentException("Failed to apply overlay from '" + overlay.source() + "' to option '" + option.id + "'", e);
-                            }
+                        var optionOverlays = overlays.get(option.id);
+                        if (optionOverlays == null || optionOverlays.isEmpty()) {
+                            continue;
+                        }
+                        if (optionOverlays.size() > 1) {
+                            throw new IllegalArgumentException("Multiple overlays for option '" + optionOverlays.getFirst().target() + "'! Sources: " + optionOverlays.getFirst().source() + " and " + optionOverlays.getLast().source());
```
douira, line 153: Why does this throw?

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/Config.java
Kira-NT: For the same reason it did before - multiple mods have registered overlays with the same priority, so instead of selecting a random one as a winner, which would result in non-deterministic and hard-to-debug behavior, Sodium chooses to crash.

At this point, `optionOverlays` and `optionOverrides` contain the elements with the highest priority. And there must always be exactly one such element for the operation to succeed.

Kira-NT: Although, your question got me thinking: technically, since overlays now have a priority value attached to them, we could also combine them from highest priority to lowest, with the base option at the very bottom. Is that what you meant and/or would you like me to implement it this way? Should be a pretty minor patch to `OptionBuilderImpl` if so.

I'm not sure how good of an idea that is, though, as, similarly to overrides, mods probably won't expect their implementations to be mixed and matched with somebody else's.

Kira-NT: Yeah, no, I take it back. Say, one mod provides an overlay with a custom `enabledProvider` and `valueFormatter`, and then another one with a higher priority only sets its `enabledProvider`, creating a disconnect between it and the effective formatter. Only the highest-priority overlay should take effect to avoid such cases.

douira: I found the exception string somewhat confusing given the API changes but I see now.

douira, line 153: maybe instead of reporting an error with the first and last overlay of the same priority we could simply raise this error immediately when we detect this on line 113? This way we'd also not need to store lists of overlays if non-singleton lists will end up as an error anyway.

douira (changes requested):

common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/Config.java
Kira-NT: Nope, we can't. There is, in fact, a reason I moved the check out of the first passthrough and had to store lists instead of individual items.

Consider the case where mods `A` and `B` register the same override/overlay with the same priority. If they are the only mods present, that should result in an error. Now, however, also consider mod `C`, which registers the same override/overlay, but with a higher priority.

These mods can be loaded in any order: `ABC`, `BAC`, `CAB`, `ACB`, `BCA`, or `CBA`. If you only compare the current element against the previously seen one with the highest priority, you end up with a situation where the exact same scenario fails 2 out of 6 times *(when `C` is encountered after both `A` and `B`)*, but succeeds the other 4 out of 6 times. This is not acceptable.

For the priority system to be reliable and deterministic, we **must** be able to store more than one element we've encountered so far. Once you do that, however, there are several ways you can make use of that information.

One option is to enforce the behavior where registering two identical entries with the same priority is always considered an error. However, as demonstrated by cases such as `ACB` and `BCA`, this requires recording **every** entry encountered and then checking all of them against each other.

My implementation, on the other hand, treats mod `C` as a tie-breaker. Even though `A` and `B` compete for the same spot, once `C` comes in, their conflict no longer matters, because `C` overrides them both. Imho, this approach makes the most sense, it's easier to implement, and it allows for flexible scenarios where one mod can resolve a conflict between two others. It also means we only need to store the entries with the highest priority encountered so far, rather than literally everything we've seen.

---

Now, I do admit that a list is not the absolute best data structure for this scenario, as we only ever need to store two elements at most in order to be able to identify a fail state. So, something like a mutable `Pair` would be a better fit. However, I'm not aware of anything like that being readily available in the context of Sodium's codebase, and creating a new data structure just for this one case seemed excessive. So, for simplicity's sake, I went with a list with the default capacity of 1, as that's everything we need for the absolute majority of valid cases.

PR #3877, opened 2026-08-14
Optimize pitcher plants

krabsticks1:
This PR adds some minor optimizations to pitcher plant models, which affect all Sodium versions they are present in (1.20.1, 1.21.1, 1.21.11, 26.1.x, 26.2.x). It makes the following changes:

1. Eliminates excess faces in the `pitcher_plant` models. The vanilla model defines six faces per plane when only two are needed.
2. Adds cullfaces to the "bulbs" of `pitcher_crop` models. Unfortunately these will not cull against farmland blocks (which the crop can only be placed on in vanilla survival), but this might be useful to builders or mapmakers who can utilise the blocks more freely. It doesn't hurt to have them.

I did not change the mapping of the affected models or the texture references, only the changes listed above. They should appear identical and this seemed to be the case in some quick testing I did.

I do have some questions about further changes to the model, so I will mark this as a draft for now and ask on the Discord.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,3 +27,4 @@ Thank you to all the contributors who added to this release!
 Alpha series changelog:
 - Fixed crash when opening video settings
 - Disabled automatic enabling of the arena buffer debug widget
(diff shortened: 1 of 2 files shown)


krabsticks1:
Marking for review since I've decided the changes aren't necessary

PR #3878, opened 2026-08-15
fix vertex indexing, shadow attributes, normal fallbacks, transform math, and option equality

Spagles:
- fixed argument ordering in `ShadowFeatureRendererMixin` where `LightCoordsUtil.FULL_BRIGHT` and `OverlayTexture.NO_OVERLAY` were passed in reverse order to `EntityVertex.write()`. Because `EntityVertex.write()` expects `overlay` followed by `light`, passing them inverted caused lightmap coords to be written to the overlay attribute and overlay texture coords to be written to the lightmap attribute, corrupting shadow vertex data.
- reset `vertexIndex = 0` immediately upon reaching 4 vertices in `ChunkVertexConsumer.potentiallyEndVertex()`. Previously, if `collector.appendQuad()` returned `true` (indicating a culled/degenerate translucent quad), the method returned early without resetting `vertexIndex`, causing the subsequent `addVertex()` call to attempt writing to index 4 of the 4-element `vertices` array and throwing an `ArrayIndexOutOfBoundsException`.
- cached initial `x`, `y`, and `z` values into local variables in `QuadToPosPipe.accept()`. Mutating `this.vec.x` in place with `transformPositionX()` caused the subsequent `transformPositionY()` and `transformPositionZ()` calls to read the already-transformed `x` coordinate instead of the original component, corrupting the resulting `y` and `z` positions.
- changed the fallback return value of `getVertexNormal()` in `BakedQuadMixin` (NeoForge) from `-1` to `0` when `bakedNormals` is null or `UNSPECIFIED`. Sodium uses `0` as the sentinel value indicating absent vertex normals to trigger fallback to `getFaceNormal()`; returning `-1` (`0xFFFFFFFF`) caused the renderer to treat `-1` as a valid packed normal vector, unpacking garbage floats and breaking model lighting on NeoForge.
- replaced reference inequality (`!=`) with `!Objects.equals()` in `StatefulOption`. For boxed types outside the JVM cache range of `-128..127` (such as chunk fade duration at 750 or FPS limit at 260), separate `Integer` instances with identical values evaluated to not equal, causing options to spuriously report `hasChanged() == true` and keep the apply button and dirty indicators active even after reverting changes.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/StatefulOption.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/client/config/structure/StatefulOption.java
@@ -11,6 +11,7 @@ import net.minecraft.network.chat.Component;
 import net.minecraft.resources.Identifier;
 
 import java.util.Collection;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -78,7 +79,7 @@ public abstract class StatefulOption<V> extends Option {
     }
 
     public void modifyValue(V value) {
-        if (this.modifiedValue != value) {
+        if (!Objects.equals(this.modifiedValue, value)) {
             this.modifiedValue = value;
             this.state.invalidateDependents(this.dependents);
         }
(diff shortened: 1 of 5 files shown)


jellysquid3:
https://www.reddit.com/r/feedthebeast/comments/1ve0uq5/likely_ai_github_account_submitting_flawed_prs_to/

PR #3890, opened 2026-08-23
Fix OpenGL performance regression on some systems since 26.2

kupa22:
26.2 seems to have introduced a major OpenGL performance regression, which halves the framerate on some systems in almost all situations (at least on Apple Silicon devices).

Could reproduce on an M1 Max system in a newly created flat world (~230 FPS without the fix, ~580 FPS with the fix).

Problem is present since 26.2 pre-release 1 (since this version, a GL fence sync seems to be created unconditionally for each frame).

This PR contains a small mixin to work around this, which fixed the problem in my case (and hopefully also for others).

Since I don't have much experience with OpenGL or rendering, I'm not sure if this could have unintended side effects, but I haven't encountered any issues in my testing.

Would be great to have this tested on different systems and hardware configurations to see whether the issue is specific to my setup or affects other systems as well.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/sodium/mixin/core/GlFenceSyncMixin.java
+++ b/common/src/main/java/net/caffeinemc/mods/sodium/mixin/core/GlFenceSyncMixin.java
@@ -7,13 +7,7 @@ import org.spongepowered.asm.mixin.injection.Redirect;
 @Mixin(targets = "com.mojang.blaze3d.opengl.GlCommandEncoder")
 public class GlFenceSyncMixin {
 
-    @Redirect(
-            method = "submit",
-            at = @At(
-                    value = "INVOKE",
-                    target = "Lorg/lwjgl/opengl/GL33C;glFenceSync(II)J"
-            )
-    )
+    @Redirect(method = "submit", at = @At(value = "INVOKE", target = "Lorg/lwjgl/opengl/GL33C;glFenceSync(II)J"))
     private long sodium$disableGlFenceSync(int condition, int flags) {
         return 0L;
     }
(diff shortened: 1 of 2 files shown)


ILUvMyCLiQuELiKeKaNyEWeSt:
bro can u send jar i want to test cause im gettin fps drop as well

douira:
We've merged some related code recently. Is the performance problem fixed on the current `dev` branch?

kupa22:
> We've merged some related code recently. Is the performance problem fixed on the current `dev` branch?

I've just tested - makes no difference in my case (i.e. problem is not fixed).

ILUvMyCLiQuELiKeKaNyEWeSt:
not fixed bro

> We've merged some related code recently. Is the performance problem fixed on the current `dev` branch?

GeForceLegend:
> it breaks memory safety, and results in undefined behavior. even on mac, the fences are necessary

According to RogueLogix, your fix is completely wrong. You need someway to reduce latency, instead of just disable glFence.

Can you check if applying https://github.com/CaffeineMC/sodium/pull/3850 on sodium before (not including) commit 2c9463c would fix it? If the performance problem is caused by buffer create and destroy in every few frame, I may have some idea on how to let glFence work better on staging buffers.

jellysquid3:
Removing the fence entirely is not an acceptable solution to anything.

kupa22:
> Can you check if applying #3850 on sodium before (not including) commit [2c9463c](https://github.com/CaffeineMC/sodium/commit/2c9463c05df8835296b5bf330b830a56a9c0a336) would fix it? If the performance problem is caused by buffer create and destroy in every few frame, I may have some idea on how to let glFence work better on staging buffers.

I've checked #3850 - on my system it had a slight positive effect on framerate in the flat world
test (~280 FPS compared to ~230 FPS without the patch).

PR #3905, opened 2026-09-10
Mixin Cleanup & Fixes

AViewFromTheTop:
- Cleaned up duplicate/redundant invocations of Minecraft.getInstance()
- Re-implemented a few mixin Overwrites without using Overwrite
- Implemented missing Dry Foliage color optimization
- Replaced multiple instances of mixin methods using cir.getReturnValue() with new methods simply using ModifyReturnValue
- Misc. cleanup

I personally _would_ argue for altering the codestyle of the mod altogether in some places if that's something the Sodium team is interested in, there are many places it can improve readability (especially mixin parameters/AT annotations.)

I have also noticed that many classes that only provide static methods aren't marked as final and don't have private constructors, and there is major inconsistency between whether methods, new fields, new methods, etc., in mixins are prefixed with "sodium$" or not, as wall as frequent usage of "var" for local variables instead of providing their proper types. For the sake of consistency and readability for new contributors, I would advocate for a consensus on these too!

As I am not a member of the Sodium team I opted to not alter the codestyle much, and only made changes where it seemed to benefit consistency given what was already there.

On a side-note, the name of this pull request is likely sub-optimal. I'm not sure what it should be changed to, but this description should explain it well enough.


common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/entity/ModelPartMixin.java
```diff
@@ -1,51 +1,81 @@
...
-        }
-
-        if (this.xRot != 0.0F || this.yRot != 0.0F || this.zRot != 0.0F) {
-            MatrixHelper.rotateZYX(matrixStack.last(), this.zRot, this.yRot, this.xRot);
-        }
-
-        if (this.xScale != 1.0F || this.yScale != 1.0F || this.zScale != 1.0F) {
-            matrixStack.scale(this.xScale, this.yScale, this.zScale);
-        }
+    @WrapWithCondition(
```
douira, line 37: can you explain what what this set of changes is doing and why?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/gui/hooks/debug/DebugScreenOverlayInsertMixin.java
```diff
@@ -46,21 +46,22 @@ public class DebugScreenOverlayInsertMixin {
             }
         }
 
-        var sb = new StringBuilder();
+        StringBuilder builder = new StringBuilder();
```
douira, line 49: why is this necessary or better?

common/src/main/java/net/caffeinemc/mods/sodium/client/checks/ResourcePackScanner.java
```diff
@@ -98,9 +98,9 @@ private static void printToasts(Collection<ScannedResourcePack> resourcePacks) {
     }
 
     private static void printCompatibilityReport(Collection<ScannedResourcePack> scanResults) {
-        var builder = new StringBuilder();
+        StringBuilder builder = new StringBuilder();
```
douira, line 101: is this better or necessary?

douira (changes requested):
I agree with most of the changes but I'm not quite sure about some of the changes from var to specific types. We have a lot of usage of var throughout the codebase and this is largely intentional as in many places the type is obvious or verbose. Can you explain why in these particular places you found it useful to expand to the full type?

common/src/main/java/net/caffeinemc/mods/sodium/mixin/features/render/entity/ModelPartMixin.java
AViewFromTheTop: This is re-implementing the Overwrite, but using MixinExtras functionality in its place. I don’t like the conflict-heavy nature of Overwrite, but I also honestly didn’t expect this change in particular to be well-accepted.

PR #3916, opened 2026-09-20
Fix Fluid Culling For Non-Full Blocks With Unusual Shapes

douira:
I rewrote how we do this type of culling to actually take into account the fluid height when doing this calculation. I'm doing some direct voxel grid scanning here because generating a billion slice shapes and joining them is tedious, slow, and in my opinion, rather silly.

Fixes #3197

@Lolothepro please test that this indeed fixes the issue and doesn't regress anything. The minimal test I did seemed to be ok. Since the reproduction case is on 26.1 that is what this branch targets but I can also cherry pick it to other branches that have a similar version of the fluid renderer.


Lolothepro:
works fine

